PR #3, opened 2016-06-04
Don't double-iterate during MultiDict initialization

### LINKED ISSUE #2: Python MultiDict constructor discards generator content
When using the pure Python implementation, if a `MultiDict` is constructed with a generator argument:

    headers = CIMultiDict(
        (
            k.decode('utf-8', 'surrogateescape'),
            v.decode('utf-8', 'surrogateescape'),
        )
        for k, v in event.headers
    )

then the resulting `MultiDict` will be empty, instead of containing the key/value pairs as expected. This is because the generator is iterated over twice, and the first iteration discards all of the pairs.

jchampio:
Previously, the MultiDict constructor iterated twice over its argument:
once as a sanity check on its contents, and once again to add those
contents to the dictionary. If a generator was passed as an argument,
the first iteration would drain it, and the resulting dictionary would
be empty.

To fix this, save the contents of the generator in another list. This
fixes #2.

### DIFF
--- a/multidict/_multidict_py.py
+++ b/multidict/_multidict_py.py
@@ -185,12 +185,13 @@ class MultiDict(_Base, abc.MutableMapping):
             elif hasattr(arg, 'items'):
                 items = arg.items()
             else:
+                items = []
                 for item in arg:
                     if not len(item) == 2:
                         raise TypeError(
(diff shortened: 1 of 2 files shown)


asvetlov:
thanks

PR #41, opened 2017-01-16
Improve make clean

bachp:
While packaging multidict for OpenEmbedded I noticed several issues with `make clean`.

This is my proposal on how to fix them.


AraHaan:
This looks good to me. However I do not have merge permission. As such I would also like @fafhrd91, @asvetlov to also review this in case they want this merged.

asvetlov:
Thanks.
Sorry for late response.

PR #77, opened 2017-05-23
fix pickling of multidicts

thehesiod:
fixes: https://github.com/aio-libs/multidict/issues/78
fixes when sending multidicts across processes

testcase:
```python
import asyncio
from concurrent.futures import ProcessPoolExecutor
from aiohttp.web import Response
from multidict import CIMultiDict


class JSONResponseException(Exception):
    def __init__(self, http_status_code:int, payload_object=None, response_headers: CIMultiDict = None):
        self.__http_status_code = http_status_code
        self.__payload_object = payload_object if payload_object is not None else {}
        self.__response_headers = response_headers if response_headers is not None else CIMultiDict()

    @property
    def http_status_code(self):
        return self.__http_status_code

    @property
    def payload_object(self):
        return self.__payload_object

    @property
    def response_headers(self):
        return self.__response_headers


def foo():
    err = JSONResponseException(404)
    raise err

async def main(loop):
    with ProcessPoolExecutor() as executor:
        try:
            await loop.run_in_executor(executor, foo)
        except JSONResponseException as e:
            return Response(body=b'',
                            headers=e.response_headers,
                            status=e.http_status_code,
                            content_type='application/json')

if __name__ == '__main__':
    _loop = asyncio.get_event_loop()
    _loop.run_until_complete(main(_loop))
```

### DIFF
--- a/multidict/_multidict.pyx
+++ b/multidict/_multidict.pyx
@@ -1,7 +1,7 @@
+import cython
 import sys
 from collections import abc
 from collections.abc import Iterable, Set
-from operator import itemgetter
 
 
 _marker = object()
@@ -197,46 +197,42 @@ cdef class _Base:
 
 
 cdef class MultiDictProxy(_Base):
+    _proxy_classes = (MultiDict, MultiDictProxy)
+    _base_class = MultiDict
 
     def __init__(self, arg):
         cdef _Base base
-        if not isinstance(arg, (MultiDict, MultiDictProxy)):
+        if not isinstance(arg, self._proxy_classes):
             raise TypeError(
-                'ctor requires MultiDict or MultiDictProxy instance'
+                'ctor requires {} instance'
                 ', not {}'.format(
+                    ' or '.join(self._proxy_classes),
                     type(arg)))
 
         base = arg
         self._items = base._items
 
+    def __reduce__(self):
+        return (
+            self._base_class,
+            tuple([list(self.items())])
+        )
+
     def copy(self):
         """Return a copy of itself."""
-        return MultiDict(self._items)
+        return self._base_class(self._items)
 
 abc.Mapping.register(MultiDictProxy)
 
 
 cdef class CIMultiDictProxy(MultiDictProxy):
-
-    def __init__(self, arg):
-        cdef _Base base
-        if not isinstance(arg, (CIMultiDict, CIMultiDictProxy)):
-            raise TypeError(
-                'ctor requires CIMultiDict or CIMultiDictProxy instance'
-                ', not {}'.format(
-                    type(arg)))
-
-        base = arg
-        self._items = base._items
+    _proxy_classes = (CIMultiDict, CIMultiDictProxy)
+    _base_class = CIMultiDict
 
     cdef str _title(self, s):
         if type(s) is self._istr:
             return <str>s
         return s.title()
-
-    def copy(self):
-        """Return a copy of itself."""
-        return CIMultiDict(self._items)
 
 
 abc.Mapping.register(CIMultiDictProxy)
--- a/multidict/_multidict.pyx
+++ b/multidict/_multidict.pyx
@@ -247,8 +247,13 @@ cdef class MultiDict(_Base):
 
     def __init__(self, *args, **kwargs):
         self._items = []
-
         self._extend(args, kwargs, 'MultiDict', True)
+
+    def __reduce__(self):
+        return (
+            self.__class__,
+            tuple(self.items()),
+        )
 
     cdef _extend(self, tuple args, dict kwargs, name, bint do_add):
         cdef _Pair item
--- a/multidict/_multidict.pyx
+++ b/multidict/_multidict.pyx
@@ -1,6 +1,5 @@
 from __future__ import absolute_import
 
-import pickle
 import sys
 from collections import abc
 from collections.abc import Iterable, Set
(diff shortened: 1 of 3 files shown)


thehesiod:
this can also be solved via the 5th item in the tuple by returning an iterator that will set each key/value pair.

alexmac:
👍  Thanks for the fix!

asvetlov:
Looks good but please add a test.

thehesiod:
ok fixed, with some clean-up.  btw, if I read this right it's using a list to store the keys which seems really inefficient, why doesn't it use a dict with lower key, and value tuple of orig key and value?

thehesiod:
also perhaps we should throw an exception when pickling the proxies as we can't re-create the proxy as we can no longer reference the base object w/o re-creating it.  thoughts?

asvetlov:
> btw, if I read this right it's using a list to store the keys which seems really inefficient, why doesn't it use a dict with lower key, and value tuple of orig key and value?

Multidict should keep the order and it's used for relative small dictionaries where lookup could be fast enough.

> we should throw an exception when pickling the proxies as we can't re-create the proxy as we can no longer reference the base object w/o re-creating it.

Complicated question. What is behavior of `types.MappingProxyType` on forking?

thehesiod:
order: I think you could have used an `OrderedDict` with special values
MappingProxy example:
```python
>>> from types import MappingProxyType
>>> import pickle
>>> d = {'a': 1, 'b': 2}
>>> m = MappingProxyType(d)
>>> b = pickle.dumps(m)
_pickle.PicklingError: Can't pickle <class 'mappingproxy'>: attribute lookup mappingproxy on builtins failed
```

asvetlov:
Not sure about `OrderedDict` but proxies should explicitly forbid pickling like `MappingProxyType` does.

asvetlov:
Now I'm working on several multidict improvements.
At the end I'll take a look on pickling.
It should be done until multidict 3.0 release

thehesiod:
cool, sorry I didn't get to this, fighting a flea infestation out our place due to stray cats and a bad change to the AHPS file format :(  Would you like me to fix the proxy thing in this PR or hold off for your change to close this?

asvetlov:
Please go ahead.
All my changes will not touch pickling specifics.
But please make comprehensive tests. 
I mean pickling for MultiDict/CIMultiDict and forbidding it for proxies in both pure Python and Cython

thehesiod:
ok let me know what you think

multidict/_multidict.pyx
```diff
@@ -224,37 +224,34 @@ cdef class _Base:
...
+                'ctor requires {} instance'
                 ', not {}'.format(
+                    ' or '.join(self._proxy_classes),
                     type(arg)))
 
         base = arg
         self._items = base._items
 
+    def __reduce__(self):
+        raise pickle.PicklingError("Unsupported")
```
asvetlov: `MappingProxyType` raises `TypeError: can't pickle mappingproxy objects`.
Why do you use `PicklingError` exception? Is just `TypeError` not enough?
Also I'd like to have more comprehensive error text like `MappingProxyType` has.

thehesiod: look at my example about, it actually raises: `_pickle.PicklingError: Can't pickle <class 'mappingproxy'>: attribute lookup mappingproxy on builtins failed`

asvetlov:
Say again, on Python 3.6 pickling of `MappingProxyType` raises TypeError:
```
In [17]: from types import MappingProxyType

In [18]: import pickle

In [19]: d = {'a': 1, 'b': 2}

In [20]: m = MappingProxyType(d)

In [21]: pickle.dumps(m)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-21-34f44ac8a44e> in <module>()
----> 1 pickle.dumps(m)

TypeError: can't pickle mappingproxy objects

In [22]: 
```
I just don't want to import other module.

thehesiod:
behavior is different in 3.5:
```python
Python 3.5.2 (default, Oct 11 2016, 04:59:56) 
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.38)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from types import MappingProxyType
>>> import pickle
>>> d = {'a': 1, 'b': 2}
>>> m = MappingProxyType(d)
>>> pickle.dumps(m)
_pickle.PicklingError: Can't pickle <class 'mappingproxy'>: attribute lookup mappingproxy on builtins failed
```

and verified 3.6 does as you describe, so do we need two versions of the exception?

asvetlov:
No. I think just `TypeError` should be enough.
It's not very big change but I'd like to have `TypeError` for every python version

thehesiod:
ok changed, note though that this means that <3.6 users will not get the exception type they expect.

asvetlov:
pls fix failed test

asvetlov:
fix pep8 please

asvetlov:
Thank you very much!

thehesiod:
ty2!

asvetlov:
I mean thank for your patient.
I'll release new version very soon.
New `CIMultiDict` will keep header name casing.
Are you ready?

thehesiod:
btw ty2 = thank you too :)
thanks, after this lands we in aiobotocore can stop using aiohttp's `raw_headers`.   @alexmac will be using the new pickling support

Issue #90, opened 2017-06-08
Replace istr()


asvetlov:
Deriving from `str` is too expensive.
We need fast and simple way to figure out is key already titled (canonized) or not.

The alternative could be `multidict.canonize(s: str) -> str` function.
It converts parameter to `s.title()`, intern it by `sys.intern(...)` call and store the result into global dictionary.
Thus we have a very fast check for  canonized strings.

The drawback or proposal is the canonized strings are never deleted (I think it's fine for `aiohttp.hdrs` usage). All `canonize()` calls should be done on module import stage, that's why I propose a new function  instead of adding new tricks to `istr`.

`istr` will be supported for years anyway but `canonize` is the only way to keep multidicts fast.
Right now `istr` calls is 10-25% slower on `MultiDict` -- its not acceptable (but it's still much faster on `CIMultiDict` lookups than plain `str`).

asvetlov:
Now CIMultiDict doesn't use `istr` internally.
The class is still useful for making module-level constants like `istr("Content-Length")`, the performance of this usage is pretty satisfactory.

PR #96, opened 2017-06-18
Implement istr as C Extension

asvetlov:


gagoman:
Is https://github.com/aio-libs/multidict/issues/90 still relevant?
I can try to implement it.

asvetlov:
Honestly I don't know.
Now istr is much faster, I have no numbers about possible future speedups.

Issue #105, opened 2017-07-07
`istr` causes memory leak


gpfei:
Hi,

## Long story short

```python
# Warning: This script may cost about 100MB of memory.
import gc
import time

from multidict import istr


for i in range(1000000):
    # the memory of this temporary instance should be released after this loop,
    # but the result is not.
    istr('Content-Type')

gc.collect()
print('Done.')
time.sleep(300)
```

## Environment

- Python 3.5.3
- multidict 3.1.0
- Mac OS 10.12.5 & Centos Linux 6.6

Thanks!

thehesiod:
just came to this same conclusion!

thehesiod:
note: after fixed please add unittest for leak checking, can use resource.getrusage, psutil or something similar

thehesiod:
I've been trying to find the leak, and with the following code only istr1 + istr2 leak, so it seems to point to the leak being in the bottom part of `istr_new `

```python
import asyncio
import multidict
from fbn.utils.profiling import MemTracer


def test():
    istr1 = multidict.istr('foobarbaz')
    istr2 = multidict.istr()
    istr3 = multidict.istr(istr2)
    for _ in range(100):
        istr1.title()
        str(istr1)

async def main():
    tracer = MemTracer(5, '/tmp/traces.txt')

    # warmup
    test()

    while True:
        test()
        tracer.tick()


if __name__ == '__main__':
    _loop = asyncio.get_event_loop()
    _loop.run_until_complete(main())
```

gpfei:
`PyUnicode_Type.tp_new()` is called in `istr_new()`, but `PyUnicode_Type.tp_dealloc()` is missed in `istr_dealloc()`.

I tried the code below, it seems to work.

```c
void istr_dealloc(istrobject *self)
{   
    Py_XDECREF(self->canonical);
    PyUnicode_Type.tp_dealloc((PyObject*)self);  // <- free the memory
} 
```

thehesiod:
@gpfei hah!  I had tried `Py_TYPE(self)->tp_free(obj);` and it didn't work, nice find!  What's interesting is that I went all the way to trying with valgrind and it didn't find any leaks just now ;)

thehesiod:
@gpfei I created a PR with your fix.  I'll add a testcase after returning from the food trucks.  I'm kinda anxious for fix because our prod checkers are failing due to this.

gpfei:
@thehesiod failing due to this fix?

[...]

PR #106, opened 2017-07-08
leakfix

thehesiod:
fix from @gpfei

resolves issue: https://github.com/aio-libs/multidict/issues/105


asvetlov:
LGTM.
Do you want to write a test?
Making reliable test for this case is hard: the problem is not missing DECREF, collecting memory info from the whole process might vary.

thehesiod:
added test, seems to work

asvetlov:
multidict 3.1.1 released

Thanks for contribution

thehesiod:
thanks to @gpfei ! I was racking my brain trying to figure out what was wrong :)

PR #123, opened 2017-08-20
Fix MyPy type stubs

jaens:
The current type stubs can not actually be used from a client
perspective, since MyPy considers the declared classes abstract due to
missing definitions for methods such as `__getitem__`.

Example error message:
```
tests/test_mypy.py:4: error: Cannot instantiate abstract class 'MultiDict' with abstract attributes '__delitem__', '__getitem__', '__iter__', '__len__' and '__setitem__'
tests/test_mypy.py:5: error: Cannot instantiate abstract class 'CIMultiDict' with abstract attributes '__delitem__', '__getitem__', '__iter__', '__len__' and '__setitem__'
tests/test_mypy.py:6: error: Cannot instantiate abstract class 'MultiDictProxy' with abstract attributes '__getitem__', '__iter__' and '__len__'
tests/test_mypy.py:7: error: Cannot instantiate abstract class 'CIMultiDictProxy' with abstract attributes '__getitem__', '__iter__' and '__len__'
```

The commit add the necessary type definitions to make all the classes
non-abstract and adds a "test" to make sure these type definitions can
actually be used.


Makefile
```diff
@@ -21,7 +21,7 @@ rmcache:
 
 
 mypy: .develop
-	mypy multidict
+	mypy multidict tests/test_mypy.py
```
webknjaz: Why exactly did you change this command?

jaens: It's in the pull request description:
The commit add the necessary type definitions to make all the classes
non-abstract and **adds a "test" to make sure these type definitions can
actually be used**.

(this is the type-checking test, which is outside the multidict module obviously)

webknjaz: oh.. I missed the point of autodiscovery

asvetlov:
thanks!

PR #124, opened 2017-08-27
Fix equality check when other contains more keys

gagoman:
Test shows the issue.

Only Python version is affected.

P.S. I can not come up with more test cases, so I'm not sure that fix is final

### DIFF
--- a/multidict/_multidict_py.py
+++ b/multidict/_multidict_py.py
@@ -120,7 +120,7 @@ class _Base:
                 if i1 != i2 or v1 != v2:
                     return False
             return True
-        if len(self) != len(other):
+        if len(self._impl._items) != len(other):
             return False
         for k, v in self.items():
             nv = other.get(k, _marker)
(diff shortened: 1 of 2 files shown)


multidict/_multidict_py.py
```diff
@@ -120,6 +120,8 @@ def __eq__(self, other):
                 if i1 != i2 or v1 != v2:
                     return False
             return True
+        if len(self) != len(other):
```
webknjaz: It must be possible to move this to before line `114` and remove check at line `117`. They are equal, so no need to have it copy-pasted in two places.

asvetlov: I disagree.
Check from line 117 is faster than line 123: no python magic method `__len__` is involved.
Code duplication is very low, let's merge PR as is.

webknjaz: If it's about speed, then I'd convert `len(self)` into `self._impl._items`

asvetlov:
done

PR #133, opened 2017-09-13
Prevent Travis CI from spawning deploy job if !tag

webknjaz:
Ref: https://blog.travis-ci.com/2017-09-12-build-stages-order-and-conditions


asvetlov:
thanks

PR #139, opened 2017-09-16
Don't include python3.6 so into every binary wheel

asvetlov:


webknjaz:
How do these changes relate to the PR name?

asvetlov:
Sorry, it was my mistake

PR #146, opened 2017-09-29
Fix test runners integration mess

webknjaz:
* Add pytest-runner integration instead of self-crafted class
* Drop tox from requirements as nobody uses it (there's even no `tox.ini` present)

Includes #145

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,2 +1,5 @@
+[aliases]
+test = pytest
+
 [metadata]
 license_file = LICENSE
--- a/setup.py
+++ b/setup.py
@@ -65,7 +65,7 @@ def read(f):
 
 
 NEEDS_PYTEST = {'pytest', 'test'}.intersection(sys.argv)
-pytest_runner = ['pytest-runner']
+pytest_runner = ['pytest-runner'] if NEEDS_PYTEST else []
 
 tests_require = ['pytest']
 
--- a/setup.py
+++ b/setup.py
@@ -6,7 +6,6 @@ from setuptools import setup, Extension
 from distutils.errors import (CCompilerError, DistutilsExecError,
                               DistutilsPlatformError)
 from distutils.command.build_ext import build_ext
-from setuptools.command.test import test as TestCommand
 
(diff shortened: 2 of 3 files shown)


setup.py
```diff
@@ -65,14 +64,8 @@ def read(f):
...
 
-class PyTest(TestCommand):
-    user_options = []
-
-    def run(self):
-        import subprocess
-        errno = subprocess.call([sys.executable, '-m', 'pytest', 'tests'])
-        raise SystemExit(errno)
-
+NEEDS_PYTEST = {'pytest', 'test'}.intersection(sys.argv)
```
asvetlov: How it works?

setup.cfg
```diff
@@ -0,0 +1,2 @@
+[aliases]
+test = pytest
```
asvetlov: Please no.
I pretty happy with `pytest` name.

setup.py
webknjaz: this just checks for convenient argument at adds pytest-runner to setup_requires only if it's needed, meaning it won't install `pytest-runner` if you just do `sdist` or `bdist_wheel`

setup.cfg
webknjaz: It's an official way of integration + it worked this way before: https://github.com/aio-libs/multidict/pull/146/files/9e271b8286eda8237ad569495095c2b3793d81bc#diff-2eeaed663bd0d25b7e608891384b7298L104

webknjaz: AFAIK you don't even use this

webknjaz: I'd disable possibility for `python setup.py test` to invoke `unittest`

asvetlov: Finally got it.

setup.py
asvetlov: But you never check the value, isn't it?

webknjaz: Oops, it's intended to be checked on the next line, I'll fix this.

webknjaz: @asvetlov done

PR #152, opened 2017-10-01
Make tests compatible with PyPy3.5-5.8.0

webknjaz:


tests/test_copy.py
```diff
@@ -1,28 +1,75 @@
...
                                      MultiDictProxy as PyMultiDictProxy,
                                      CIMultiDictProxy as PyCIMultiDictProxy)
 
 
-@pytest.fixture(params=[MultiDict, PyMultiDict,
-                        CIMultiDict, PyCIMultiDict],
-                ids=['MultiDict', 'PyMultiDict',
-                     'CIMultiDict', 'PyCIMultiDict'])
+@pytest.fixture(
+    params=(
```
asvetlov: I prefer more dense formatting but can live with your notation

webknjaz: It looked less maintainable and less readable to me when it was dense.

PR #154, opened 2017-10-03
Use gc for memory usage control instead of psutils

asvetlov:


tests/test_istr.py
```diff
@@ -69,20 +66,13 @@ def _create_strs():
...
-        info2 = p.memory_info()
-        rss_diff = info2.rss - info.rss
-        if sys.platform == 'win32':
-            # on windows one page could be not returned
-            # to OS
-            assert rss_diff <= 4096
-        else:
-            assert rss_diff == 0
+        cnt2 = len(gc.get_objects())
+        assert abs(cnt - cnt2) < 10  # on PyPy these numbers are not equal
```
webknjaz: That's because PyPy doesn't immediately call `__del__()`

webknjaz: I saw somewhere in scrappy guys call `gc.collect()` twice in a row

asvetlov:
Let's live with current test for a while -- it is green now.

Issue #102, opened 2017-07-02
public class _Base


aamalev:
To effective detect a group of classes

    if isinstance(a, BaseMultidict):
        # dosomething1
    elif isinstances (a, Mapping):
        # dosomething2

The simplest implementation is less effective

    BaseMultidict = (CIMultiDict, MultiDict, MultiDictProxy)

asvetlov:
No. It is implementation detail and subject of future changes.
Next versions may have no such base class at all.

BTW what use cases for checking this type? Theoretically what you need is iteration over `a.items()` regardless concrete `Mapping` implementation.

aamalev:
The main project classes have a common interface described by common methods. Based on the principles of parametric polymorphism, it makes sense to single out a formally generalized interface. For python parametric polymorphism goes along with annotations and types such as Mapping, but Mapping exclude getall

In this case https://github.com/aamalev/aiohttp_apiset/blob/master/aiohttp_apiset/swagger/route.py#L132 could get away from duck typing

asvetlov:
In this case you need multidict ABCs for extending standard Mapping and MutableMapping.
I'm open for PR like this (documentation should be updated as well).
Personally I don't have very strong motivation for fixing it myself right now.

PR #157, opened 2017-10-04
Introduce abstract base classes

asvetlov:
Fix for #102


kxepal:
I hope you'll remove master merge commit on PR merge?

multidict/_multidict.pyx
```diff
@@ -238,7 +239,7 @@ cdef class MultiDictProxy(_Base):
         """Return a copy of itself."""
         return self._base_class(self)
 
-abc.Mapping.register(MultiDictProxy)
+MultiMapping.register(MultiDictProxy)
```
webknjaz: IMHO using it as a class decorator would look cleaner

kxepal: Is that possible in Cython?

webknjaz: cython must support decorators

asvetlov: What decorators are your talking about?
Cython doesn't support deriving from abc classes IIRC.
`ABC.register()` is a standard way for registering C Extension classes

asvetlov:
@kxepal I'll squash the PR on merging anyway.

kxepal:
cool!

multidict/_multidict.pyx
webknjaz: What about
```python
@ABC.register
class ...:
```
?

asvetlov: Aaah. Ok.

asvetlov: Cython cannot apply decorators to cdef classes :(

kxepal: That's what I expected.

PR #167, opened 2017-10-11
Refactor unit tests

gagoman:
Changes:

* merged test_instantiate__from_arg0 and test_instantiate__from_arg0_dict
* tests from `_BaseMutableMultiDictTests` are at `TestMutableMultiDict`
* `TestNonProxyCIMultiDict#test_extend_with_istr` was moved to `TestMutableMultiDict`
* `_TestProxy/_TestCIProxy#test_copy` was moved to `test_proxy_copy`

Questions:

* test_instantiate__with_kwargs: should we rely on order and drop `sorted`?
* test_instantiate__empty: is duplication done by purpose?
* test_items__contains: is duplication done by purpose?
* test_get duplicates test_getone
* `_CIMultiDictTests` was inherited from `_Root` and not `_BaseTest`
* `test__iter__types` was done only for non-ci dicts. Should I add CI classes?

PR #176, opened 2017-10-23
fix wrong test statement

gagoman:
Was introduced at #167 and led to coverage drop

[Original one](https://github.com/aio-libs/multidict/blob/v3.2.0/tests/test_multidict.py#L46)


asvetlov:
Good catch!

PR #180, opened 2017-10-30
renamed chained_ctor to chained_callable

gagoman:
* added documentation for a function

I'm bad at writing docs so it would be nice to have feedback


gagoman:
@webknjaz this is to address your request in https://github.com/aio-libs/multidict/pull/167

asvetlov:
Thanks!

PR #187, opened 2017-11-09
Add Changelog to docs

webknjaz:
Also:
* Enable autolinking of github issues and PEPs


docs/conf.py
```diff
@@ -313,3 +319,31 @@
...
+            dict(
+                pattern=r'((Issue|PR)\s?)?#(?P<issue_or_pr>\d+)',
+                url='{GH}/aio-libs/{project}/issues/{issue_or_pr}',
+            ),
+            dict(
+                pattern=r'^(?m)((?P<scm_version>v?\d+(\.\d+){1,2}))\n[-=]+\n',
+                with_scm='{text}\n{rev[timestamp]:%d %b %Y}\n',
+            ),
+            dict(
+                pattern=r'PEP[- ](?P<pep_number>\d+)',
```
kxepal: Better use [builtin markup](http://www.sphinx-doc.org/en/stable/markup/inline.html#role-pep).

docs/conf.py
```diff
@@ -313,3 +319,31 @@
...
+
+link_files = {
+    f: dict(
+        using=dict(
+            GH='https://github.com',
+        ),
+        replace=[
+            dict(
+                pattern=r'((Issue|PR)\s?)?#(?P<issue_or_pr>\d+)',
+                url='{GH}/aio-libs/{project}/issues/{issue_or_pr}',
```
kxepal: I would suggest to use extlinks instead like [this](https://github.com/apache/couchdb-documentation/blob/master/src/conf.py#L114). It's better for point of markup.

docs/conf.py
```diff
@@ -313,3 +319,31 @@
...
+            ),
+            dict(
+                pattern=r'^(?m)((?P<scm_version>v?\d+(\.\d+){1,2}))\n[-=]+\n',
```
kxepal: What this rule does?

asvetlov:
I welcome the intention but introduced regex kung fu scares me.
The plugin is not documented.
Is there better supported alternative?

docs/conf.py
webknjaz: Fair enough, but would it render nicely at github.com or pypi.org?

docs/conf.py
webknjaz: Oh, it's automatic replacement of version titles appending the release date from git tags.

docs/conf.py
webknjaz: Again, what about viewing this on GitHub?

webknjaz:
@asvetlov it's crafted by another CherryPy maintainer for the sake of reusability. It's basically designed for the sake of reusability by Jason. It's main purpose is to do some automatic replacement like linking issue, versions or adding dates etc.
Still, it's just a show-off of what could be done and we can use other tools as well.

docs/conf.py
kxepal: No, it wouldn't - it's not a part of standard reST. But if want follow the way "rendered on GitHub like a candy" so you should drop sphinx in favour of plain markdown. But markdown is not for documentation (;

I think that rendering issue is not a problem since we have RTD which does everything right.

docs/conf.py
kxepal: No, it wouldn't. But I wouldn't care about this as well.

webknjaz: @kxepal https://github.com/aio-libs/multidict/blob/master/setup.py#L77 `CHANGES.rst` is included there. We can't just break PYPI page in favor of SphinX, unless @asvetlov approves dropping it.

webknjaz:
@asvetlov @kxepal I've addressed all the comments

webknjaz:
@asvetlov turns out, `setup check` is not ready for such RST roles from SphinX.

The solution would be to not include changelog in `long_description` or inject some RST pre-processor replacing unknown stuff with canonical links.

webknjaz:
@asvetlov ?

webknjaz:
ping @asvetlov

asvetlov:
Now looks much better.
Please merge when green.

webknjaz:
@asvetlov what do you think about https://github.com/aio-libs/multidict/pull/187#issuecomment-355938287 ?

asvetlov:
A link on changelog page hosted on readthedocs is enough I think.

PR #203, opened 2018-01-28
Fix CIMultiDict output casing in Python implementations (to match Cython)

### LINKED ISSUE #202: Cython and Python versions of CIMultiDict treat key case differently
First noticed this in 3.3.2 (probably existed before that), still exists in 4.0.0:

As one would expect (IMO) the Cython version uses the title-ized key for matching and preserves the original key for output/serialization. The Python version throws away the original key completely and uses the title-ized key for both purposes, breaking output. Again, that's my opinion, but in any case, the behavior should be the same.

```python
In [1]: from multidict._multidict import CIMultiDict  # Cython

In [2]: md = CIMultiDict()
   ...: md['some_key'] = 1
   ...: list(md.items())
   ...:
Out[2]: [('some_key', 1)]  # outputs original key

In [3]: from multidict._multidict_py import CIMultiDict  # Python

In [4]: md = CIMultiDict()
   ...: md['some_key'] = 1
   ...: list(md.items())
   ...:
Out[4]: [('Some_Key', 1)]  # outputs title-ized key
```

mikenerone:
Fixes #202

### DIFF
--- a/multidict/_multidict_py.py
+++ b/multidict/_multidict_py.py
@@ -261,11 +261,11 @@ class MultiDict(_Base, MutableMultiMapping):
         self._replace(key, value)
 
     def __delitem__(self, key):
-        key = self._title(key)
+        identity = self._title(key)
         items = self._impl._items
         found = False
         for i in range(len(items) - 1, -1, -1):
-            if items[i][0] == key:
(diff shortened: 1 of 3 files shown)


asvetlov:
Thanks!

PR #236, opened 2018-05-21
Make simplier logic for pair_list_get_all and pair_list_pop_all

iemelyanov:


asvetlov:
The logic is intentional: for happy path, multidict has no or the only one value for any header.
In this case the code is as fast as possible.

PR #237, opened 2018-05-22
case insensitive pair

asvetlov:
Introduce key normalizer and case insensitive pair

### DIFF
--- a/multidict/_pair_list.c
+++ b/multidict/_pair_list.c
@@ -1,9 +1,7 @@
 #include <string.h>
 #include "_pair_list.h"
 
-#include "Python.h"
-#include "object.h"
-#include "structmember.h"
+#include <Python.h>
 
 // fix for VisualC complier used by Python 3.4
 #ifdef __GNUC__
@@ -22,6 +20,9 @@ static PyObject * _istr_type;
 _Py_IDENTIFIER(title);
 
 
+typedef PyObject * (*calc_identity_func)(PyObject *key);
+
+
 /* Global counter used to set ma_version_tag field of dictionary.
  * It is incremented each time that a dictionary is created and each
  * time that a dictionary is modified. */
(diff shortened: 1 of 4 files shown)


multidict/_pair_list.c
```diff
@@ -61,6 +67,66 @@ str_cmp(PyObject *s1, PyObject *s2)
...
+    }
+    PyErr_SetString(PyExc_TypeError,
+                    "MultiDict keys should be either str "
+                    "or subclasses of str");
+    return NULL;
+}
+
+
+static PyObject *
+key_to_identity(PyObject *key)
```
iemelyanov: I cant see a difference between this function and the previous key_to_str. Do we have to merge them in one?

multidict/_pair_list.c
```diff
@@ -61,6 +67,66 @@ str_cmp(PyObject *s1, PyObject *s2)
...
+        return PyObject_Str(key);
+    }
+    PyErr_SetString(PyExc_TypeError,
+                    "MultiDict keys should be either str "
+                    "or subclasses of str");
+    return NULL;
+}
+
+
+static PyObject *
```
iemelyanov: Mb add INLINE?

multidict/_pair_list.c
```diff
@@ -94,8 +160,9 @@ pair_list_resize(pair_list_t *list, Py_ssize_t new_capacity)
     return 0;
 }
 
-PyObject *
-pair_list_new(void)
+
+static PyObject *
```
iemelyanov: Mb add INLINE?

multidict/_pair_list.c
```diff
@@ -188,6 +269,40 @@ pair_list_add_with_hash(PyObject *op,
...
+              PyObject *key,
+              PyObject *value)
+{
+    Py_hash_t hash;
+    PyObject *identity = NULL;
+    PyObject *str_key = NULL;
+    pair_list_t *list = (pair_list_t *)op;
+    int ret;
+
+    identity = list->calc_identity(key);
```
iemelyanov: Indirect call by function ptr can be issued for performance optimization. We should understand this. Potentially we can use a preprocessor, but it will be more dirty code.

multidict/_pair_list.c
asvetlov: Done

multidict/_pair_list.c
asvetlov: I understand it but I think the code is ok.
It's like a virtual function call, very widely used technic by Python Core itself.

PR #242, opened 2018-05-22
fix setup to read version on windows.

alexpantyukhin:


asvetlov:
I'm curious why Windows build on codecov was successful before the PR?
Do you encounter the problem when building from git or using pip?

alexpantyukhin:
hm.. yes, you are right! I jumped to conclusions. It was working well before also.

PR #245, opened 2018-05-23
Improve pair list

asvetlov:
* Massive refactoring
* API simplification
* Keys restriction is relaxed to accept any str-derived types

### DIFF
--- a/multidict/_pair_list.c
+++ b/multidict/_pair_list.c
@@ -350,7 +350,7 @@ _pair_list_drop_tail(PyObject *op, PyObject *identity, Py_hash_t hash,
     return found;
 }
 
-int
+static int
 _pair_list_del_hash(PyObject *op, PyObject *identity,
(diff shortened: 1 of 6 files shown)


multidict/_pair_list.c
```diff
@@ -612,40 +669,48 @@ pair_list_pop_one(PyObject *op, PyObject *ident, PyObject *key)
...
         }
         else if (tmp < 0) {
-            return NULL;
+            goto fail;
         }
     }
 
     PyErr_SetObject(PyExc_KeyError, key);
-    return NULL;
+    goto fail;
```
iemelyanov: This "goto" not needed

asvetlov: I think this style is a little cleaner (more readable) than `goto` omitting.
Generated code should be the same.

iemelyanov:
Looking good to me.

PR #247, opened 2018-05-28
Move all identity calc to c

iemelyanov:


multidict/_pair_list.c
```diff
@@ -994,6 +992,13 @@ pair_list_update(PyObject *op1, PyObject *op2)
 }
 
 
+PyObject *
+pair_list_calc_identity(PyObject *op, PyObject *key)
```
asvetlov: Ok with the method as a part of the transition.
In the end, we probably will not need this function, at least not in public API.

multidict/_pair_list.c
```diff
@@ -93,19 +93,17 @@ ci_key_to_str(PyObject *key)
 {
     PyTypeObject *type = Py_TYPE(key);
     if ((PyObject *)type == _istr_type) {
-        // replace with direct .canonical attr
-        return _PyObject_CallMethodId(key, &PyId_title, NULL);
+        return PyObject_CallMethod(key, "lower", NULL);
     }
     if (PyUnicode_Check(key)) {
-        return _PyObject_CallMethodId(key, &PyId_title, NULL);
+        return PyObject_CallMethod(key, "lower", NULL);
```
asvetlov: Please keep it.
Passing `char *` is slow. For working in `LIMITED_API` boundaries we need to pass pre-existing `PyObject *` but I don't care at the moment.

multidict/_pair_list.c
```diff
@@ -93,19 +93,17 @@ ci_key_to_str(PyObject *key)
...
     }
     PyErr_SetString(PyExc_TypeError,
                     "CIMultiDict keys should be either str "
                     "or subclasses of str");
-    return 0;
+    return NULL;
```
asvetlov: Nice!

webknjaz:
@iemelyanov it looks like you did commits using email unbound from GitHub your account. You might want to fix that :)

multidict/_pair_list.c
iemelyanov: Yes. This is just for a transition.

iemelyanov:
@webknjaz Thanks, but I don't care :)

PR #251, opened 2018-06-02
Add new function for update list from mapping without temporary pair list

### LINKED ISSUE #243: Don't create a temporary pair list when updating from mapping

iemelyanov:
Resolves #243

### DIFF
--- a/multidict/_pair_list.c
+++ b/multidict/_pair_list.c
@@ -906,11 +906,14 @@ _pair_list_update(PyObject *op, PyObject *key,
                   PyObject *value, PyObject *used_keys,
                   PyObject *identity, Py_hash_t hash)
 {
-    Py_ssize_t pos;
-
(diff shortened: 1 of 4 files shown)


multidict/_pair_list.c
```diff
@@ -900,20 +900,86 @@ _pair_list_post_update(pair_list_t *list, PyObject* used_keys, Py_ssize_t pos)
...
+        else if (ident_cmp_res < 0) {
+            return -1;
+        }
+    }
+
+    if (!found) {
+        if (_pair_list_add_with_hash(op, identity, key, value, hash) < 0) {
+            return -1;
+        }
+        // TODO: mb here should by Py_INCREF (identity)?!
```
asvetlov: No. `PyDict_SetItem` increments a key and value.

asvetlov: Please just drop a comment.

PR #253, opened 2018-06-03
Add CFLAGS to setup.

iemelyanov:


setup.py
```diff
@@ -25,13 +26,19 @@
     else:
         macros = []
 
+    CFLAGS = ['-O2']
+    # CFLAGS = ['-g']
+    if platform.uname().system != 'Windows':
```
webknjaz: `platform.platform()`

iemelyanov: platform.platform() gives long string with lots of additional information, not just system name

webknjaz: Oh, I meant `platform.system()`

iemelyanov: Ok, thanks)

PR #257, opened 2018-06-04
Move mapping comparsion from cython to C.

iemelyanov:


multidict/_pair_list.c
```diff
@@ -1115,13 +1115,112 @@ pair_list_update_from_seq(PyObject *op, PyObject *seq)
...
+    }
+
+    // TODO: mb we can iterate by dict insted this overhead
+    items = PyMapping_Items(other);
+    if (items == NULL) {
+        return -1;
+    }
+
+#if PY_MINOR_VERSION <= 5
+    if (PyList_Sort(items) < 0) {
```
asvetlov: Let's change iteration order. Roughly
``` for k, v in self:
         if k in other and other[k] == v:
              ...
```
By this, we are independent of a mapping order.
Keep in mind: `other` can be not only `dict` instance but any dict-like object.
The check is a comparison to mapping, not to iterable of pairs.
The dict is not comparable with a list of pairs:
```
In [1]: {'a': 1} == [('a', 1)]
Out[1]: False
```
That's fine for multidict too.

iemelyanov: Thanks. Good decision!

multidict/_pair_list.c
```diff
@@ -1115,13 +1115,66 @@ pair_list_update_from_seq(PyObject *op, PyObject *seq)
...
+    }
+
+    if (pair_list_len(op) != PyMapping_Length(other)) {
+        return 0;
+    }
+
+    pos = 0;
+    is_eq = 1;
+    while (pair_list_next(op, &pos, NULL, &key, &avalue)) {
+        if (!PyMapping_HasKey(other, key)) {
```
asvetlov: The check is redundant.
`PyObject_GetItem` should be enough.
If its return value is `NULL` call `PyErr_Clear()` and assign `is_eq` to `0`.
Comparison to mapping is not the hot path (no need a specialization for dict instance) but I see no reasons to lookup a mapping twice.

iemelyanov: Yes, you right.

PR #262, opened 2018-06-20
Try fading to black and other linters experiment

webknjaz:
Hey @asvetlov, here's my experiment on standardizing linters + trying to bring black in place.

What do you think?


webknjaz:
@asvetlov failures can be adjusted, but I need you to validate the idea.

asvetlov:
Maybe apply black with default settings?
Otherwise I'd like to have `make black` command.
But strings normalization and 90 chars line length looks OK if applied consistently.
IIRC emacs can read line length for the project from config file.

webknjaz:
There's a number of other linters in pre-commit I'd like to keep and its config is quite a good place to consolidate options. We can try black with line-length arg, also I've applied ignore mode for quotes normalization.
`make` is not a cross-platform thing and the way it's used now is really harmful especially for new non-experienced contributors. I'm going to give tox another try a bit later. It could *unify* the CI experience and help contributors a lot.

asvetlov:
The main library contributor is me.
I use Linux and found `make` very handy.
Pre-commit hook is perfect but I'd like to keep `make` tool.
Also, I don't use tox: it takes too long for building all possible environments.

webknjaz:
I think you'd like to use tox the way I configure it: I'd run only your current python :)
It looks like I just need to show you better workflow with it.

Of course, we can wrap it all with `make`, I just want to untangle spaghetti there.
Also, think big: it's better to standardize tooling across all repos and in this case contributors matter.

I like GNU make, and I've used it for quite a while, but for the problem that I describe it's just not a right tool.

webknjaz:
@asvetlov I've made a demo of how it might look like with tox-centric setup. It cleans up CI setup a bit and makes test runner truly cross-platform: #263

PR #265, opened 2018-06-26
Move multidict iterators to C ext.

### LINKED ISSUE #241: Rewrite iterators with C
As a part of getting rid of Cython

iemelyanov:
Resolves #241


multidict/_multidict_iter.c
```diff
@@ -0,0 +1,309 @@
...
+    0,                                              /* tp_members */
+    0,                                              /* tp_getset */
+    0,                                              /* tp_base */
+    0,                                              /* tp_dict */
+    0,                                              /* tp_descr_get */
+    0,                                              /* tp_descr_set */
+    0,                                              /* tp_dictoffset */
+    0,                                              /* tp_init */
+    0,                                              /* tp_alloc */
+    (newfunc)multidict_iter_new,                    /* tp_new */
```
asvetlov: Please take a look on `PyDictIterKey_Type` and family.
I think we don't need a `__new__` method but a C function to create an iterator.
Iterator instances should not be instantiated by user Python code but by views only.

multidict/_multidict_iter.c
```diff
@@ -0,0 +1,309 @@
...
+    -1,                          /* m_size */
+    NULL,                        /* m_methods */
+    NULL,                        /* m_reload */
+    NULL,                        /* m_traverse */
+    NULL,                        /* m_clear */
+    NULL                         /* m_free */
+};
+
+PyMODINIT_FUNC
+PyInit__multidict_iter(void)
```
asvetlov: We don't need a module for iterators. I'm ok with putting them into a separate `.c` file but no python module.
Right now `pair_list.c` intentionally is not a module too but a file with `init` function.
Later we'll convert it into `_multidict` C extention which is built from several .c files but is a single python module.

multidict/_multidict_iter.c
iemelyanov: Ok. I added `__new__` just for compatibility with `_ItemsView` https://github.com/aio-libs/multidict/blob/master/multidict/_multidict.pyx#L497, etc.

And what we should do with `_ItemsView`, `_ValuesView`, `_KeysView`? Rewrite on C in this PR or separate?

multidict/_multidict_iter.c
iemelyanov: Ok

asvetlov:
Thanks. Sorry for delay.

PR #279, opened 2018-09-04
Exclude _istr.pyd from sdist

cgohlke:
`multidict-4.4.0.tar.gz` on PyPI contains `_istr.pyd`.


webknjaz:
@cgohlke feel free to update this PR

asvetlov:
Fixed by de2df30

PR #294, opened 2018-10-06
[WIP] Issue#98: refactor test suite

gyermolenko:
Disclaimer: work in progress. I need initial review to proceed.  

**Main points:**
- split tests by topics (instantiation, contents, comparisons)
- create parametrized fixtures - all 4 types of  instances (`[MultiDict, CIMultiDict, MultiDictProxy, CIMultiDictProxy]`) are pre-created and used for tests which concern not instantiation of those classes themselves 
(unfortunately, pytest so far [doesn't support fixtures in pytest.mark.parametrize] )(https://github.com/pytest-dev/pytest/issues/349)
- ~~use `autouse=True` class level hack (since we can't use `__init__`) to be able to use `self.d` instead of long fixture names.~~

Since my goal is restructuring - I didn't touch asserts themselves. I think that such comments can be addressed after the fact.

### DIFF
--- a/tests/test_multidict_.py
+++ b/tests/test_multidict_.py
@@ -1,0 +1,133 @@
+import pytest
+
+from multidict import MultiDict, CIMultiDict, MultiDictProxy, CIMultiDictProxy
+
+MUTABLE_DICTS = [MultiDict, CIMultiDict]
+PROXY_DICTS = [MultiDictProxy, CIMultiDictProxy]
+
+
+@pytest.fixture(params=MUTABLE_DICTS)
+def md_simple(request):
+    return request.param([('key1', 'value1')])
+
+@pytest.fixture(params=MUTABLE_DICTS)
+def md_w_multivalue_per_key(request):
+    return request.param([('key1', 'one'), ('key2', 'two'), ('key1', 3)])
+
+
+@pytest.mark.parametrize("cls", MUTABLE_DICTS)
+class TestInstantiation:
+
+    def test_instantiate__empty(self, cls):
+        d = cls()
+        assert d == {}
+        assert len(d) == 0
+        assert list(d.keys()) == []
+        assert list(d.values()) == []
+        assert list(d.items()) == []
+
+        assert cls() != list()
+        with pytest.raises(TypeError, match='\(2 given\)'):
+            cls(('key1', 'value1'), ('key2', 'value2'))
+
+    @pytest.mark.parametrize('arg0', [
+        [('key', 'value1')],
+        {'key': 'value1'}
+    ])
+    def test_instantiate__from_arg0(self, cls, arg0):
+        d = cls(arg0)
+
+        assert d == {'key': 'value1'}
+        assert len(d) == 1
+        assert list(d.keys()) == ['key']
+        assert list(d.values()) == ['value1']
+        assert list(d.items()) == [('key', 'value1')]
+
+    def test_instantiate__with_kwargs(self, cls):
+        d = cls([('key', 'value1')], key2='value2')
+
+        assert d == {'key': 'value1', 'key2': 'value2'}
+        assert len(d) == 2
+        assert sorted(d.keys()) == ['key', 'key2']
+        assert sorted(d.values()) == ['value1', 'value2']
+        assert sorted(d.items()) == [('key', 'value1'),
+                                     ('key2', 'value2')]
+
+    def test_instantiate__from_generator(self, cls):
+        d = cls((str(i), i) for i in range(2))
+
+        assert d == {'0': 0, '1': 1}
+        assert len(d) == 2
+        assert sorted(d.keys()) == ['0', '1']
+        assert sorted(d.values()) == [0, 1]
+        assert sorted(d.items()) == [('0', 0), ('1', 1)]
+
+    def test_cannot_create_from_unaccepted(self, cls):
+        with pytest.raises(TypeError):
+            cls([(1, 2, 3)])
+
+
+@pytest.mark.parametrize("cls", MUTABLE_DICTS+PROXY_DICTS)
+class TestContents:
+
+    @pytest.fixture(autouse=True)
+    def _autoassign_md_to_d(self, md_w_multivalue_per_key):
+        self.d = md_w_multivalue_per_key
+
+    def test_getting_items(self, cls):
+        assert self.d.getone('key1') == 'one'
+        assert self.d.get('key1') == 'one'
+        assert self.d['key1'] == 'one'
+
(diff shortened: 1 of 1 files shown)


gyermolenko:
@asvetlov sorry to bother, just a ping
Since this WIP needs your preliminary review before I proceed (or abort) with it.

asvetlov:
Will do.
Sorry for a pause

tests/test_multidict_.py
```diff
@@ -0,0 +1,133 @@
...
+    def test_cannot_create_from_unaccepted(self, cls):
+        with pytest.raises(TypeError):
+            cls([(1, 2, 3)])
+
+
+@pytest.mark.parametrize("cls", MUTABLE_DICTS+PROXY_DICTS)
+class TestContents:
+
+    @pytest.fixture(autouse=True)
+    def _autoassign_md_to_d(self, md_w_multivalue_per_key):
```
asvetlov: I like the parametrization approach but this particular fixture looks like a hack.

Instead of `self.d` we can always use a fixture for multidict instance explicitly.
This allows getting rid of `self.d` entirely.
`self.d` is necessary for `unittest` based approach but redundant for pytest

webknjaz: Also, classes are normally not necessary. If you need namespacing, flat is better than nested: just use separate test modules :)

tests/test_multidict_.py
```diff
@@ -0,0 +1,133 @@
+import pytest
```
webknjaz: I understand that this might look not needed, but it'd be great to see docstrings around :)
pytest picks those up from test functions and module docstring is a good teaser of what's going to be tested in here in general.
Although it's a nitpick, I'm pretty sure that tests are no exception.

tests/test_multidict_.py
asvetlov: Agree but I don't insist.
Can live with classes if they are already exist.

tests/test_multidict_.py
asvetlov: Docstrings in tests?
Looks like a sort of overengeneering, especially for such slow evolved and limited library as multidict.
Many people *use* multidict directly or indirectly but how many send patches?

webknjaz: It's more about being clean and caring about future self + others, who might be not sending patches also because it's hard or takes a lot of time to understand.

tests/test_multidict_.py
webknjaz: Yeah, but classes often implicitly encourage ppl to have state. Which might lead to tests depending on each other. Why rely on human reviewers to catch those when simple rules are more effective in terms of discouraging this? :)

tests/test_multidict_.py
```diff
@@ -0,0 +1,133 @@
...
+    def test_instantiate__empty(self, cls):
+        d = cls()
+        assert d == {}
+        assert len(d) == 0
+        assert list(d.keys()) == []
+        assert list(d.values()) == []
+        assert list(d.items()) == []
+
+        assert cls() != list()
+        with pytest.raises(TypeError, match='\(2 given\)'):
```
webknjaz: please use raw-string literals instead of escaping everything

tests/test_multidict_.py
asvetlov: Is getting rid of classes the suggestion or requirement?

tests/test_multidict_.py
asvetlov: Honestly, I don't care.
@webknjaz if you want to add docstrings everywhere in test suite -- you are welcome!

webknjaz: I want people to do this when contributing new things at least. Going over the existing code base is more resource consuming...

tests/test_multidict_.py
webknjaz: Suggestion. A requirement would need actual linting implementation.

webknjaz: (But still, it's also: simple is better than complex)

tests/test_multidict_.py
webknjaz: @gyermolenko consider https://youtu.be/o9pEzgHorH0

gyermolenko:
hey, @asvetlov @webknjaz 
I've updated this PR and my first comment.
Please review and leave your comments :)

tests/test_multidict_.py
```diff
@@ -0,0 +1,203 @@
...
+
+    def test_instantiate__empty(self, cls):
+        d = cls()
+        assert d == {}
+        assert len(d) == 0
+        assert list(d.keys()) == []
+        assert list(d.values()) == []
+        assert list(d.items()) == []
+
+        assert cls() != list()
```
webknjaz:
```suggestion
        assert d != list()
```

tests/test_multidict_.py
gyermolenko: @webknjaz regarding classes -> modules part:
I mostly agree but it is hard to split something in modules when you don't actually know all parts.
Like here I am still in progress figuring out the layout and all needed fixtures. 
When it is settled - I will probably come back to that thought.

tests/test_multidict_.py
```diff
@@ -0,0 +1,203 @@
...
+    def test_sub(self, md_twokeys):
+        assert {'key1'} == md_twokeys.keys() - {'key2'}
+
+    def test_sub2(self, md_twokeys):
+        assert {'key3'} == {'key1', 'key2', 'key3'} - md_twokeys.keys()
+
+    def test_xor(self, md_twokeys):
+        assert {'key1', 'key3'} == md_twokeys.keys() ^ {'key2', 'key3'}
+
+    def test_xor2(self, md_twokeys):
```
webknjaz: This  belongs to the previous test.
```suggestion
```

tests/test_multidict_.py
```diff
@@ -0,0 +1,203 @@
...
+    def test_and(self, md_simple):
+        assert {'key1'} == md_simple.keys() & {'key1', 'key2'}
+
+    def test_and2(self, md_simple):
+        assert {'key1'} == {'key1', 'key2'} & md_simple.keys()
+
+    def test_or(self, md_simple):
+        assert {'key1', 'key2'} == md_simple.keys() | {'key2'}
+
+    def test_or2(self, md_simple):
```
webknjaz: All test assertions with a reversed order, but checking basically the same should go to the same test.

gyermolenko:
hey @asvetlov , just a friendly ping.
It would be nice to have your comments on updated PR

webknjaz:
If this is still relevant after #915, please recreate the PR with those fixtures.

PR #296, opened 2018-10-12
Add 4th pickle protocol version to tests

gyermolenko:
also fix deprecation warning with raw string in one test


tests/gen_pickles.py
```diff
@@ -21,7 +21,7 @@ def write(name, proto):
 def generate():
     if not USE_CYTHON:
         raise RuntimeError("Cython is required")
-    for proto in range(pickle.HIGHEST_PROTOCOL):
+    for proto in range(pickle.HIGHEST_PROTOCOL+1):
```
webknjaz: please use spaces around operators.

gyermolenko: flake8 was alright with that version.
I don't mind the change but do you have any refs to support spacing?

webknjaz: It's more readable and a common standard to put spaces around operator (except in kwargs in function calls)

webknjaz: For example, https://www.python.org/dev/peps/pep-0008/#id28

Congee: Any reason it's not merged yet?

webknjaz: Just that we're all busy humans :)

PR #306, opened 2018-11-22
Fix memory leak

asvetlov:


decaz:
@asvetlov I'm afraid that the bug is not fixed, - still experiencing memory leak problems.

asvetlov:
Well, it fixes aiohttp server at least.
Thanks for letting me know.

pauleikis:
I can confirm that this does not fix aiohttp client

Olegt0rr:
Please fix memory leak
```
#1: aiohttp/client_reqrep.py:300: 40187.8 KiB
self.headers = CIMultiDict()
```

PR #321, opened 2019-01-05
Fix gc track for iterator and pair_list.

### LINKED ISSUE #320: Call PyObject_GC_Track() for views and iterators after creation
Calls like `PyObject_GC_New` doesn't track returned objects, an explicit `PyObject_GC_Track` call is needed IIRC

/cc @iemelyanov

iemelyanov:
Resolve #320 

Only iterators and pair_list contains this bug.

### DIFF
--- a/multidict/_multidict_iter.c
+++ b/multidict/_multidict_iter.c
@@ -41,6 +41,7 @@ multidict_items_iter_new(PyObject *impl)
 
     _init_iter(it, impl);
 
+    PyObject_GC_Track(it);
     return (PyObject *)it;
 }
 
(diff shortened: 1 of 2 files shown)


asvetlov:
I think views should be tracked as well

iemelyanov:
> I think views should be tracked as well

Mb I am missing something but all objects created by gc_new in views don't have this problem.

asvetlov:
LGTM, please merge

iemelyanov:
> LGTM, please merge

What I should use for merge - "Squash and merge" or "Rebase and merge"? :)

asvetlov:
Squash.
Please re-read the squashed commit message. 
If it is too long or contains non-valuable lines like `Fixed tests` etc -- edit the message before committing.

iemelyanov:
> Squash.
> Please re-read the squashed commit message.
> If it is too long or contains non-valuable lines like `Fixed tests` etc -- edit the message before committing.

Ok. I understand. Thank you.

PR #342, opened 2019-05-03
Fix TypeError with MultiDictProxy

blueyed:
Previously it would raise:

> TypeError('sequence item 0: expected str instance, type found')

Uses a fixed list/string with non-cython version: https://github.com/blueyed/multidict/blob/44e9ffd672f278499e0f4ef293291d66fd95cd03/multidict/_multidict_py.py#L171-L173

### DIFF
--- a/multidict/_multidict.pyx
+++ b/multidict/_multidict.pyx
@@ -142,11 +142,10 @@ cdef class MultiDictProxy(_Base):
         cdef _Base base
         if not isinstance(arg, self._proxy_classes):
             raise TypeError(
-                'ctor requires {} instance'
-                ', not {}'.format(
-                    ' or '.join(self._proxy_classes),
-                    type(arg)))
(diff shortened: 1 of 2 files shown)


multidict/_multidict.pyx
```diff
@@ -142,11 +142,10 @@ cdef class MultiDictProxy(_Base):
...
             raise TypeError(
-                'ctor requires {} instance'
-                ', not {}'.format(
-                    ' or '.join(self._proxy_classes),
-                    type(arg)))
-
+                "ctor requires {} instance, not {}".format(
+                    " or ".join(x.__name__ for x in self._proxy_classes), type(arg)
+                )
+            )
```
blueyed: Message can be improved maybe, let me know.

asvetlov:
Thanks!

PR #363, opened 2019-08-29
Add pickle v5 test files

### LINKED ISSUE #362: [TODO] Add Pickle v5 to the test suite
I guess it's the same as #296.

gyermolenko:
Fixes #362 

Since 3.8 is in beta and it can be important
how test files were generated:
```pycon
$ python gen_pickles.py

$ python -c "import sys; print(sys.version)"
3.8.0b3+ (heads/3.8:0d45d50e42, Aug 29 2019, 12:01:48)
```


webknjaz:
I think that beta is unlikely to change the format. There should only be some bug fixed...

webknjaz:
Let's merge this once the corresponding CI jobs pass.

PR #381, opened 2019-11-06
Integrate towncrier for the changelog generation

asvetlov:
Make it mimic the approach used by aiohttp and others.

P.S.
CI will be tuned later.


CHANGES/.TEMPLATE.rst
```diff
@@ -0,0 +1,36 @@
+{# TOWNCRIER TEMPLATE #}
```
webknjaz: I think, this folder also needs a separate README explaining what to do. GitHub will then render it in the directory view.

pyproject.toml
```diff
@@ -1,2 +1,11 @@
 [build-system]
-requires = ["pip>=18", "setuptools>=40", "wheel"]
+requires = ["pip>=19.3", "setuptools>=40", "wheel"]
```
webknjaz: maybe?
```suggestion
requires = ["pip>=19,!=19.1", "setuptools>=40", "wheel"]
```
It looks like pip 19.2 dropped support for Python 3.4.

CHANGES/.TEMPLATE.rst
asvetlov: The idea is good; the README is not mandatory though.
aiohttp doesn't have it; we have instructions in CONTRIBUTING document.
I think it can work for multidict as well.

pyproject.toml
asvetlov: I'm not sure if pip is a build system requirement at all.
`python setup.py install` works without `pip`; it requires `setuptools` only.

Maybe we can drop `[build-system]` entirely, the implicit fallback to `setuptools` "just works".

CHANGES/.TEMPLATE.rst
webknjaz: I saw it work here well: https://github.com/tox-dev/tox/tree/master/docs/changelog.

pyproject.toml
webknjaz: No. We shouldn't call the back-end directly.

asvetlov: Please elaborate, do you agree with dropping `pip` or the entire section?

asvetlov: Reverted the line; the change worth a separate PR

webknjaz: I don't think we should touch it now.

CHANGES/.TEMPLATE.rst
asvetlov: If you want to add README here -- you are welcome. In a separate PR.

asvetlov:
@webknjaz what are the `rtd` comments?
Why they are here? And why is this crappy stacktrace from chrome?

webknjaz:
@asvetlov probably because you forgot to add `rtd-bot` to the RTD project as a maintainer. This is an integration that shows previews of docs changes for PRs.

asvetlov:
I did not add this bot to the project; I have no idea what is the bot for

PR #394, opened 2019-11-21
Replace Cython with pure C

asvetlov:
Co-authored-by: Igor Emelianov <cybercjp@gmail.com>


multidict/_multidict_c.c
```diff
@@ -0,0 +1,1587 @@
...
+    }   /* sentinel */
+};
+
+static PyTypeObject multidict_proxy_type = {
+    PyVarObject_HEAD_INIT(NULL, 0)
+    "multidict._multidict.MultiDictProxy",           /* tp_name */
+    sizeof(MultiDictProxyObject),                    /* tp_basicsize */
+    0,                                               /* tp_itemsize */
+    (destructor)multidict_proxy_tp_dealloc,          /* tp_dealloc */
+    0,                                               /* tp_print */
```
iemelyanov: Mb we should change this comment and other the same comments to /* tp_vectorcall_offset */ according to https://www.python.org/dev/peps/pep-0590/#id12?

multidict/_multidict_c.c
```diff
@@ -0,0 +1,1587 @@
...
+        Py_RETURN_NOTIMPLEMENTED;
+    }
+
+    if (MultiDict_CheckExact(other) || CIMultiDict_CheckExact(other)) {
+        cmp = _multidict_eq(
+            (MultiDictObject*)self,
+            (MultiDictObject*)other
+        );
+        if (cmp < 0) {
+            // TODO: set exception
```
iemelyanov: What about this?

multidict/_multidict_c.c
```diff
@@ -0,0 +1,1587 @@
+#include "_multidict_c.h"
+#include "_pair_list.h"
+#include "_multidict_views.h"
+
+#include <structmember.h>
+
+// fix for VisualC complier used by Python 3.4
```
iemelyanov: Is this necessary if having in mind the current version of python 3.8?

asvetlov: Python 3.4 is dropped but Python 3.5 is the minimal supported version now.
It will take a long when the minimal version became 3.8

asvetlov:
Thanks for raised questions, I'll address them before merging.

@iemelyanov you have write access to the repo and this branch, just in case :)

multidict/_multidict_c.c
asvetlov: Good idea

multidict/_multidict_c.c
asvetlov: `cmp < 0` if `_multidict_eq()` returns -1.
The Python exception is set here by `PyObject_RichCompareBool()`

multidict/_multidict_c.c
asvetlov: Dropped

asvetlov:
I've dropped cythonization by the last commit but want to keep pxd files at least for a while.

Maybe we can reuse them as a part of our future provided public Cython-level API for easy using multidict from Cython extensions.
I can imagine how it can be useful for the aiohttp project.

asvetlov:
Thanks for the review!

asvetlov:
And for the vast majority of the PR's code :)

iemelyanov:
> And for the vast majority of the PR's code :)

Thank you for lots of suggestions and mentoring in this issue.

PR #395, opened 2019-11-22
Embed pair_list_t structure into MultiDict Python object

asvetlov:
It removes one indirection level.
No need for extra memalloc, plus multidict with embedded pair-list fits CPU cache a little better.

### DIFF
--- a/multidict/_pair_list.h
+++ b/multidict/_pair_list.h
@@ -9,46 +9,74 @@ extern "C" {
 #include <stdint.h>
 #include "Python.h"
 
-PyObject* pair_list_new(void);
-PyObject* ci_pair_list_new(void);
+// INTERNAL structures
 
-Py_ssize_t pair_list_len(PyObject *list);
+typedef PyObject * (*calc_identity_func)(PyObject *key);
 
-int pair_list_clear(PyObject *list);
+typedef struct pair {
+    PyObject  *identity;  // 8
+    PyObject  *key;       // 8
+    PyObject  *value;     // 8
+    Py_hash_t  hash;      // 8
+} pair_t;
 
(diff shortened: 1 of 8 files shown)


multidict/_pair_list.h
```diff
@@ -9,46 +9,74 @@ extern "C" {
...
-int pair_list_add(PyObject *list, PyObject *key, PyObject *value);
 
-int _pair_list_next(PyObject *list, Py_ssize_t *ppos,
+typedef struct pair_list {  // 40
+    Py_ssize_t  capacity;   // 8
+    Py_ssize_t  size;       // 8
+    uint64_t  version;      // 8
+    calc_identity_func calc_identity;  // 8
+    pair_t *pairs;          // 8
+} pair_list_t;
```
iemelyanov: I think it will be nice to refactoring this name. Sufix "_t" is not so good because of POSIX use it for type definitions. Mb something like this `MultiDictVec` and `MultiDictVecItem`? But if rename structs it will be nice to rename and functions too, instead `pair_list_*` to `multidict_vec_*`, what do you think?

asvetlov: This is the interesting question.
I've found that the separation of multidict PyObject class and pair_list_t plain C structure makes sense, at least for the current code stage.
Thinking more, pair list doesn't exist alone. It is a part of MultiDict or CIMultiDict struct. Maybe later we come up to idea of joining not only pair_list_t and MultiDict object structures but APIs also? In this case, the pair-list will go.

The pair list was crucial for the transition from Cython to pure C, I'm not sure if it still is needed.
Let's shake the code and see. If the list should stay I completely agree with `MultiDictVec` / `MultiDictVecItem` proposal. Let's decide later

PR #396, opened 2019-11-22
Embed the pairs buffer 

asvetlov:
To amortize the memory usage of multidicts smaller than 29 elements
I expect that only very few HTTP messages have a bigger amount of headers, 
the same for GET query arguments or multipart forms.

### DIFF
--- a/multidict/_pair_list.c
+++ b/multidict/_pair_list.c
@@ -4,7 +4,8 @@
 
 #include <Python.h>
 
-#define MIN_LIST_CAPACITY 32
+#define MIN_CAPACITY 63
+#define CAPACITY_STEP 64
 
 static PyObject * _istr_type;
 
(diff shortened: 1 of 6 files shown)


multidict/_pair_list.c
```diff
@@ -85,23 +86,68 @@ pair_list_get(pair_list_t *list, Py_ssize_t i)
...
-    if (new_capacity < MIN_LIST_CAPACITY) {
-        new_capacity = MIN_LIST_CAPACITY;
+
+    if (list->size < list->capacity) {
+        return 0;
     }
-    if (list->capacity == new_capacity) {
-        // No need to resize
+
+    if (list->pairs == list->buffer) {
```
iemelyanov: I don't understand why to compare two ptr? This looks like UB. Mb this should be something like `list->size == list->capacity`?

PR #397, opened 2019-11-22
Fix memory leak in multidict proxy destruction

asvetlov:
:D


asvetlov:
I don't want to write a test for this case.
The possible test is subtle and useless; it cannot catch the real problem and fails on PyPy.
The strategy for making sure that the library doesn't leak is running a code that used preferably the most of APIs in a cycle and monitor the process's RSS consumption.
I used simple aiohttp client + server scripts to do this.

PR #398, opened 2019-11-23
Fix dereference null pointer.

iemelyanov:
I catch segfault on python 3.7.

If I am right the problem in `multidict_proxy_tp_clear` which invoke` Py_CLEAR` and set `self-> md` to` NULL`.


asvetlov:
You are right.

PR #406, opened 2019-11-24
Use new-style struct definition

asvetlov:
For describing C Extensions types


asvetlov:
Since Python 3.4 support is gone we can use the new syntax.

asvetlov:
@iemelyanov I'll merge to unblock the future work.
Please feel free to post-comment if you'll find something.

PR #407, opened 2019-11-24
Since multidicts can be inherited they should call tp_free slot explicitly

asvetlov:


multidict/_multidict.c
```diff
@@ -589,7 +589,7 @@ multidict_tp_dealloc(MultiDictObject *self)
         PyObject_ClearWeakRefs((PyObject *)self);
     };
     pair_list_dealloc(&self->pairs);
-    PyObject_GC_Del(self);
+    Py_TYPE(self)->tp_free((PyObject *)self);
```
iemelyanov: What about adding macro `#define MultiDictTpFree(ob) (Py_TYPE(ob)->tp_free((PyObject *)ob))` ?

asvetlov: This macro will be used at most twice: once for MultiDict/CIMultiDict and once for their proxies.
Do we really need the macro for such a case?
The expanded version is better readable from my perspective.

iemelyanov: Ok, you right.

PR #409, opened 2019-11-25
Merge istr into _multidict module

asvetlov:
We need the only module for C Extension.

`_istr` used module state for saving global objects, it doesn't work well.
Maybe PEP 573 will be more convenient, I don't care now


asvetlov:
Thanks for the review!

PR #413, opened 2019-12-05
Return class from __class_getitem__ to simplify subclassing

asvetlov:
E.g.
```
class MyDict(MultiDict[str]):
    pass
```

[//]: # (rtdbot-start)

URL of RTD document: https://multidict.readthedocs.io/en/fix-class_getitem/ Documentation Status

[//]: # (rtdbot-end)

### DIFF
--- a/multidict/_multidict.c
+++ b/multidict/_multidict.c
@@ -801,7 +801,7 @@ PyDoc_STRVAR(multidict_update_doc,
 "Update the dictionary from *other*, overwriting existing keys.");
 
 static PyObject *
-multidict_class_getitem(PyObject *self, PyObject *args)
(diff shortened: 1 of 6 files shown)


multidict/_multidict.c
```diff
@@ -922,7 +923,7 @@ static PyMethodDef multidict_methods[] = {
     {
         "__class_getitem__",
         multidict_class_getitem,
-        METH_VARARGS | METH_STATIC,
+        METH_VARARGS | METH_CLASS,
```
iemelyanov: It is not so necessary but mb `METH_O | METH_CLASS`?

asvetlov: Good idea, done.

PR #420, opened 2019-12-11
Fix regression

### LINKED ISSUE #418: MultiDict class construction regression
**Describe the bug**

With multidict 4.7.0, you cannot create a MultiDict with an iterable of key-valu data.

**To Reproduce**   

```
>>> import sys
>>> import multidict
>>> multidict.__version__
'4.7.0'
>>> sys.version
'3.7.3 (default, Apr 20 2019, 14:54:00) \n[GCC 7.3.0]'
>>> headers=[['User-Agent', 'HTTP Stuff']]
# This is a valid input for the python dict() function
>>> foo = dict(headers)
>>> foo
{'User-Agent': 'HTTP Stuff'}
# Direct MultiDict use failse
>>> multidict.MultiDict(headers)
TypeError: MultiDict keys should be either str or subclasses of str
# aiohttp uses CiMultiDict
>>> multidict.CIMultiDict(headers)
TypeError: CIMultiDict keys should be either str or subclasses of str
```

**Expected behavior**

I expect to make a CIMultiDict / MultiDict with the input list of items.

**Logs/tracebacks**

Included in the above example.

**Your version of the Python**

3.7.3 compiled via pyenv on ubuntu 18.04

**Your version of the multidict distribution**

4.7.0

**Additional context**

I ran into this with testing aiohttp related calls where headers were being passed as a list of key, value pairs.

aiohttp typedefs their header arguments with a struct they call LooseHeaders defined here https://github.com/aio-libs/aiohttp/blob/v3.6.2/aiohttp/typedefs.py#L42

their docs refer to header arguments as either dicts or list of key/valu pairs https://aiohttp.readthedocs.io/en/stable/client_reference.html#client-session

That is passed into the multidict struct here https://github.com/aio-libs/aiohttp/blob/v3.6.2/aiohttp/client.py#L827

This is valid according to the multidict documentation here:

https://multidict.readthedocs.io/en/stable/multidict.html#multidict

Specifically the line `Accepted parameters are the same as for dict.`. This is no longer valid with the 4.7.0 release.

asvetlov:
multidict can be constructed from arbitrary iterable of pairs again.

Fixes #418

### DIFF
--- a/multidict/_multidict.c
+++ b/multidict/_multidict.c
@@ -135,28 +135,40 @@ _multidict_append_items_seq(MultiDictObject *self, PyObject *arg,
     }
 
     while ((item = PyIter_Next(iter)) != NULL) {
-        if (PyObject_Length(item) != 2) {
-            PyErr_Format(
-                PyExc_TypeError,
-                "%s takes either dict or list of (key, value) tuples",
-                name,
-                NULL
-            );
-            Py_DECREF(item);
-            Py_DECREF(iter);
-            return -1;
+        if (PyTuple_CheckExact(item)) {
+            if (PyTuple_GET_SIZE(item) != 2) {
+                goto invalid_type;
+            }
+            key   = PyTuple_GET_ITEM(item, 0);
+            Py_INCREF(key);
+            value = PyTuple_GET_ITEM(item, 1);
+            Py_INCREF(value);
         }
-
-        key   = PyTuple_GET_ITEM(item, 0);
-        value = PyTuple_GET_ITEM(item, 1);
+        else if (PyList_CheckExact(item)) {
+            if (PyList_GET_SIZE(item) != 2) {
+                goto invalid_type;
+            }
+            key   = PyList_GET_ITEM(item, 0);
+            Py_INCREF(key);
+            value = PyList_GET_ITEM(item, 1);
+            Py_INCREF(value);
+        }
+        else if (PySequence_Check(item)) {
+            if (PySequence_Size(item) != 2) {
+                goto invalid_type;
+            }
+            key = PySequence_GetItem(item, 0);
+            value = PySequence_GetItem(item, 1);
+        } else {
+            goto invalid_type;
+        }
 
         if (pair_list_add(&self->pairs, key, value) < 0) {
-            Py_DECREF(item);
-            Py_DECREF(iter);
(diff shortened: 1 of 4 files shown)


multidict/_multidict.c
```diff
@@ -166,6 +178,20 @@ _multidict_append_items_seq(MultiDictObject *self, PyObject *arg,
...
 
     return 0;
+invalid_type:
+    PyErr_Format(
+        PyExc_TypeError,
+        "%s takes either dict or list of (key, value) pairs",
+        name,
+        NULL
+    );
+    goto fail;
```
iemelyanov: This is not needed

multidict/_multidict.c
```diff
@@ -135,28 +135,40 @@ _multidict_append_items_seq(MultiDictObject *self, PyObject *arg,
...
-            );
-            Py_DECREF(item);
-            Py_DECREF(iter);
-            return -1;
+        if (PyTuple_CheckExact(item)) {
+            if (PyTuple_GET_SIZE(item) != 2) {
+                goto invalid_type;
+            }
+            key   = PyTuple_GET_ITEM(item, 0);
+            Py_INCREF(key);
```
iemelyanov: This is a potential memleak because we increment key and val ref in pair_list_add -> _pair_list_add_with_hash.

multidict/_multidict.c
asvetlov: Indeed, you are right.
I believe this goto increases readability a little, the C compiler optimizer will drop this instruction anyway.

multidict/_multidict.c
iemelyanov: I missed `Py_CLEAR` at the end of the loop. Ok, but I don't think that we need `Py_INCREF` and `Py_CLEAR` for key and val, `pair_list_add` do this job.

asvetlov: `key` and `value` are decrefed at the end of `while` loop, see `Py_CLEAR()` calls.

Bugs on CPython taught me that borrowed references are things that should be avoided.
That's hard but still possible to write a custom `key` class in pure Python that will crash the interpreter when used with a multidict.  Say, it can destroy a list in its own `__hash__` function.
Holding a reference until `pair_list_add` is called prevents this weird-but-still-possible bug.

Issue #432, opened 2019-12-12
multidict 4.7.1 causes segfault when bundled with PyInstaller


gmarull:
**Describe the bug**

After upgrading to `multidict` 4.7.1 we have found that the applications we bundle using PyInstaller crash. Actually, I discovered this problem because we use `aiohttp` which depends on `multidict`. Version 4.6.1 does not have this problem.

**To Reproduce**   

1. Create a new virtual environment
2. `pip install multidict==4.7.1 pyinstaller`
3. Create a simple `test.py` script with:
```
print("Hello!")
import multidict
```
4. Bundle it `pyinstaller test.py`
5. Run it:
```
./dist/test/test
Hello!
[1]    22548 segmentation fault (core dumped)  ./dist/test/test
```

**Expected behavior**

Application prints "Hello!" and quits nicely.

**Logs/tracebacks**

Provided above.

**Your version of the Python**

Observed with 3.6.9 and 3.7.5

**Your version of the multidict distribution**

4.7.1

**Additional context**

N/A

razaqq:
Im getting this with every version > 4.6.1  (4.7.0 and 4.7.1)

asvetlov:
Thanks for the report!

Varriount:
This also happens when multidict is used with [nuitka](https://nuitka.net/).

@razaqq A temporary workaround is to set the `MULTIDICT_NO_EXTENSIONS` environment variable to something before installing via pip, and to then `pip install --no-binary multidict multidict` (you might need to add `--upgrade --force-reinstall --ignore-installed` too, if the package is already installed).

Varriount:
@asvetlov GDB states that the segfault occurs at https://github.com/aio-libs/multidict/blob/master/multidict/_multidict.c#L1595 , though I don't know if that's 100% accurate, as I wasn't working with a debug build (I don't think).

asvetlov:
Decref of static type is definitely a mistake but it happens on import failing only as I see.
I wonder why the `import multidict` fails at all?

iemelyanov:
> Decref of static type is definitely a mistake but it happens on import failing only as I see.
> I wonder why the `import multidict` fails at all?

The fail happens here - https://github.com/aio-libs/multidict/blob/master/multidict/_multidict.c#L1494

asvetlov:
@iemelyanov thanks!
I'll have time to work on the fix in 2 hours.

Oberon00:
[...]

PR #433, opened 2019-12-13
Fix segfault import when multidict bundled with pyinstaller.

iemelyanov:
## What do these changes do?

Fix segfault import when multidict bundled with pyinstaller.

No

## Related issue number

https://github.com/aio-libs/multidict/issues/432

## Checklist

- [X] I think the code is well written
- [x] Add a new news fragment into the `CHANGES` folder


asvetlov:
Sorry, I don't understand how it fixes the project.
Why did it fail before?

iemelyanov:
> Sorry, I don't understand how it fixes the project.
> Why did it fail before?

Mb better split this fix on two.

1. Fix segfault which happened when failed get attribute https://github.com/aio-libs/multidict/blob/master/multidict/_multidict.c#L1494 by deleting `decrefs` of static objects.
2. Fix incorrect get attribute.

One interesting thing is that `multidict` correct imported and work event if internally happened `fail` and a big part of code in `PyInit__multidict` not executed!

asvetlov:
Things that bother me:
1. `_mdrepr` is imported from `multidict._multidict_base` on the current master, why additional import in `__init__.py` is needed?
2. If the import of `multidict._multidict` module fails `PyInit__multidict` returns `NULL`, the module object is even not created. Why `pyinstaller` things that the import was succeded. I expect some kind of `ImportError` exception raising.

In conclusion, `pyinstaller`'s import machinery looks buggy at least partially. 
I would know what part is broken. For now, the fix looks like making really very weird things to slip between pyinstaller bugs.
Please correct me if I'm wrong.

iemelyanov:
> Things that bother me:
> 
> 1. `_mdrepr` is imported from `multidict._multidict_base` on the current master, why additional import in `__init__.py` is needed?
> 2. If the import of `multidict._multidict` module fails `PyInit__multidict` returns `NULL`, the module object is even not created. Why `pyinstaller` things that the import was succeded. I expect some kind of `ImportError` exception raising.
> 
> In conclusion, `pyinstaller`'s import machinery looks buggy at least partially.
> I would know what part is broken. For now, the fix looks like making really very weird things to slip between pyinstaller bugs.
> Please correct me if I'm wrong.

You right. For me, this `fails` looks strange too and mb we should more deeply investigate this issue.

iemelyanov:
I propose to close this PR and live #433 open until we don't understand what's going on in `pyinstaller` and why `import multidict` work event if `fail` happen here - https://github.com/aio-libs/multidict/blob/master/multidict/_multidict.c#L1494. What do you think?

asvetlov:
Agree

PR #438, opened 2019-12-14
Slightly reorganize initialization code

asvetlov:
Curiously, #432 problem seems gone. 
I don't understand clearly why but the code doesn't contain explicit tricks, it is clearly readable at least.

[//]: # (rtdbot-start)

URL of RTD document: https://multidict.readthedocs.io/en/cleanup/ Documentation Status

[//]: # (rtdbot-end)

### DIFF
--- a/CHANGES/432.bugfix
+++ b/CHANGES/432.bugfix
@@ -1,0 +1,1 @@
+Fix crashing when multidict is used pyinstaller
--- a/docs/spelling_wordlist.txt
+++ b/docs/spelling_wordlist.txt
@@ -60,6 +60,7 @@ pickable
 pre
 proxied
 pyenv
+pyinstaller
 refactor
 refactored
 regex
--- a/multidict/_multidict.c
+++ b/multidict/_multidict.c
@@ -13,7 +13,6 @@
 static PyObject *collections_abc_mapping;
 static PyObject *collections_abc_mut_mapping;
 static PyObject *collections_abc_mut_multi_mapping;
-static PyObject *istr;
 
 static PyTypeObject multidict_type;
 static PyTypeObject cimultidict_type;
@@ -1436,7 +1435,6 @@ getversion(PyObject *self, PyObject *md)
 static void
 module_free(void *m)
 {
-    Py_CLEAR(istr);
     Py_CLEAR(collections_abc_mapping);
     Py_CLEAR(collections_abc_mut_mapping);
     Py_CLEAR(collections_abc_mut_multi_mapping);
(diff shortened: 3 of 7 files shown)


asvetlov:
Hmm, actually it doesn't fix but masks a problem.
Import error is raised and the system switches to pure python under pyinstaller.
The difference is that import doesn't crash now.

Need to investigate what behavior was in previous multidict versions? Maybe pyinstaller was never good with the multidict?

iemelyanov:
> Curiously, #432 problem seems gone.
> I don't understand clearly why but the code doesn't contain explicit tricks, it is clearly readable at least.
> 
> URL of RTD document: https://multidict.readthedocs.io/en/cleanup/ Documentation Status

It looks like everything ok. But `fail` still happens and now here - https://github.com/aio-libs/multidict/blob/cleanup/multidict/_multidict.c#L1482. Mb this is `pyinstaller` issue and not important for us.

iemelyanov:
> Hmm, actually it doesn't fix but masks a problem.
> Import error is raised and the system switches to pure python under pyinstaller.
> The difference is that import doesn't crash now.
> 
> Need to investigate what behavior was in previous multidict versions? Maybe pyinstaller was never good with the multidict?

Previous versions v4.6.1 and lower works correct because not use C-extension.

Varriount:
@asvetlov The fact that #432 is gone for no perceivable reason is worrying. To me, that indicates that there was some form of undefined behavior causing trouble. Do tests run with memory sanitizer/valgrind enabled?

asvetlov:
@Varriount as I see the reason for the error is: `pyinstaller` splits `multidict` in two locations: Python files are bundled with all other sources into executable image and multidict folder with compiled shared library.
When the shared library tries to import `multidict._multidict_base` python module the import fails because of the split. Seems like `pyinstaller` supports a bunch of hooks for popular libraries to avoid this problem but I didn't dig into the project too deeply.

The PR "fixes" the problem to correctly process the import failure and switching back to pure-python implementation (which is much slower but still works).

Also, #443 has fixed another crash unrelated to this issue.

asvetlov:
@Varriount as I see the reason for the error is: `pyinstaller` splits `multidict` in two locations: Python files are bundled with all other sources into executable image and multidict folder with compiled shared library.
When the shared library tries to import `multidict._multidict_base` python module the import fails because of the split. Seems like `pyinstaller` supports a bunch of hooks for popular libraries to avoid this problem but I didn't dig into the project too deeply.

The PR "fixes" the problem to correctly process the import failure and switching back to pure-python implementation (which is much slower but still works).

Also, #443 has fixed another crash unrelated to this issue.

asvetlov:
Inlining pair_list adds 15-25% boost.
Other inlines seem don't add anything but I'll keep it for the sake of consistency.

PR #449, opened 2019-12-24
Implement __sizeof__ function

### LINKED ISSUE #444: Implement __sizeof__ function to correctly calculate all internal structures size.
See also https://docs.python.org/3/library/sys.html#sys.getsizeof

asvetlov:
Fixes #444

### DIFF
--- a/multidict/_multidict.c
+++ b/multidict/_multidict.c
@@ -869,12 +869,11 @@ PyDoc_STRVAR(sizeof__doc__,
 "D.__sizeof__() -> size of D in memory, in bytes");
 
 static inline PyObject *
-_multidict_sizeof(PyObject *self)
+_multidict_sizeof(MultiDictObject *self)
 {
-    MultiDictObject *md = (MultiDictObject *)self;
     Py_ssize_t size = sizeof(MultiDictObject);
-    if (md->pairs.pairs != md->pairs.buffer) {
-        size += sizeof(pair_t) * md->pairs.capacity;
+    if (self->pairs.pairs != self->pairs.buffer) {
+        size += (Py_ssize_t)sizeof(pair_t) * self->pairs.capacity;
     }
     return PyLong_FromSsize_t(size);
 }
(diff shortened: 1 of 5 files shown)


multidict/_multidict.c
```diff
@@ -982,6 +999,12 @@ static PyMethodDef multidict_methods[] = {
         METH_O | METH_CLASS,
         NULL
     },
+    {
+        "__sizeof__",
+        _multidict_sizeof,
```
iemelyanov: `(PyCFunction)_multidict_sizeof` more explicit

multidict/_multidict.c
```diff
@@ -863,6 +864,22 @@ multidict_class_getitem(PyObject *self, PyObject *arg)
     return self;
 }
 
+
+PyDoc_STRVAR(sizeof__doc__,
+"D.__sizeof__() -> size of D in memory, in bytes");
+
+static inline PyObject *
```
iemelyanov: `inline` is not necessary

multidict/_multidict.c
```diff
@@ -863,6 +864,22 @@ multidict_class_getitem(PyObject *self, PyObject *arg)
...
+_multidict_sizeof(PyObject *self)
```
iemelyanov: Mb more correct `_multidict_sizeof(PyObject *self, PyObject *args)` according to `PyCFunction` and `PyMethodDef` ?

multidict/_multidict.c
asvetlov: Yes, I know.
I've converted all functions to inlines after adding `_multilib`, let's keep this style at least for a while. It doesn't hurt.

multidict/_multidict.c
asvetlov: We use `METH_O` convention in other places.
I've updated the PR to accept `MultiDictObject * self` as an argument.

asvetlov:
Thanks for the review!

PR #450, opened 2019-12-30
Fix crashes in popone/popall when default is returned

asvetlov:
The returned value was not increfed.


asvetlov:
@iemelyanov thanks for the review!

PR #453, opened 2020-01-10
Fix keys iterator memory leak

### LINKED ISSUE #452: Multidict 4.7.3 memory leak
Seems that multidict 4.7.x is leaking, that only happens when I iterate over multidict instance

Please check following code snippet:
```python
import gc
import multidict
import objgraph

print("Multidict:", multidict.__version__)
objgraph.growth() # save current objects state

list(multidict.MultiDict((("arb", "value"), ))) # iter over keys
print("Unreleased objects:")
objgraph.show_growth()
```

```
Multidict: 4.7.3
Unreleased objects:
_KeysView        1        +1
MultiDict        1        +1
```
On the other hand for  4.6.1
```
Multidict: 4.6.1
Unreleased objects:
```

**Your version of the Python**
3.8.1

**Your version of the multidict distribution**
4.7.3

**Snippet without objgraph package**
```python
import gc
import multidict

print("Multidict:", multidict.__version__)

gc.collect()
before = len(gc.get_objects())
for i in range(5):
    list(multidict.MultiDict((("arb", "value"),)))
    gc.collect()
    print("Unreleased:", len(gc.get_objects())-before)
```

```
Multidict: 4.7.3
Unreleased: 1
Unreleased: 3
Unreleased: 5
Unreleased: 7
Unreleased: 9
```

Reskov:
## What do these changes do?

Use directly iterator instead of `multidict_keysview_new`

None

## Related issue number

Fixes #452 

## Checklist

- [x] Add a new news fragment into the `CHANGES` folder

### DIFF
--- a/multidict/_multidict.c
+++ b/multidict/_multidict.c
@@ -572,7 +572,7 @@ multidict_sq_contains(MultiDictObject *self, PyObject *key)
 static inline PyObject *
 multidict_tp_iter(MultiDictObject *self)
 {
-    return PyObject_GetIter(multidict_keysview_new((PyObject*)self));
+    return multidict_keys_iter_new(self);
 }
 
 static inline PyObject *
--- a/CHANGES/452.bugfix
+++ b/CHANGES/452.bugfix
@@ -1,0 +1,1 @@
+``MultiDict.iter`` fix memory leak when used iterator over `multidict` instance.


multidict/_multidict.c
```diff
@@ -572,7 +572,7 @@ multidict_sq_contains(MultiDictObject *self, PyObject *key)
 static inline PyObject *
 multidict_tp_iter(MultiDictObject *self)
 {
-    return PyObject_GetIter(multidict_keysview_new((PyObject*)self));
```
webknjaz: @Reskov could you please explain the implications of dropping `PyObject_GetIter()`?

Reskov: @webknjaz Suddenly, I'm not familiar with python c-api, as far as I get `PyObject_GetIter` returns new iterator from the object `multidict_keysview_new`, which is actually equivalent to call multidict_keys_iter_new.

So in other words, I just eliminate unnecessary creation of the keysview structure, but result of the call should be the same therefore shouldn't have any impact.

Also my code-snippet from the issue now shows, that fix helps
```
Multidict: 4.7.3
Unreleased objects:
```

P.S. anyway I'm waiting for the review, because I'm not sure that I was right.

webknjaz: That's probably because you don't create an object. Usually the trick is to deallocate objects after the usage. If I were to guess I'd say look for a missing decref

webknjaz: @asvetlov @gyermolenko should a Py_DECREF be added somewhere here https://github.com/aio-libs/multidict/blob/dda85cf/multidict/_multidict.c#L637 instead?

asvetlov:
@Reskov thanks for the patch, it is 100% correct.
Another alternative is adding decref:
```
static inline PyObject *
multidict_tp_iter(MultiDictObject *self)
{
    PyObject *view = multidict_keysview_new((PyObject*)self);
    if (view == NULL) {
        return NULL;
    }
    PyObject* ret = PyObject_GetIter(view);
    Py_DECREF(view);  // missed decref
    return ret;
}
```

You version avoids the creation of temporary *keys view* object, that's why I prefer it.

Issue #500, opened 2020-07-03
Wheel support for linux aarch64


odidev:
**Summary**
Installing multidict on aarch64 via pip using command "pip3 install multidict" tries to build wheel from source code

**Problem description**
multidict don't have wheel for aarch64 on PyPI repository. So, while installing multidict via pip on aarch64, pip builds wheel for same resulting in it takes more time to install multidict. Making wheel available for aarch64 will benefit aarch64 users by minimizing multidict installation time.

**Expected Output**
Pip should be able to download multidict wheel from PyPI repository rather than building it from source code.

@multidict-taem, please let me know if I can help you building wheel/uploading to PyPI repository. I am curious to make multidict wheel available for aarch64. It will be a great opportunity for me to work with you.

webknjaz:
Hello, there's no standard for such wheels so it's not technically possible to put them on PyPI. It only accepts manylinux wheels.

odidev:
@webknjaz Manylinux2014 has given Linux AArch64 support. I have successfully build manylinux2014 wheel for same platform. I can help in this activity.

Also please reopen the issue so that we can work on this and let me know if you have any questions.

Thanks

webknjaz:
TIL

webknjaz:
This will probably require changing https://github.com/aio-libs/multidict/blob/master/.azure-pipelines/deploy.yml

webknjaz:
The only concern I have is that it's probably hard to cross-compile aarch64 in Azure.

webknjaz:
I'm curious if https://www.worksonarm.com/projects/ would help with this.

odidev:
@webknjaz We can use qemu to run an aarch64 build on x86 machine. Please let me know your thoughts here.

webknjaz:
If you can make it work, send a PR and I'll review it.

odidev:
Thanks @webknjaz I will work on it and update you.

webknjaz:
I have one extra note on this. Adding a new platform implies increasing the maintenance burden. Please do not expect the maintainers to invest too much effort into this if it starts failing in the future.

If you want to ensure that it'll keep working, you'll need to commit to contributing some effort should any related problems arise.

odidev:
sure @webknjaz, I will be taking care of the failures in future. Thanks.

odidev:
Hi @webknjaz ,

I have added support for azure pipeline for aarch64 using qemu. I am successfully able to run the test cases.
[...]

PR #519, opened 2020-08-18
Azure pipeline support for arm64 using qemu

odidev:
Added azure pipeline support for arm64 using qemu.

Related to issue number #500

### DIFF
--- a/.azure-pipelines/ci.yml
+++ b/.azure-pipelines/ci.yml
@@ -17,6 +17,19 @@ pr:
 variables:
 - group: codecov
 
+resources:
+  containers:
+  - container: manylinux
+    image: quay.io/pypa/manylinux1_x86_64
+  repositories:
+  - repository: templates
+    type: github
+    name: aio-libs/azure-pipelines
+    endpoint: aio-libs
+
 stages:
 - template: stage-lint.yml
 - template: stage-test.yml
+- template: templates/stage-test-aarch64.yml@templates
+  parameters:
+    steps: $PYTHON setup.py install && $PYTHON -m pip install -r requirements/pytest.txt && \
--- a/.azure-pipelines/deploy.yml
+++ b/.azure-pipelines/deploy.yml
@@ -25,6 +25,10 @@ stages:
 
 - template: stage-test.yml
 
+- template: templates/stage-test-aarch64.yml@templates
+  parameters:
+    steps: $PYTHON setup.py install && $PYTHON -m pip install -r requirements/pytest.txt && \
+
 - stage: build
   displayName: 'Build'
 
@@ -150,6 +154,8 @@ stages:
             - script: |
                 python setup.py bdist_wheel
               displayName: 'Make wheel'
+
+- template: templates/stage-deploy-aarch64.yml@templates
 
 - template: templates/stage-publish.yml@templates
   parameters:


.azure-pipelines/ci.yml
```diff
@@ -20,3 +20,4 @@ variables:
 stages:
 - template: stage-lint.yml
 - template: stage-test.yml
+- template: stage-aarch64.yml
```
webknjaz, line 23: this should probably go to `deploy.yml`

.azure-pipelines/stage-aarch64.yml
```diff
@@ -0,0 +1,45 @@
+stages:
```
webknjaz, line 1: this should go to https://github.com/aio-libs/azure-pipelines

.azure-pipelines/scripts/aarch64_test_steps.sh
```diff
@@ -0,0 +1,70 @@
+echo "exporting Mutlidict root directory"
```
webknjaz, line 1: this should probably go to https://github.com/aio-libs/azure-pipelines too

asvetlov:
I agree with @webknjaz 

BTW, looks like composite actions (https://docs.github.com/en/actions/creating-actions/creating-a-composite-run-steps-action) can be used as a functional replacement for azure templates if we move to github workflows eventually.
This is not related to the issue though.

webknjaz:
Oh, nice, I haven't heard of the composite actions yet!

asvetlov:
On my job I currently working with workflows a lot; recently learned by composite actions by accident.
I have no idea how long ago they were added.

webknjaz:
https://github.blog/changelog/2020-08-07-github-actions-composite-run-steps/

.azure-pipelines/ci.yml
odidev: @webknjaz Sorry for the confusion. stage-aarch64.yml(renamed as stage-test-aarch64) is for running pytest for aarch64, same as stage-test.yml is for x86, Mac and Windows.. I have renamed it to stage-test-aarch64 to avoid the confusion.
I am running this test in deploy.yml as well.

.azure-pipelines/stage-aarch64.yml
odidev: As mentioned above stage-aarch64.yml(renamed as stage-test-aarch64) is for running pytest for aarch64, same as stage-test.yml is for x86, Mac and Windows.
In my opinion it should be kept in same path as stage-test.yml. Please give your feedback and suggestions regarging this.

.azure-pipelines/scripts/aarch64_test_steps.sh
odidev: I have removed the script and accumulated all the changes in stage-test-aarch64.yml itself.

webknjaz: This doesn't change the need to have it in a shared repo rather than copy-pasted all around the place.

webknjaz (changes requested):
As per earlier comments, this should be implemented in the shared templates @ https://github.com/aio-libs/azure-pipelines and only referred from here.

mattsplats:
@webknjaz Looks like odidev has followed up on your request: https://github.com/aio-libs/azure-pipelines/pull/1

Tests are failing on this PR because the pipelines repo PR has not merged yet. Could you take a look at the pipeline change?

asvetlov:
Done by multidict 5.0.0 release

Thanks for the PR, I've borrowed many implementation details from it while building my own version

Issue #620, opened 2021-08-26
potential pypy mutability bug


iscai-msft:
### Describe the bug

when using `pypy` and mutating a `CIMultiDict`, I get a `RuntimeError` of `"Dictionary changed during iteration"`. This doesn't happen using regular python, and I don't see this error when using `pypy` and other kinds of dictionaries.

### To Reproduce

I've added the following tests to repro with `pypy`

```python
@pytest.mark.asyncio
async def test_headers_response_items_multidict_compare():
    h = CIMultiDict({'a': '123, 456', 'b': '789'})
    before_mutation_items = h.items()
    h['c'] = '000'
    with pytest.raises(RuntimeError):
        list(before_mutation_items)

def test_headers_response_items_requests_compare():
    from requests.structures import CaseInsensitiveDict
    h = CaseInsensitiveDict({'a': '123, 456', 'b': '789'})
    before_mutation_items = h.items()
    h['c'] = '000'
    list(before_mutation_items)

def test_headers_response_items_python_compare():
    h = {'a': '123, 456', 'b': '789'}
    before_mutation_items = h.items()
    h['c'] = '000'
    list(before_mutation_items)
```

### Expected behavior

My thinking is that if mutating the dictionary doesn't raise for regular python is that it shouldn't raise for `pypy`. Is this expected behavior?

### Logs/tracebacks

```python-traceback
self = _ItemsView('a': '123, 456', 'b': '789', 'c': '000')

    def _iter(self):
        for i, k, v in self._impl._items:
            if self._version != self._impl._version:
>               raise RuntimeError("Dictionary changed during iteration")
E               RuntimeError: Dictionary changed during iteration

envs/corepypy3venv/site-packages/multidict/_multidict_py.py:459: RuntimeError
```

### Python Version

```console
$ python --version
Python 3.7.10 (51efa818fd9b24f625078c65e8e2f6a5ac24d572, Apr 08 2021, 17:43:00)
[PyPy 7.3.4 with GCC Apple LLVM 12.0.0 (clang-1200.0.32.29)]
```

### aiohttp Version

```console
$ python -m pip show aiohttp
Version: 3.7.4.post0
```

### multidict Version

```console
$ python -m pip show multidict
Version: 5.1.0
```

### yarl Version

```console
$ python -m pip show yarl
Version: 1.6.3
```

### OS

macOS

### Related component

Client

### Code of Conduct

- [X] I agree to follow the aio-libs Code of Conduct

[...]

PR #627, opened 2021-09-22
Add test for incorrect mutation error

iscai-msft:
## What do these changes do?

I'm reproing the pypy issue I found here #620 

No

## Related issue number

# 620

## Checklist

- [x ] I think the code is well written
- [ x] Unit tests for the changes exist
- [ x] Documentation reflects the changes
- [x ] Add a new news fragment into the `CHANGES` folder

### DIFF
--- a/tests/test_multidict.py
+++ b/tests/test_multidict.py
@@ -517,7 +517,8 @@ class TestCIMultiDict(BaseMultiDictTest):
         assert repr(d.values()) == "_ValuesView('value1', 'value2')"
 
     def test_mutate_multidict(self, cls):
-        d = cls({'a': '123, 456', 'b': '789'})
+        d = cls({"a": "123, 456", "b": "789"})
         before_mutation_items = d.items()
-        d['c'] = '000'
-        list(before_mutation_items)  # this will raise a RuntimeError when running with pypy
+        d["c"] = "000"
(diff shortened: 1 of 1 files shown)


Dreamsorcerer:
Probably needs to go in test_mutable_multidict.py, as half the test iterations are failing due to immutable classes being used.

However, it looks like this test fails on all platforms, not just pypy.

Of the 4 iterations, only the `multidict._multidict-CIMultiDict` class actually passes and allows this behaviour. I see that this one is skipped on pypy, so that class is probably not supported on pypy. But, on all platforms the `multidict._multidict_py-CIMultiDict` version fails here.

Not sure what the correct behaviour is, but it looks like the C optimised version allows this mutation, while the python version does not. Presumably the C optimised version is used by default on CPython, but is not supported on pypy.

Dreamsorcerer:
A regular `dict` does support this behaviour:
```
>>> d = {"a": "123, 456", "b": "789"}
>>> before_mutation_items = d.items()
>>> d["c"] = "000"
>>> list(before_mutation_items)
[('a', '123, 456'), ('b', '789'), ('c', '000')]
```

So, I'd say the Python version of multidict should probably support this as well.

asvetlov:
Thanks a lot for the report and test.
#628 should fix the bug

PR #629, opened 2021-09-30
Add support for Python 3.10 and `musl`-based runtimes

asvetlov:


webknjaz (approved):

CHANGES/629.feature
```diff
@@ -0,0 +1 @@
+Support Python 3.10, build Linux Alpine images
```
webknjaz, line 1: I think this could be improved as follows:
```suggestion
Added support for Python 3.10 and started shipping platform-specific wheels with the ``musl`` tag targeting typical Alpine Linux runtimes.
```

asvetlov: Thanks

asvetlov:
@webknjaz I've switched our build system to cibuildwheel, works like a charm (after 10+ CI script fixes).
My current plan is: merge the PR tomorrow, make another yet alpha release, wait for success, publish the final release.
The follow-up steps are applying a similar procedure to yarl and aiohttp itself.
Python 3.10 is coming, we should be prepared.

Regarding aiohttp: we can publish 3.8 branch without any critical blocker, isn't it? 

/cc @Dreamsorcerer

.github/workflows/ci.yml
```diff
@@ -134,10 +134,10 @@ jobs:
     steps:
     - name: Checkout
       uses: actions/checkout@v2.3.4
-    - name: Setup Python 3.8
+    - name: Setup Python 3.9
       uses: actions/setup-python@v2.2.2
       with:
-        python-version: 3.8
+        python-version: 3.9
```
Dreamsorcerer, line 140: I noticed it is possible to do `'3.x'`, which would save needing to keep updating this one.

Dreamsorcerer:
> Regarding aiohttp: we can publish 3.8 branch without any critical blocker, isn't it?

I think we should be good. We recently fixed a couple of missing deprecation errors, so it should be ready for Python 3.10 now.
I have a couple of bug fix PRs open that might be worth merging before the release (https://github.com/aio-libs/aiohttp/pull/5853, https://github.com/aio-libs/aiohttp/pull/5328, maybe https://github.com/aio-libs/aiohttp/pull/5393, but still needs the test to be cleaned up).

.github/workflows/ci.yml
webknjaz: FWIW 3.x is probably the default so the whole `with:` could be dropped. By the way, the whole Composite GitHub Actions thing got a bit better and almost usable recently so maybe we'll be able to even deduplicate the maintenance efforts (I've been experimenting with this and other CI improvements over the past year, hope to apply some in aio-libs as well).

webknjaz:
> @webknjaz I've switched our build system to cibuildwheel, works like a charm (after 10+ CI script fixes).

Yep. I saw that. Actually, I've been meaning to try it out and have been collaborating with the maintainers a few times in different corners of GH. I think it's great to have. Hope to see it improved even more over time, with Composite Actions that can now invoke other actions.
And great job picking up the recently available `musl` tag, I've been meaning to start integrating it into the projects too.

> My current plan is: merge the PR tomorrow, make another yet alpha release, wait for success, publish the final release. The follow-up steps are applying a similar procedure to yarl and aiohttp itself. Python 3.10 is coming, we should be prepared.
> 
> Regarding aiohttp: we can publish 3.8 branch without any critical blocker, isn't it?

I haven't checked the automation side of it since I've addressed the CVE in the 3.7 stream. But it should be fine. Except, maybe, we need to check if Towncrier still works as expected — after its recent major release, I've faced a regression in other projects.

@asvetlov

There is one blocker, though, that is important for my employer. I've even been collaborating with a few people across the company to get the TLS-in-TLS support in aiohttp. My plan was to polish + https://github.com/aio-libs/aiohttp/pull/5992, include this feature in v3.8, and publish it soon (I was actually going to do it around now or in a few days). It'll probably need some warnings in place because of the need to patch CPython and I haven't yet decided on whether to allow a hack in aiohttp directly or try to make it possible to patch it externally — you could actually help with it by seeing https://github.com/python/cpython/pull/28073 through. The team/project should've applied this patch in production today AFAIK but there's a need to get it into upstream properly. They did manual testing but we need to integrate it into our CI — the attempts to do this are currently failing (https://github.com/aio-libs/aiohttp/pull/6002), I haven't figured out why but my suspicion is that it may have something to do with the fact that the proxy library we're trying to integrate is synchronous and it probably needs to be executed in a separate thread :man_shrugging:.

webknjaz:
@asvetlov since you're already refactoring some of the publishing aspects, I'll add a few improvements to this branch too:
1) GHA for publishing that I've created under PyPA
2) Release approvals and restricted access to the publishing token via per-env secrets

.github/workflows/ci.yml
```diff
@@ -147,97 +147,52 @@ jobs:
...
+    - uses: actions/upload-artifact@v2
       with:
         name: dist
-        path: dist
+        path: ./wheelhouse/*.whl
 
   deploy:
     name: Deploy
-    needs: [build-linux, build-binary, build-tarball]
+    environment: release
```
webknjaz, line 184: @asvetlov with this, you'll need to approve the job for it to run. Try it out before you merge. Start the release flow as usual (by pushing a tag) and you should get a notification to approve when the workflow gets to this place. I've also set a 15-minute cooldown period so that accidental releases could be canceled if anything arises.

.github/workflows/ci.yml
asvetlov: Done, thanks for the suggestions!

.github/workflows/ci.yml
asvetlov: Thanks, I'm willing to try cooldown+approval (not sure if these steps are necessary, maybe because of the lack of experience).
Per-env secrets are definitely the thing!

asvetlov:
Cool!
I've marked mentioned aiohttp PRs with '3.8' milestone to don't miss them.

Regarding release -- I'd like to reuse https://github.com/aio-libs/create-release action for publishing nice-looking GitHub release notes generated from towncrier changelog. The action is usable now. Will do it in a separate PR.

.github/workflows/ci.yml
webknjaz: Yep, it's an extra level of security they've added and I've only started using it recently.

PR #630, opened 2021-10-01
Use aio-libs/create-release action

asvetlov:


CHANGES/.gitignore
```diff
@@ -0,0 +1 @@
+!.gitignore
```
webknjaz, line 1: By the way, how about adding a more useful file, like README? It'd show up whenever the users open this dir in the GH web UI. So it's nice to have some instructions.
For example, here's what I've added for setuptools: https://github.com/pypa/setuptools/tree/main/changelog.d#adding-change-notes-with-your-prs

(as a bonus, you may `.. include::` it into the sphinx docs as well: https://setuptools.pypa.io/en/latest/development/developer-guide.html#adding-change-notes-with-your-prs)

asvetlov: Good idea!

.github/workflows/ci.yml
```diff
@@ -180,14 +180,14 @@ jobs:
...
-    - name: Publish a Python distribution to PyPI
-      uses: pypa/gh-action-pypi-publish@release/v1
-      with:
-        password: ${{ secrets.PYPI_API_TOKEN }}
+        changes_file: CHANGES.rst
+        name: multidict
+        github_token: ${{ secrets.GITHUB_TOKEN }}
+        pypi_token: ${{ secrets.PYPI_API_TOKEN }}
+        artifact: dist
+        version_file: multidict/__init__.py
```
webknjaz, line 191: Since the code is not fetched, this file is not accessible by the action. Why is it needed anyway? They artifacts should have a version by this point. You could extract the version from the artifacts if necessary.

asvetlov: Correct. Let me refactor the action

CHANGES/README.rst
```diff
@@ -0,0 +1,86 @@
...
+
+.. code-block:: rst
+
+    Droped Python 3.5 support; Python 3.6 is the minimal supported Python version.
+
+File :file:`CHANGES/550.bugfix.rst`:
+
+.. code-block:: rst
+
+    Provided x86 Windows wheels
```
webknjaz, line 76:
```suggestion
    Started shipping Windows wheels for the x86 architecture.
```

CHANGES/README.rst
```diff
@@ -0,0 +1,86 @@
...
+
+File :file:`CHANGES/553.feature.rst`:
+
+.. code-block:: rst
+
+    Supported ``GenericAliases`` (``MultiDict[str]``) for Python 3.9+
```
webknjaz, line 82:
```suggestion
    Added support for ``GenericAliases`` (``MultiDict[str]``) under Python 3.9 and higher.
```

.github/workflows/ci.yml
asvetlov: I spent about a day trying to fix the release action without success. Got some strange errors.
Let's merge as-is, please feel free to refactor later.
I'm personally ok with reading the version from `__init__.py`; the action needs repo clone to read `CHANGES.rst` content anyway.
I consider this step as "write and don't touch" and prefer investing my spare time in more urgent tasks, e.g. publishing yarl/aiohttp binary wheels for Python 3.10

Thanks for your understanding.

webknjaz: Looks like this one succeeded: https://github.com/aio-libs/multidict/actions/runs/1300483727. What sort of errors are you observing? Or are those the action-related ones?
FWIW I agree that it's not blocking for this specific PR. Feel free to mark this conversation as resolved and merge the PR.
If you have a minute, just make sure to document the problems somewhere, and maybe I'll be able to look at those later, after I address some of the aiohttp blockers.

PR #677, opened 2022-01-19
Fixed type annotations for keys of multidict mapping classes

q0w:
ref: 
https://github.com/python/mypy/issues/8477

## Related issue number
Fixes #644 

## Checklist

- [x] I think the code is well written
- [x] Unit tests for the changes exist
- [ ] Documentation reflects the changes
- [x] Add a new news fragment into the `CHANGES` folder
  * name it `<issue_id>.<type>` (e.g. `588.bugfix`)
  * if you don't have an `issue_id` change it to the pr id after creating the PR
  * ensure type is one of the following:
    * `.feature`: Signifying a new feature.
    * `.bugfix`: Signifying a bug fix.
    * `.doc`: Signifying a documentation improvement.
    * `.removal`: Signifying a deprecation or removal of public API.
    * `.misc`: A ticket has been closed, but it is not of interest to users.
  * Make sure to use full sentences with correct case and punctuation, for example: `Fix issue with non-ascii contents in doctest text files.`

PR #682, opened 2022-01-23
Drop wrong key parametrization for concrete classes

asvetlov:
https://github.com/aio-libs/multidict/pull/677 added generic parameters for keys in accepted mappings and abstract multimapping classes. This is correct.
But the pep also incorrectly added generic key param for concrete classes, e.g. `MultiDict`. This is wrong, the key type of `MultiDict` is always `str` by definition (while the container can accept any str-like keys in its methods, including `istr`).
The same is true for `CIMultiDict`, the key type is always `istr` while any str-like key is accepted and converted internally.


multidict/__init__.pyi
```diff
@@ -63,9 +63,9 @@ class MutableMultiMapping(
     @abc.abstractmethod
     def popall(self, key: _S, default: _D) -> Union[List[_T], _D]: ...
 
-class MultiDict(MutableMultiMapping[_S, _T], Generic[_S, _T]):
+class MultiDict(MutableMultiMapping[str, _T]):
     def __init__(self, arg: _Arg[_S, _T] = ..., **kwargs: _T) -> None: ...
```
Dreamsorcerer, line 67: I think this fails due to the `_S`. mypy now expects only 1 type in the annotation, but also wants 2 types to be defined (_S, _T), which appears to be impossible.

PR #684, opened 2022-01-23
Setup mypy for regression checks

Dreamsorcerer:
Multidict 6.0.0 and 6.0.1 have messed up typing.

This adds type checking to atleast the test_multidict.py file to check for regressions in future.

Still needs regressions to be fixed though, in order for the test to pass.


asvetlov (dismissed):
Thanks for the PR.
The current master (and 6.0.2) uses a different approach: it enumerates all possible calls in `tests/test_mypy.py` and runs it in `--strict` mode.

Adding strict annotations to other tests (and maybe `multidict/_*.py` files) makes sense to me. Please feel free to update the PR in this manner.
If you have no capacity to convert all files -- it's fine, please fix the merge conflict at least.

.mypy.ini
```diff
@@ -0,0 +1,26 @@
+[mypy]
+files = tests/test_multidict.py, tests/test_mypy.py
+check_untyped_defs = True
+disable_error_code = comparison-overlap  # Causes comparison errors with Dict.
```
Dreamsorcerer, line 4: Not sure if there is something we can do to improve this (maybe inherit from `dict`?). But, throughout the tests I get errors when comparing with dicts, like: `Non-overlapping equality check (left operand type: "Union[MultiDict[str], CIMultiDict[str]]", right operand type: "Dict[str, str]")`

.mypy.ini
```diff
@@ -0,0 +1,26 @@
+[mypy]
+files = tests/test_multidict.py, tests/test_mypy.py
```
Dreamsorcerer, line 2: We can expand this to other files later. But, the previous mypy check was not checking anything anyway.

.mypy.ini
```diff
@@ -0,0 +1,27 @@
...
+disallow_any_unimported = True
+disallow_incomplete_defs = True
+disallow_subclassing_any = True
+disallow_untyped_calls = True
+disallow_untyped_decorators = True
+disallow_untyped_defs = True
+follow_imports_for_stubs = True
+implicit_reexport = False
+no_implicit_optional = True
+show_error_codes = True
```
webknjaz, line 17: also show line numbers?

Dreamsorcerer: That's not a config option. :P
It always shows line numbers, otherwise it'd be pretty useless..

webknjaz: Also, looks like with https://github.com/python/mypy/pull/8192, there's a possibility to have `strict = True` in the config now (additionally, not instead of the individual toggles).

webknjaz (approved):
@Dreamsorcerer feel free to merge once the requests are addressed.

.mypy.ini
webknjaz: @Dreamsorcerer I learned that you can also have `enable_error_code = ["ignore-without-code"]` to prevent wholesale ignores.

webknjaz:
@Dreamsorcerer I think that we should get this in so that it's merged by the time we make a new release (I plan on doing this once Python 3.11 final is out).

.mypy.ini
Dreamsorcerer: color_output, error_summary default to True (not that I've ever seen any colour in the output...

Would be really useful if we could have a global mypy config shared between all our repos with local overrides...

.mypy.ini
Dreamsorcerer: @webknjaz Not sure if you want to try addressing this in the next release or not? I'll just continue ignoring it for now.

Dreamsorcerer:
Think it's blocked on the timeline protection...

.mypy.ini
webknjaz: > color_output, error_summary default to True (not that I've ever seen any colour in the output...

@Dreamsorcerer many tools have TTY auto-detection and use plain text output in non-interactive invocations. OTOH, most CIs support rendering ANSI colored terminal output. GHA runners don't allocate a TTY for job invocations (some other CIs do this). This means that all tools that don't special-case the CIs (which can be detected via env vars), don't produce ANSI output to be on the safe side. A lot of these apps have a CLI flag to force the color output anyway. And there's a conventional env var that is gaining popularity over the past years for doing the same — `$FORCE_COLOR`. If it is set, it is expected that a program supporting it would produce colorized output regardless of the presence of a TTY. Its popular counterpart `$NO_COLOR` even has its own website — https://no-color.org. Also, some software that doesn't yet support `FORCE_COLOR` (like old MyPy), has own env vars (e.g. `MYPY_FORCE_COLOR`).

As you can see, there's a number of reasons you haven't seen the colored output, especially if you're mostly checking out the CI logs.
Here's what I've been recently integrating into CIs, having realized that most of my pytest and mypy runs didn't produce fancy logs: [jaraco/skeleton@`e22ae08` (#67)](https://github.com/jaraco/skeleton/pull/67/commits/e22ae083f7e1ab41dfa8bb81bc7027281ad9dae6#diff-7829468e86c1cc5d5133195b5cb48e1ff6c75e3e9203777f6b2e379d9e4882b3R14-R17).

Back to the `color_output` setting — I'm going to guess it may only default to `true` in certain envs, but not the others. So adding it would still be warranted. Alternatively (or additionally), it'd be nice to have that patch I linked above integrated too.

Not sure about `error_summary` but sometimes I just follow the principle of explicit being better than implicit. From the PoV of a config reader, this reduces the mental overhead of having to figure out what the other settings are present implicitly. Plus, this saves from surprise defaults changes on updates.

> Would be really useful if we could have a global mypy config shared between all our repos with local overrides...

Yes, I've been toying (currently in my mind only) with an idea of having multiple skeletons for different aspects of projects being merged into an org-shared skeleton with typical file contents.
Something like https://github.com/jaraco/skeleton (which jaraco uses in all his projects, including setuptools and importlib-metadata, for example). But a more distributed ecosystem. Such skeletons are merged into other repositories downstream with `git merge --allow-unrelated-histories` (only for the first time, following update-merges shouldn't need this extra CLI flag).

webknjaz:
> Think it's blocked on the timeline protection...

I don't see a red cross there. It reports a `neutral` outcome, which is treated by the branch protection as passing. If there was to label specified, it'd block with a `failed` outcome.

So maybe GH was having problems itself.

.mypy.ini
```diff
@@ -0,0 +1,30 @@
...
+disable_error_code = comparison-overlap
+disallow_any_decorated = True
+disallow_any_generics = True
+disallow_any_unimported = True
+disallow_incomplete_defs = True
+disallow_subclassing_any = True
+disallow_untyped_calls = True
+disallow_untyped_decorators = True
+disallow_untyped_defs = True
+enable_error_code = redundant-expr, truthy-bool, ignore-without-code, unused-awaitable
```
webknjaz, line 14: I wonder if this can be multilined

.mypy.ini
Dreamsorcerer: > As you can see, there's a number of reasons you haven't seen the colored output, especially if you're mostly checking out the CI logs.

I've not seen it in my local terminal, and I did try explicitly setting the option (`git diff` or `pytest` produce coloured output...). Weird..

webknjaz: https://github.com/python/mypy/issues/7771 says "mypy ignores `color_output=true` if there is no tty" as I suspected.

Have you tried requesting it explicitly (as in `MYPY_FORCE_COLOR=1 mypy`)? Maybe it's still not forced somehow?
Here's somebody else's screenshot: https://github.com/python/mypy/issues/7771#issuecomment-1056173224.

P.S. Have you made sure there's actual errors? I don't remember if there's anything colored in successful runs.

PR #689, opened 2022-01-24
Document difference between `getone()` and `get()` methods

sebbASF:
Document difference between getone and get


asvetlov:
Fix linter please

PR #798, opened 2022-12-19
Revert collections import

Dreamsorcerer:


Dreamsorcerer:
@webknjaz I suspect this is the fix, but haven't tested it yet.

.github/workflows/ci.yml
```diff
@@ -26,7 +26,7 @@ jobs:
     - name: Setup Python
       uses: actions/setup-python@v4
       with:
-        python-version: 3.x
```
Dreamsorcerer, line 29: I think if we pin this to an older release, we're less likely to miss this in future.

Dreamsorcerer:
@webknjaz This should be correct now. If we merge and release this, it should fix aiohttp.

setup.cfg
```diff
@@ -8,7 +8,7 @@ long_description = file: README.rst
 
 
 [flake8]
-ignore = E302,E701,E305,E704,F811,N811, W503
+ignore = E302,E701,E305,E704,F811,N811,W503,Y026,Y027
```
webknjaz, line 11: What are these?

.github/workflows/ci.yml
webknjaz: Honestly, it's better to start implementing https://github.com/aio-libs/aiohttp/pull/7054#issuecomment-1358515510 rather than guessing which Python version would catch most problems. Any version from the supported runtime range may turn out to have differing typing info, and we wouldn't catch that…

setup.cfg
Dreamsorcerer: The rules that lead to me breaking the code originally.
https://github.com/PyCQA/flake8-pyi/issues/284

CHANGES/798.bugfix
```diff
@@ -0,0 +1 @@
+Fix a regression from 6.0.2 in type annotations when using Python <3.10.
```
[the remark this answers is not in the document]
Dreamsorcerer: I think the main cause is actually the collections.abc import.

multidict/__init__.pyi
```diff
@@ -1,12 +1,19 @@
...
+    MutableMapping,
+    TypeVar,
+    overload,
+)
 
 class istr(str): ...
 
 upstr = istr
 
-_S: TypeAlias = str | istr
```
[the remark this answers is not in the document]
Dreamsorcerer: TypeAlias doesn't really do anything, so I'd be inclined to leave it out until supported in all our Python versions. I expect pyupgrade will do this automatically anyway.

Again, I only made these changes in the first place, because flake8-pyi highlighted them and suggested that it was safe for all Python versions because it is a stub file and not read at runtime. We don't do this in other projects with inline typing.

setup.cfg
webknjaz: For the future: since flake8, unlike pylint, does not have human-readable rule identifiers, it's often hard to understand why a rule is ignored. Or what it does, even. This is why, it's a good idea to always annotate them with code comments.

CHANGES/798.bugfix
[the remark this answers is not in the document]
webknjaz: Ah, I didn't notice that. But FWIW `TypeAlias` is also new. The abc moves are old IIRC.
I'll leave it to you to improve the change note accuracy, then.

multidict/__init__.pyi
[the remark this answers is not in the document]
webknjaz: Okay, resolving this thread.

webknjaz (approved):
@Dreamsorcerer feel free to merge after dealing with the change note.

setup.cfg
Dreamsorcerer: Indeed, I normally do that, but that lack of existing comments caused me to miss it.

.github/workflows/ci.yml
eakmochi66: image

eakmochi66:
https://t.snapchat.com/PpyXxoUe

PR #829, opened 2023-04-07
Fix setuptools warning: "multidict._multilib not in packages list"

hoodmane:
This fixes the following build warning:

```
    ############################
    # Package would be ignored #
    ############################
    Python recognizes 'multidict._multilib' as an importable package,
    but it is not listed in the `packages` configuration of setuptools.

    'multidict._multilib' has been automatically added to the distribution only
    because it may contain data files, but this behavior is likely to change
    in future versions of setuptools (and therefore is considered deprecated).

    Please make sure that 'multidict._multilib' is included as a package by using
    the `packages` configuration field or the proper discovery methods
    (for example by using `find_namespace_packages(...)`/`find_namespace:`
    instead of `find_packages(...)`/`find:`).

    You can read more about "package discovery" and "data files" on setuptools
    documentation page.
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -80,7 +80,7 @@ args = dict(
         "GitHub: repo": "https://github.com/aio-libs/multidict",
     },
     license="Apache 2",
-    packages=["multidict"],
+    packages=["multidict", "multidict._multilib"],
     python_requires=">=3.7",
     include_package_data=True,
     exclude_package_data={"": ["*.c", "*.h"]},


setup.py
```diff
@@ -80,7 +80,7 @@ def read(f):
         "GitHub: repo": "https://github.com/aio-libs/multidict",
     },
     license="Apache 2",
-    packages=["multidict"],
+    packages=["multidict", "multidict._multilib"],
```
webknjaz, line 83: It's not a package — it only contains C-headers. So I don't think it's correct to add it here. Perhaps, this needs to be addressed differently.

setup.py
```diff
@@ -80,7 +80,7 @@ def read(f):
...
     python_requires=">=3.7",
     include_package_data=True,
```
webknjaz, line 85: I posted a question @ https://github.com/pypa/setuptools/issues/3340#issuecomment-1869926919, but it seems `include_package_data=True` should be re-assessed — do we need it enabled? Why? Would disabling it also remove the warning?

webknjaz: I experimented with this and it turned out that removing the setting yields the same distribution contents while not emitting the warning so I've gone with that: https://github.com/aio-libs/multidict/commit/31e11709ad59753f38b5f1a89716788025dc4718.

PR #924, opened 2024-01-17
Refactor pytests to make pickle testing in-memory

### LINKED ISSUE #922: [TODO] Migrate `tests/gen_pickles.py` into a pytest fixture
The idea is simple:
* [ ] Make a session-scoped fixture in `tests/conftest.py`
* [ ] Move the logic from `gen_pickles.py` there and generate the pickles in a `tmppath`-provided dir
* [ ] Integrate that into the pickling tests
* [ ] Delete `tests/gen_pickles.py` and the pickle files in the `tests/` dir
* [ ] _Bonus points_: keep pickles in-memory, so they don't hit the disk
* [ ] _Bonus 2_: check if a session fixture is needed and switch to the normal function scope, if not

vrdn-23:
## What do these changes do?

Refactors the pickling tests so that the pickles are stored in-memory and not on disk and move them into a purest function-scoped fixture.

No behavior changes for end user.

## Related issue number

Fixes #922 

## Checklist

- [?] I think the code is well written


tests/cimultidict-c-extension.pickle.1
webknjaz, line 1: Looks like this pickle file hasn't been removed. It needs to be removed and probably added to `tests/.gitignore`.

There's more files missed like this.

tests/conftest.py
```diff
@@ -158,6 +158,28 @@ def multidict_getversion_callable(multidict_module: ModuleType) -> Callable:
     return multidict_module.getversion
 
 
+def write(cls: Any, proto: int) -> bytes:
+    d = cls([("a", 1), ("a", 2)])
+    return pickle.dumps(d, proto)
+
+
+@pytest.fixture
+def in_memory_pickle_classes(multidict_module: ModuleType) -> dict[str, bytes]:
```
webknjaz, line 167: It's best to depend on `any_multidict_class` and get the  tag/module name from `multidict_implementation`. So there's no need to have internal loops. The protocols could be moved into a dedicated fixture with `params=`.

The return value would be just the pickled bytes.

tests/conftest.py
```diff
@@ -158,6 +158,28 @@ def multidict_getversion_callable(multidict_module: ModuleType) -> Callable:
     return multidict_module.getversion
 
 
+def write(cls: Any, proto: int) -> bytes:
```
webknjaz, line 161: This helper should have a useful name and a docstring. Or move this into the fixture.

tests/conftest.py
```diff
@@ -6,7 +6,7 @@
 from importlib import import_module
 from sys import version_info as _version_info
 from types import ModuleType
-from typing import Callable, Type
+from typing import Any, Callable, Type
```
webknjaz, line 9: Avoid using `Any` and follow other examples of signatures in this module.

tests/conftest.py
```diff
@@ -158,6 +158,28 @@ def multidict_getversion_callable(multidict_module: ModuleType) -> Callable:
...
 
 
+def write(cls: Any, proto: int) -> bytes:
+    d = cls([("a", 1), ("a", 2)])
+    return pickle.dumps(d, proto)
+
+
+@pytest.fixture
+def in_memory_pickle_classes(multidict_module: ModuleType) -> dict[str, bytes]:
+    """Generates a dict for in-memory storage of pickled classes"""
```
webknjaz, line 168: A docstring title must use an imperative mood and be a full sentence with a trailing period.

tests/cimultidict-c-extension.pickle.1
vrdn-23: So I think the file was removed the first time around. But I think Github shows the diffs for binary files that don't end with the extension `.0` as a bit different for the other files. I tried adding them back, checked that there were no diffs and then deleted them again using `git rm` and I do see the files being shown in the delete mode
```
 delete mode 100644 tests/multidict-pure-python.pickle.1
 delete mode 100644 tests/multidict-pure-python.pickle.2
 delete mode 100644 tests/multidict-pure-python.pickle.3
 delete mode 100644 tests/multidict-pure-python.pickle.4
 delete mode 100644 tests/multidict-pure-python.pickle.5
```

But the UI shows the same. Let me know if you have any other ideas on how to address this

tests/conftest.py
vrdn-23: Rewritten. If we're not using a dictionary there was no need for a tag-implementation level string so I simplified this. let me know if there's something else you had in mind.

tests/conftest.py
vrdn-23: Refactored away!

tests/conftest.py
vrdn-23: Removed

tests/conftest.py
vrdn-23: Fixed

tests/cimultidict-c-extension.pickle.1
webknjaz: @vrdn-23 your console snippet shows a `multidict` file, not `cimultidict`.

webknjaz: Try `git rm -f tests/*.pickle.*`.

tests/conftest.py
```diff
@@ -158,6 +158,17 @@ def multidict_getversion_callable(multidict_module: ModuleType) -> Callable:
     return multidict_module.getversion
 
 
+@pytest.fixture(params=list(range(pickle.HIGHEST_PROTOCOL + 1)))
```
webknjaz, line 161: Does this work without turning the generator into a list?

webknjaz: If not, please use tuple instead.

tests/conftest.py
```diff
@@ -158,6 +158,17 @@ def multidict_getversion_callable(multidict_module: ModuleType) -> Callable:
...
+def in_memory_pickle_object(
+    request: pytest.FixtureRequest,
+    any_multidict_class: Type[MutableMultiMapping[str]],
+) -> bytes:
+    """Generate an in-memory pickle of the multi-dict object."""
+    proto = request.param
+    d = any_multidict_class([("a", 1), ("a", 2)])
```
webknjaz, line 168: Please, use a reasonable variable name for the data you're storing there.

tests/multidict-c-extension.pickle.0
webknjaz, line 1: Looks like this file just got replaced with an empty one.

.gitignore
```diff
@@ -86,3 +86,6 @@ TAGS
 
 # pyenv
 .python-version
+
+# .pkl files in tests folder
+ tests/*.pickle.*
```
webknjaz, line 91: Could you move this into `tests/.gitignore`?

Also, having text files without a trailing LF is a bad tone.

tests/conftest.py
webknjaz: There's already a `pickle_protocol` that is injected through `pytest_generate_tests`. Can't you depend on it instead of maintaining a copy of the same logic?

tests/test_pickle.py
```diff
@@ -21,18 +21,12 @@ def test_pickle_proxy(any_multidict_class, any_multidict_proxy_class):
...
 
-def test_load_from_file(any_multidict_class, multidict_implementation, pickle_protocol):
-    multidict_class_name = any_multidict_class.__name__
-    pickle_file_basename = "-".join(
-        (
-            multidict_class_name.lower(),
-            multidict_implementation.tag,
-        )
-    )
+def test_load_from_file(
```
webknjaz, line 24: This test might need renaming since it no longer uses a file on disk.
Also, it seems like `test_pickle` might cover this case already.

webknjaz: OTOH, upon further reflection, I think I might know why the actual files are committed in Git — this is to ensure that newer multidict versions can load pickles of the older ones.
So we might be treating this incorrectly.

We might need to have a CLI interface for making the files in pytest but keep them. A fixture w/o an in-memory strucutre would be useful, though.

tests/cimultidict-c-extension.pickle.1
webknjaz: I guess, it might be GitHub's poor data representation, after all (although, I haven't checked in detail).

Anyway, per https://github.com/aio-libs/multidict/pull/924/files#r1463633143, we might need to keep the files, after all.

vrdn-23: Okay. I'm following the different threads but I'm not sure what the final resolution is. What is the final specification/requirements we want from this change? Could you update #922 with the new changes/ set of requirements that we need?

webknjaz: You're right. I'll try to circle back later.

Issue #921, opened 2024-01-16
[TODO] Hit 100% coverage in `multidict/_multidict_base.py`


webknjaz:
Currently, the coverage of this module is partial. Staring at https://app.codecov.io/github/aio-libs/multidict/commit/61f241fbfc4b507e8c44d7d8e60d3fc4414cc91f/blob/multidict/_multidict_base.py will help the contributor identify fully or partially uncovered lines.

The related tests are in https://github.com/aio-libs/multidict/blob/master/tests/test_multidict.py. The contributions can be made in small chunks. Submitting even a single test that improves coverage is great! It's also easier to accept/review smaller PRs.

In most cases, making new tests can be done by example. Starting with copying an existing test and tweaking it to hit different code branches...

Jamim:
_Hi @webknjaz,_

It's nice that this project have the highest coverage target possible, but the fact that Codecav enforces this target makes **any** PR marked with ❌ now, even though a PR could increase coverage, such as #927. I suggest disabling that check or lowering the target until 100% coverage is actually achieved.

_Thanks!_

webknjaz:
That check is not enforced. Besides, due to Codecov's flakiness, it's not always accurate.

a5r0n:
The last uncovered line is the `GenericAlias` for python versions older than 3.9

PR #928, opened 2024-01-20
Add tests to have full code coverage of the function `_viewbaseset_richcmp`

skinnyBat:
## What do these changes do?

Add tests to improve the code coverage of `_multidict_base_py`, by fully covering the `_viewbaseset_richcmp` function.

None

## Related issue number

Partially addresses https://github.com/aio-libs/multidict/issues/921.

## Checklist

### DIFF
--- a/CHANGES/928.contrib.rst
+++ b/CHANGES/928.contrib.rst
@@ -1,1 +1,3 @@
-Added tests to have full code coverage of the ``_viewbaseset_richcmp`` function -- by :user:`skinnyBat`
+Added tests to have full code coverage of the
+``multidict._multidict_base._viewbaseset_richcmp()`` function
+-- by :user:`skinnyBat`.
--- a/CHANGES/928.contrib.rst
+++ b/CHANGES/928.contrib.rst
@@ -1,0 +1,1 @@
+Added tests to have full code coverage of the ``_viewbaseset_richcmp`` function -- by :user:`skinnyBat`
--- a/multidict/_multidict_base.py
+++ b/multidict/_multidict_base.py
@@ -46,6 +46,8 @@ def _viewbaseset_richcmp(view, other, op):
             if elem not in view:
                 return False
         return True
+    else:  # pragma: no cover
+        return NotImplemented
 
 
 def _viewbaseset_and(view, other):
--- a/multidict/_multidict_base.py
+++ b/multidict/_multidict_base.py
@@ -4,10 +4,7 @@ from collections.abc import ItemsView, Iterable, KeysView, Set, ValuesView
 if sys.version_info >= (3, 11):
     from typing import assert_never
 else:
-    from typing import Any, NoReturn
-
-    def assert_never(value: Any) -> NoReturn:
-        raise AssertionError(f"Expected code to be unreachable, but got: {value!r}")
+    from typing_extensions import assert_never
 
 
 def _abc_itemsview_register(view_cls):
--- a/multidict/_multidict_base.py
+++ b/multidict/_multidict_base.py
@@ -1,4 +1,13 @@
+import sys
 from collections.abc import ItemsView, Iterable, KeysView, Set, ValuesView
+
+if sys.version_info >= (3, 11):
+    from typing import assert_never
+else:
+    from typing import Any, NoReturn
+
+    def assert_never(value: Any) -> NoReturn:
+        raise AssertionError(f"Expected code to be unreachable, but got: {value!r}")
 
 
 def _abc_itemsview_register(view_cls):
@@ -47,7 +56,7 @@ def _viewbaseset_richcmp(view, other, op):
                 return False
         return True
     else:  # pragma: no cover
-        return NotImplemented
+        assert_never(op)
 
 
 def _viewbaseset_and(view, other):
--- a/setup.cfg
+++ b/setup.cfg
@@ -44,7 +44,7 @@ classifiers =
 [options]
 python_requires = >= 3.7
 install_requires =
-  typing-extensions >= 4.1.0;python_version<'3.11'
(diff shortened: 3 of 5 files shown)


multidict/_multidict_base.py
```diff
@@ -46,6 +46,8 @@ def _viewbaseset_richcmp(view, other, op):
             if elem not in view:
                 return False
         return True
+    else:  # pragma: no cover
+        return NotImplemented
```
webknjaz, line 50: Why is this needed? Are there any other `op`s that exist?

Should this be `assert_never()` per https://stackoverflow.com/a/74535743/595220?

If there's other `op`s, let's cover this branch by using them in tests.

tests/test_multidict.py
```diff
@@ -281,34 +281,72 @@ def test_keys_is_set_less(self, cls: Type[MutableMultiMapping[str]]) -> None:
...
+    def test_keys_is_set_less_equal(
+        self,
+        cls: Type[MutableMultiMapping[str]],
+        contents: List[Tuple[str, str]],
+        expected: bool,
+    ) -> None:
+        d = cls(contents)
 
-        assert d.keys() <= {"key"}
+        assert (d.keys() <= {"key", "key2"}) == expected
```
webknjaz, line 301: For booleans, it is customary to use an identity check, just like with `None`:
```suggestion
        assert (d.keys() <= {"key", "key2"}) is expected
```

Although, I'd probably put the check result into a variable to avoid having complex nested inline code like this.

tests/test_multidict.py
```diff
@@ -281,34 +281,72 @@ def test_keys_is_set_less(self, cls: Type[MutableMultiMapping[str]]) -> None:
...
+        assert (d.keys() >= set_) == expected
+
+    @pytest.mark.parametrize("op", (operator.le, operator.lt, operator.ge, operator.gt))
+    def test_keys_compare_type_error(
+        self, cls: Type[MutableMultiMapping[str]], op: Callable[[object, object], bool]
     ) -> None:
         d = cls([("key", "value1")])
 
-        assert {"key"} >= d.keys()
+        with pytest.raises(TypeError):
```
webknjaz, line 335: When using `pytest.raises()`, we should always have a `match=` regex set in the check. With starting `^` and trailing `$`.

multidict/_multidict_base.py
webknjaz: If this is needed, add a bugfix change note. If not, don't include it and add a contrib change note.

https://github.com/aio-libs/multidict/tree/master/CHANGES#adding-change-notes-with-your-prs

skinnyBat: `op` can only be one of the six values(https://docs.python.org/3/c-api/typeobj.html#c.PyTypeObject.tp_richcompare). I added this so that the coverage doesn't report that the previous `elif` statement isn't fully covered. The previous `elif` should never to false, as these `if/elif` statements covers all legal values.

`assert_never()` only exists in Python 3.11 and above, so have also added a fallback for versions prior to that. 

As this doesn't change any user behavior, should this be a contrib change note, or a bugfix change note?

webknjaz: Contrib is good. As you mentioned — this doesn't bring end-user facing changes.

CHANGES/928.contrib.rst
```diff
@@ -0,0 +1 @@
+Added tests to have full code coverage of the ``_viewbaseset_richcmp`` function -- by :user:`skinnyBat`
```
webknjaz, line 1: We need a small punctuation correction — a trailing period. I also backed in a more specific path to the function:
```suggestion
Added tests to have full code coverage of the
``multidict._multidict_base._viewbaseset_richcmp()`` function
-- by :user:`skinnyBat`.
```

multidict/_multidict_base.py
```diff
@@ -1,5 +1,14 @@
+import sys
 from collections.abc import ItemsView, Iterable, KeysView, Set, ValuesView
 
+if sys.version_info >= (3, 11):
+    from typing import assert_never
+else:
+    from typing import Any, NoReturn
+
+    def assert_never(value: Any) -> NoReturn:
+        raise AssertionError(f"Expected code to be unreachable, but got: {value!r}")
```
webknjaz, line 10: A typical solution is to use `typing_extensions` to shim this. I'll need to be added to the runtime deps in `setup.cfg` with an environment marker, restricting the Python version.

tests/test_multidict.py
```diff
@@ -281,34 +281,79 @@ def test_keys_is_set_less(self, cls: Type[MutableMultiMapping[str]]) -> None:
...
         d = cls([("key", "value1")])
 
-        assert {"key"} >= d.keys()
+        with pytest.raises(
+            TypeError,
+            match=(
+                "^'.+' not supported between instances of '.*_KeysView' and 'NoneType'$"
+            ),
+        ):
+            op(d.keys(), None)
```
webknjaz, line 343: Could you implement a variant of https://github.com/aio-libs/multidict/pull/936#discussion_r1473671524 instead?

This would end up looking cleaner and not needing a `raises()` check. WDYT?

multidict/_multidict_base.py
webknjaz: I'm marking this thread as resolved, since I made a fallback suggestion in another place.

setup.cfg
```diff
@@ -43,6 +43,8 @@ classifiers =
 
 [options]
 python_requires = >= 3.7
+install_requires =
+  typing-extensions >= 4.1.0;python_version<'3.11'
```
webknjaz, line 47:
```suggestion
  typing-extensions >= 4.1.0; python_version < '3.11'
```

PR #936, opened 2024-01-31
Add test coverage for _multidict_base.py

a5r0n:
`NotImplemented` and `Iterable` but not `Set` for `_viewbaseset_{and, or, sub, xor}`

Partially addresses https://github.com/aio-libs/multidict/issues/921.

### DIFF
--- a/CHANGES/936.contrib.rst
+++ b/CHANGES/936.contrib.rst
@@ -1,0 +1,2 @@
+Add test coverage for _multidict_base.py 
+`NotImplemented`` and `Iterable-but-not-Set` for `and, or, sub, xor` operators
--- a/CHANGES/936.contrib.rst
+++ b/CHANGES/936.contrib.rst
@@ -1,2 +1,9 @@
-Added test coverage for the :ref:`and <python:and>`, :ref:`or <python:or>`, :py:obj:`sub <python:object.__sub__>`, and :py:obj:`xor <python:object.__xor__>` operators in the :file:`multidict/_multidict_base.py` module. It also covers  
-:py:data:`NotImplemented` and ":py:class:`~typing.Iterable`-but-not-:py:class:`~typing.Set`" cases there.
+Added test coverage for the :ref:`and <python:and>`, :ref:`or
+<python:or>`, :py:obj:`sub <python:object.__sub__>`, and
+:py:obj:`xor <python:object.__xor__>` operators in the
+:file:`multidict/_multidict_base.py` module. It also covers
+:py:data:`NotImplemented` and
+":py:class:`~typing.Iterable`-but-not-:py:class:`~typing.Set`"
+cases there.
+
+-- by :user:`a5r0n`
--- a/CHANGES/936.contrib.rst
+++ b/CHANGES/936.contrib.rst
@@ -1,2 +1,2 @@
-Added test coverage for the :ref:`and <python:and>`, :ref:`or <python:or>`, :py:obj:`sub <python:object.__sub__>`, and :py:obj:`xor <python:object.xor>` operators in the :file:`multidict/_multidict_base.py` module. It also covers  
+Added test coverage for the :ref:`and <python:and>`, :ref:`or <python:or>`, :py:obj:`sub <python:object.__sub__>`, and :py:obj:`xor <python:object.__xor__>` operators in the :file:`multidict/_multidict_base.py` module. It also covers  
 :py:data:`NotImplemented` and ":py:class:`~typing.Iterable`-but-not-:py:class:`~Set`" cases there.
--- a/CHANGES/936.contrib.rst
+++ b/CHANGES/936.contrib.rst
@@ -1,2 +1,2 @@
-Added test coverage for the :ref:`and <python:and>`, :ref:`or <python:or>`, :ref:`sub <python:sub>`, and :ref:`xor <python:xor>` operators in the :file:`multidict/_multidict_base.py` module. It also covers  
+Added test coverage for the :ref:`and <python:and>`, :ref:`or <python:or>`, :py:obj:`sub <python:object.__sub__>`, and :py:obj:`xor <python:object.xor>` operators in the :file:`multidict/_multidict_base.py` module. It also covers  
 :py:data:`NotImplemented` and ":py:class:`~typing.Iterable`-but-not-:py:class:`~Set`" cases there.
--- a/CHANGES/936.contrib.rst
+++ b/CHANGES/936.contrib.rst
@@ -1,2 +1,2 @@
-Add test coverage for _multidict_base.py 
-`NotImplemented`` and `Iterable-but-not-Set` for `and, or, sub, xor` operators
+Added test coverage for the :ref:`and <python:and>`, :ref:`or <python:or>`, :ref:`sub <python:sub>`, and :ref:`xor <python:xor>` operators in the :file:`multidict/_multidict_base.py` module. It also covers  
+:py:data:`NotImplemented` and ":py:class:`~typing.Iterable`-but-not-:py:class:`~Set`" cases there.
--- a/CHANGES/936.contrib.rst
+++ b/CHANGES/936.contrib.rst
@@ -1,2 +1,2 @@
 Added test coverage for the :ref:`and <python:and>`, :ref:`or <python:or>`, :py:obj:`sub <python:object.__sub__>`, and :py:obj:`xor <python:object.__xor__>` operators in the :file:`multidict/_multidict_base.py` module. It also covers  
-:py:data:`NotImplemented` and ":py:class:`~typing.Iterable`-but-not-:py:class:`~Set`" cases there.
+:py:data:`NotImplemented` and ":py:class:`~typing.Iterable`-but-not-:py:class:`~typing.Set`" cases there.
--- a/tests/test_multidict.py
+++ b/tests/test_multidict.py
@@ -388,7 +388,7 @@ class BaseMultiDictTest:
 
         assert {"key"} == {"key", "key2"} & d.keys()
 
-    def test_and_not_implemented(self, cls: Type[MutableMultiMapping[str]]) -> None:
+    def test_bitwise_and_not_implemented(self, cls: Type[MutableMultiMapping[str]]) -> None:
         d = cls([("key", "value1")])
 
         sentinel_operation_result = object()
@@ -400,7 +400,7 @@ class BaseMultiDictTest:
 
         assert d.keys() & RightOperand() is sentinel_operation_result
 
-    def test_and_iterable_not_set(self, cls: Type[MutableMultiMapping[str]]) -> None:
+    def test_bitwise_and_iterable_not_set(self, cls: Type[MutableMultiMapping[str]]) -> None:
         d = cls([("key", "value1")])
 
         assert {"key"} == d.keys() & ["key", "key2"]
--- a/tests/test_multidict.py
+++ b/tests/test_multidict.py
@@ -387,6 +387,17 @@ class BaseMultiDictTest:
 
         assert {"key"} == {"key", "key2"} & d.keys()
 
+    def test_and_not_implemented(self, cls: Type[MutableMultiMapping[str]]) -> None:
+        d = cls([("key", "value1")])
+
+        with pytest.raises(TypeError):
+            operator.and_(d.keys(), 1)
+
+    def test_and_iterable_not_set(self, cls: Type[MutableMultiMapping[str]]) -> None:
+        d = cls([("key", "value1")])
+
+        assert {"key"} == d.keys() & ["key", "key2"]
+
     def test_or(self, cls: Type[MutableMultiMapping[str]]) -> None:
         d = cls([("key", "value1")])
 
@@ -396,6 +407,17 @@ class BaseMultiDictTest:
         d = cls([("key", "value1")])
 
         assert {"key", "key2"} == {"key2"} | d.keys()
+
+    def test_or_not_implemented(self, cls: Type[MutableMultiMapping[str]]) -> None:
+        d = cls([("key", "value1")])
+
+        with pytest.raises(TypeError):
+            operator.or_(d.keys(), 1)
+
+    def test_or_iterable_not_set(self, cls: Type[MutableMultiMapping[str]]) -> None:
+        d = cls([("key", "value1")])
+
+        assert {"key", "key2"} == d.keys() | ["key2"]
 
     def test_sub(self, cls: Type[MutableMultiMapping[str]]) -> None:
         d = cls([("key", "value1"), ("key2", "value2")])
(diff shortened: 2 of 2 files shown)


CHANGES/936.contrib.rst
webknjaz, line 1: The Sphinx build on RTD fails because this patch contains non-RST syntax: https://readthedocs.org/projects/multidict/builds/23294665/.

It must also use proper writing style per https://github.com/aio-libs/multidict/tree/master/CHANGES#readme. It's not a commit message, so it shouldn't be imperative. The audience of the changelog expects references to what's changed since the last release, not action items.

CHANGES/936.contrib.rst
```diff
@@ -0,0 +1,2 @@
+Add test coverage for _multidict_base.py 
+`NotImplemented`` and `Iterable-but-not-Set` for `and, or, sub, xor` operators
```
webknjaz, line 2: This should clear the Sphinx errors, plus brush up the style a bit. With full sentences et al.
```suggestion
Added test coverage for the :ref:`and <python:and>`, :ref:`or <python:or>`, :ref:`sub <python:sub>`, and :ref:`xor <python:xor>` operators in the :file:`multidict/_multidict_base.py` module. It also covers  
:py:data:`NotImplemented` and ":py:class:`~typing.Iterable`-but-not-:py:class:`~Set`" cases there.
```

tests/test_multidict.py
```diff
@@ -387,6 +387,17 @@ def test_and2(self, cls: Type[MutableMultiMapping[str]]) -> None:
 
         assert {"key"} == {"key", "key2"} & d.keys()
 
+    def test_and_not_implemented(self, cls: Type[MutableMultiMapping[str]]) -> None:
+        d = cls([("key", "value1")])
+
+        with pytest.raises(TypeError):
+            operator.and_(d.keys(), 1)
```
webknjaz, line 394: How about just
```suggestion
            d.keys() and 1
```
? I think it'd be clearer to an inexperienced reader.

tests/test_multidict.py
```diff
@@ -397,6 +408,17 @@ def test_or2(self, cls: Type[MutableMultiMapping[str]]) -> None:
 
         assert {"key", "key2"} == {"key2"} | d.keys()
 
+    def test_or_not_implemented(self, cls: Type[MutableMultiMapping[str]]) -> None:
+        d = cls([("key", "value1")])
+
+        with pytest.raises(TypeError):
+            operator.or_(d.keys(), 1)
```
webknjaz, line 415:
```suggestion
            d.keys() or 1
```

tests/test_multidict.py
```diff
@@ -407,6 +429,17 @@ def test_sub2(self, cls: Type[MutableMultiMapping[str]]) -> None:
 
         assert {"key3"} == {"key", "key2", "key3"} - d.keys()
 
+    def test_sub_not_implemented(self, cls: Type[MutableMultiMapping[str]]) -> None:
+        d = cls([("key", "value1"), ("key2", "value2")])
+
+        with pytest.raises(TypeError):
+            operator.sub(d.keys(), 1)
```
webknjaz, line 436:
```suggestion
            d.keys() - 1
```

tests/test_multidict.py
```diff
@@ -387,6 +387,17 @@ def test_and2(self, cls: Type[MutableMultiMapping[str]]) -> None:
 
         assert {"key"} == {"key", "key2"} & d.keys()
 
+    def test_and_not_implemented(self, cls: Type[MutableMultiMapping[str]]) -> None:
+        d = cls([("key", "value1")])
+
+        with pytest.raises(TypeError):
```
webknjaz, line 393: It's usually important to use the `match=` regexp in the `pytest.raises()` and similar checks. Could you add it?

webknjaz: OTOH, it might be cleaner to make a class with `__rand__`, and use its instance instead of `1`. You could track that it was called.

a5r0n:
@webknjaz if I'm not wrong using `and, or, -, ^` will give errors from pytest, but I will check that once again

tests/test_multidict.py
webknjaz: This might not even need a literal class definition but a mock with that method and a check for how many times it was called + args.

webknjaz:
@a5r0n thanks for starting this effort! Those operators shouldn't cause problems in pytest. However, linters (like flake8) will be unhappy about unassigned expressions. To address that, use `# noqa` comments with specific error codes and a comment text explaining why that linting ignore is necessary.

webknjaz:
Oh, and the current CI failures don't seem to be related to your patch. They'll require addressing separately. I'll have to take a look myself. You'll need to rebase your PR branch at some point, once that's fixed.

P.S. Here's a recommendation for the future PRs on GitHub: https://hynek.me/articles/pull-requests-branch/.

CHANGES/936.contrib.rst
```diff
@@ -0,0 +1,2 @@
+Added test coverage for the :ref:`and <python:and>`, :ref:`or <python:or>`, :ref:`sub <python:sub>`, and :ref:`xor <python:xor>` operators in the :file:`multidict/_multidict_base.py` module. It also covers
```
webknjaz, line 1: Looks like the labels only exist for and/or. Let's link something else for the rest:
```suggestion
Added test coverage for the :ref:`and <python:and>`, :ref:`or <python:or>`, :py:obj:`sub <python:object.__sub__>`, and :py:obj:`xor <python:object.xor>` operators in the :file:`multidict/_multidict_base.py` module. It also covers  
```

tests/test_multidict.py
a5r0n: can you explain more about that?

CHANGES/936.contrib.rst
```diff
@@ -0,0 +1,2 @@
+Added test coverage for the :ref:`and <python:and>`, :ref:`or <python:or>`, :py:obj:`sub <python:object.__sub__>`, and :py:obj:`xor <python:object.xor>` operators in the :file:`multidict/_multidict_base.py` module. It also covers
```
webknjaz, line 1:
```suggestion
Added test coverage for the :ref:`and <python:and>`, :ref:`or <python:or>`, :py:obj:`sub <python:object.__sub__>`, and :py:obj:`xor <python:object.__xor__>` operators in the :file:`multidict/_multidict_base.py` module. It also covers  
```

tests/test_multidict.py
```diff
@@ -387,6 +387,17 @@ def test_and2(self, cls: Type[MutableMultiMapping[str]]) -> None:
 
         assert {"key"} == {"key", "key2"} & d.keys()
 
+    def test_and_not_implemented(self, cls: Type[MutableMultiMapping[str]]) -> None:
+        d = cls([("key", "value1")])
+
+        with pytest.raises(TypeError, match=r"unsupported operand type(\(s\))? for \&"):
```
webknjaz, line 393: It's usually better to match from start to finish:
```suggestion
        with pytest.raises(TypeError, match=r"^unsupported operand type(\(s\))? for \&$"):
```
(or at least, from start)

webknjaz: But have you given any thought to implementing https://github.com/aio-libs/multidict/pull/936#discussion_r1473653123?

tests/test_multidict.py
webknjaz: If the left operand's `__and__` returns `NotImplemented`, then Python would attempts calling `__rand__` from the right operand. So a properly crafted object type may have `__rand__(self, other)` defined that doesn't produce an exception. It could return a sentinel value for which we could check, for example.

CHANGES/936.contrib.rst
```diff
@@ -0,0 +1,2 @@
+Added test coverage for the :ref:`and <python:and>`, :ref:`or <python:or>`, :py:obj:`sub <python:object.__sub__>`, and :py:obj:`xor <python:object.__xor__>` operators in the :file:`multidict/_multidict_base.py` module. It also covers  
+:py:data:`NotImplemented` and ":py:class:`~typing.Iterable`-but-not-:py:class:`~Set`" cases there.
```
webknjaz, line 2: Did I miss this earlier? :thinking:
```suggestion
:py:data:`NotImplemented` and ":py:class:`~typing.Iterable`-but-not-:py:class:`~typing.Set`" cases there.
```

CHANGES/936.contrib.rst
webknjaz, line 1: At last, the docs build succeeds and the change note shows up there: https://multidict--936.org.readthedocs.build/en/936/changes/

tests/test_multidict.py
webknjaz: Here's an example without mocks: https://github.com/aio-libs/multidict/pull/936/files#r1473671524.

tests/test_multidict.py
```diff
@@ -397,6 +408,17 @@ def test_or2(self, cls: Type[MutableMultiMapping[str]]) -> None:
 
         assert {"key", "key2"} == {"key2"} | d.keys()
 
+    def test_or_not_implemented(self, cls: Type[MutableMultiMapping[str]]) -> None:
+        d = cls([("key", "value1")])
+
+        with pytest.raises(TypeError, match=r"unsupported operand type(\(s\))? for \|"):
+            d.keys() | 1
```
webknjaz, line 415: It's not `or` but a "bitwise or", no?

tests/test_multidict.py
a5r0n: `match=r"^unsupported operand type(\(s\))? for \|: '(multidict._multidict.)?_KeysView' and 'int'",`
is that better?
(`'(multidict._multidict.)?` included for c-ext tests)

tests/test_multidict.py
a5r0n: `or` will not give you any error

tests/test_multidict.py
```diff
@@ -387,6 +387,17 @@ def test_and2(self, cls: Type[MutableMultiMapping[str]]) -> None:
 
         assert {"key"} == {"key", "key2"} & d.keys()
 
+    def test_and_not_implemented(self, cls: Type[MutableMultiMapping[str]]) -> None:
+        d = cls([("key", "value1")])
+
+        with pytest.raises(TypeError, match=r"unsupported operand type(\(s\))? for \&"):
+            d.keys() & 1
```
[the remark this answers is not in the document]
a5r0n: this test case is just to cover the `NotImplemented` that shows up as `TypeError`, with your sugestions we dont cover that line, do we?

```python
if not isinstance(other, Iterable):
        return NotImplemented
```

tests/test_multidict.py
a5r0n: Okay, now i got you, i will try to do so.

Jamim:
> @webknjaz
> Oh, and the current CI failures don't seem to be related to your patch. They'll require addressing separately. I'll have to take a look myself. You'll need to rebase your PR branch at some point, once that's fixed.

Those failures are addressed in #937.

tests/test_multidict.py
webknjaz: @a5r0n I know but the test is called incorrectly, then. That name implies that `or` is tested, not `|`.

tests/test_multidict.py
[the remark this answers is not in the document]
webknjaz: @a5r0n yeah, but your previous implementation didn't either, right? Anyway, you can't cover 2 `return` statements with a single invocation. So that would need to be addressed separately. Are you able to measure coverage locally, to see which lines are actually hit?

tests/test_multidict.py
```diff
@@ -397,6 +415,23 @@ def test_or2(self, cls: Type[MutableMultiMapping[str]]) -> None:
 
         assert {"key", "key2"} == {"key2"} | d.keys()
 
+    def test_or_not_implemented(self, cls: Type[MutableMultiMapping[str]]) -> None:
```
webknjaz, line 418:
```suggestion
    def test_bitwise_or_not_implemented(self, cls: Type[MutableMultiMapping[str]]) -> None:
```

tests/test_multidict.py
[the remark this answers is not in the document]
a5r0n: i didnt understand you eralier, your suggestion works fine.

the last line that dont have coverage is `if isinstance(view, Set): `

tests/test_multidict.py
webknjaz: Oh, so the old names are a bit inaccurate too. Something to be fixed separately.

webknjaz:
@a5r0n I clicked "rebase" so this PR picks up the CI fixes.

CHANGES/936.contrib.rst
```diff
@@ -0,0 +1,2 @@
+Added test coverage for the :ref:`and <python:and>`, :ref:`or <python:or>`, :py:obj:`sub <python:object.__sub__>`, and :py:obj:`xor <python:object.__xor__>` operators in the :file:`multidict/_multidict_base.py` module. It also covers  
+:py:data:`NotImplemented` and ":py:class:`~typing.Iterable`-but-not-:py:class:`~typing.Set`" cases there.
```
webknjaz, line 2:
```suggestion
Added test coverage for the :ref:`and <python:and>`, :ref:`or
<python:or>`, :py:obj:`sub <python:object.__sub__>`, and
:py:obj:`xor <python:object.__xor__>` operators in the
:file:`multidict/_multidict_base.py` module. It also covers
:py:data:`NotImplemented` and
":py:class:`~typing.Iterable`-but-not-:py:class:`~typing.Set`"
cases there.

-- by :user:`a5r0n`
```

tests/test_multidict.py
```diff
@@ -387,6 +388,23 @@ def test_and2(self, cls: Type[MutableMultiMapping[str]]) -> None:
 
         assert {"key"} == {"key", "key2"} & d.keys()
 
+    def test_and_not_implemented(self, cls: Type[MutableMultiMapping[str]]) -> None:
```
webknjaz, line 391:
```suggestion
    def test_bitwise_and_not_implemented(self, cls: Type[MutableMultiMapping[str]]) -> None:
```

tests/test_multidict.py
```diff
@@ -387,6 +388,23 @@ def test_and2(self, cls: Type[MutableMultiMapping[str]]) -> None:
...
+        d = cls([("key", "value1")])
+
+        sentinel_operation_result = object()
+
+        class RightOperand:
+            def __rand__(self, other):
```
webknjaz, line 397: @a5r0n so MyPy demands annotations here: https://results.pre-commit.ci/run/github/51206425/1706753865.XjT9kVUPShKRS-PfstvZaQ.

a5r0n: Done

webknjaz: Looks like it wants type args there: https://results.pre-commit.ci/run/github/51206425/1706754519.rj18BkaQT4qOcCG8QRnS0A.

tests/test_multidict.py
```diff
@@ -387,6 +388,23 @@ def test_and2(self, cls: Type[MutableMultiMapping[str]]) -> None:
 
         assert {"key"} == {"key", "key2"} & d.keys()
 
+    def test_bitwise_and_not_implemented(self, cls: Type[MutableMultiMapping[str]]) -> None:
+        d = cls([("key", "value1")])
+
+        sentinel_operation_result = object()
+
+        class RightOperand:
+            def __rand__(self, other: KeysView) -> object:
```
webknjaz, line 397: This should do, I guess:
```suggestion
            def __rand__(self, other: KeysView[str]) -> object:
```

tests/test_multidict.py
a5r0n: I see, thats should be `KeysView[str]` i guess?

webknjaz: Seems so.

webknjaz: Marking as resolved.

PR #937, opened 2024-02-01
🐛 Use compatible `pytest < 8` when running MyPy against Python 3.7

Jamim:
_Hello @webknjaz,_

## What do these changes do?

These changes:
  - fix `pytest`/`mypy` compatibility issue for Python 3.7
  - bump [pypa/cibuildwheel](https://github.com/pypa/cibuildwheel) to `2.16.5`

Issues on CI will be resolved.

## Related issue number

  * pypa/cibuildwheel#1748

## Checklist

_Best regards!_

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -155,7 +155,7 @@ repos:
     additional_dependencies:
     - types-docutils
     - lxml  # dep of `--txt-report`, `--cobertura-xml-report` & `--html-report`
-    - pytest
+    - pytest < 8
     - Sphinx >= 5.3.0, < 6
     args:
     - --python-version=3.7
--- a/CHANGES/937.contrib.rst
+++ b/CHANGES/937.contrib.rst
@@ -1,0 +1,3 @@
+The version of pytest is now capped below 8, when running MyPy
+against Python 3.7. This pytest release dropped support for
+said runtime.
--- a/docs/spelling_wordlist.txt
+++ b/docs/spelling_wordlist.txt
@@ -71,6 +71,7 @@ pre
 proxied
 pyenv
 pyinstaller
+pytest
 refactor
 refactored
 regex


Jamim:
Do we need a news fragment for these trivial fixes? I don't think so.

Jamim:
@webknjaz, as you can see, this PR fixes CI failures. The only reason for the ❌ mark is the Timeline protection which, I believe, might be skipped in this case.

webknjaz:
Please, make atomic PRs. The one with testing fixes would need a `contrib` change note. And shouldn't be combined with unrelated version bumps. Smashing all together hurts transparency and traceability.

webknjaz:
I merged dependabot's PR for cibuildwheel. MyPy bump should be separate, but there's no rush to do that — it doesn't seem related and I'm fine waiting for dependabot to send a PR, it's not a priority.

CHANGES/937.bugfix.rst
webknjaz, line 1: It's not a user-facing change. Use `contrib`.

.pre-commit-config.yaml
```diff
@@ -155,7 +155,7 @@ repos:
     additional_dependencies:
     - types-docutils
     - lxml  # dep of `--txt-report`, `--cobertura-xml-report` & `--html-report`
-    - pytest
+    - pytest<8
```
webknjaz, line 158: match the surrounding style
```suggestion
    - pytest < 8
```

CHANGES/937.bugfix.rst
```diff
@@ -0,0 +1 @@
+pytest/mypy compatibility issue for Python 3.7 has been fixed.
```
webknjaz, line 1: Let's make this useful with context.
```suggestion
The version of pytest is now capped below 8, when running MyPy
against Python 3.7. This pytest release dropped support for
said runtime.
```

PR #940, opened 2024-02-01
🧪 Update syntax for setting step outputs in GitHub Actions CI/CD workflow definition

Jamim:
## What do these changes do?

The `set-output` command is deprecated and will be disabled soon.
For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

## Are there changes in behavior for the user?

No.

## Related issue number

`N\A`

## Checklist

- [x] I think the code is well written
- [x] Documentation reflects the changes

PR #942, opened 2024-02-02
🧪 Fix the `Get pip cache dir` CI step for Windows runners

Jamim:
## What do these changes do?

Fix the `Get pip cache dir` CI step for [Windows runners](https://github.com/aio-libs/multidict/actions/runs/7751071979/job/21138438840#step:6:32).
Please see annotations [here](https://github.com/aio-libs/multidict/actions/runs/7751071979).

No.

## Related issue number
  * actions/runner#2224

This bug was introduced in this PR:
  * #940

## Checklist

### DIFF
--- a/CHANGES/942.contrib.rst
+++ b/CHANGES/942.contrib.rst
@@ -1,0 +1,5 @@
+In the GitHub Actions CI/CD workflow definition,
+the ``Get pip cache dir`` step has been fixed for
+Windows runners by adding ``shell: bash``.
+See `actions/runner#2224 <https://github.com/actions/runner/issues/2224>`_
+for more details.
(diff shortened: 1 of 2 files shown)


CHANGES/942.contrib.rst
```diff
@@ -0,0 +1,5 @@
+In the GitHub Actions CI/CD workflow definition,
+the ``Get pip cache dir`` step has been fixed for
+Windows runners by adding ``shell: bash``.
+See `this issue <https://github.com/actions/runner/issues/2224>`_
```
Jamim, line 4: > WARNING: Duplicate explicit target name: "this issue"

`this issue` is [not a valid link caption](https://readthedocs.org/projects/multidict/builds/23318806/) since there is already `this issue` in other news fragment 😅

webknjaz: You can use a double trailing underscore for anonymous references.

Jamim: Do you want me to update the PR or keep `actions/runner#2224` as the link caption?

webknjaz: FWIW some of the changes don't actually need a change note. Notes should be about the high level impact, not low-level internals.

PR #943, opened 2024-02-02
🧪 Fix pip cache keys interpolation

Jamim:
## What do these changes do?

Fix [rendering](https://github.com/aio-libs/multidict/actions/runs/7751071979/job/21138438717#step:6:3) of `pip` cache keys.

No.

## Related issue number

`N\A`

## Checklist

### DIFF
--- a/CHANGES/943.contrib.rst
+++ b/CHANGES/943.contrib.rst
@@ -1,0 +1,3 @@
+``pip`` cache keys rendering has been
+fixed by adding missing ``$`` signs
+in the GitHub Actions CI/CD workflow definition.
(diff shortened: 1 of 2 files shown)


Jamim:
By the way, I'm not sure whether `${{ matrix.no-extensions }}` should be part of the key.
I suppose installed packages doesn't vary because of building C Extensions.

CHANGES/943.contrib.rst
```diff
@@ -0,0 +1,3 @@
+``pip`` cache keys rendering has been
```
webknjaz, line 1:
```suggestion
Interpolation of the ``pip`` cache keys has been
```

CHANGES/943.contrib.rst
```diff
@@ -0,0 +1,3 @@
...
+fixed by adding missing ``$`` signs
```
webknjaz, line 2:
```suggestion
fixed by adding missing ``$`` syntax
```

webknjaz:
I'm not sure but it's better safe than sorry.

PR #946, opened 2024-02-03
🧪 Skip tests when building `x86_64` wheels

Jamim:
_Hello @webknjaz,_

## What do these changes do?

Testing is redundant when building wheels for `x86_64` since there are dedicated jobs that run tests for this arch.
Disabling it reduces build time noticeably.

Related documentation:
  - https://cibuildwheel.readthedocs.io/en/stable/options/#test-skip

`CI/CD` build time will be reduced.

## Related issue number

`N/A`

## Checklist

##
_Best regards!_

### DIFF
--- a/CHANGES/946.contrib.rst
+++ b/CHANGES/946.contrib.rst
@@ -1,0 +1,2 @@
+Duplicated test runs have been disabled when
+building wheels for the ``x86_64`` architecture.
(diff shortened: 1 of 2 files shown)


Jamim:
_Hello @webknjaz,_
What do you think of the proposed changes?

webknjaz:
I'm hesitant about this: cibuildwheel tests run under a different environment so I feel like we'd benefit from covering more ground this way. Yes, it'd cut the CI use but I'm not sure that's what we need. Besides, if we were to disable this, it should be on the CI configuration level, not a tool-global — local runs would still benefit from producing wheels that go through the testing stage.

webknjaz:
I'm inclined to reject this based on the above.

PR #998, opened 2024-09-09
Fix codecov reporting

bdraco:
Same fix as we did for yarl


bdraco:
I think its actually working ok now, was just behind

bdraco:
we likely will need this once we do https://github.com/aio-libs/multidict/pull/944

PR #1002, opened 2024-09-09
Python 3.13 support

bdraco:
needs https://github.com/aio-libs/multidict/pull/993 to pass

## What do these changes do?

Support Python 3.13 along with wheels


.github/workflows/ci-cd.yml
```diff
@@ -150,11 +150,10 @@ jobs:
     strategy:
       matrix:
         pyver:
-        - 3.13-dev
+        - 3.13
         - 3.12
         - 3.11
-        - >-
-          3.10
+        - 3.10
```
bdraco, line 156:
```suggestion
        - "3.10"
```

.github/workflows/ci-cd.yml
```diff
@@ -199,12 +198,7 @@ jobs:
       uses: actions/setup-python@v5
       with:
         python-version: ${{ matrix.pyver }}
-        allow-prereleases: >-
```
webknjaz, line 202: @bdraco this should be kept long-term so that adding `-dev` things in the future wouldn't require remembering to reimplement it once again.

bdraco: I matched it to yarl since I wasn't sure if `3.13-dev` would pull non `rc`/`stable` 3.13

https://github.com/aio-libs/yarl/blob/eaf76414d44e72b022a7dee3b48bb532934a2f33/.github/workflows/ci-cd.yml#L226

bdraco: It looks like keeping `-dev` will still work https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#specifying-a-python-version

bdraco: I'll revert that change

bdraco: Just realized there are other checks for `-dev` though that we want to be fatal if 3.13 breaks though....so I'm not sure if reverting it makes sense

webknjaz: Okay, leave it as is. I usually like using a different syntax `~3.13.0-0` and check for `~` in other places. But I suppose it's not important to figure out *right now*. Something to remember at some point, perhaps.

I also didn't fully realize that this is about the action input. I typically have this in the check for allowing failures which is what confused me and I overlooked the context.

PR #1005, opened 2024-09-09
Adjust env markers and deps for Python 3.8+

bdraco:
followup to #997

https://github.com/aio-libs/multidict/commit/4fffc3d5d647d631f8830282d16ba7260dabea39#r146458357
https://github.com/aio-libs/multidict/commit/4fffc3d5d647d631f8830282d16ba7260dabea39#r146458343

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -155,6 +155,7 @@ repos:
     additional_dependencies:
     - types-docutils
     - lxml  # dep of `--txt-report`, `--cobertura-xml-report` & `--html-report`
+    - pytest
     - Sphinx >= 5.3.0
     args:
     - --python-version=3.8
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -155,8 +155,7 @@ repos:
     additional_dependencies:
     - types-docutils
     - lxml  # dep of `--txt-report`, `--cobertura-xml-report` & `--html-report`
-    - pytest < 8
-    - Sphinx >= 5.3.0, < 6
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -155,8 +155,7 @@ repos:
     additional_dependencies:
     - types-docutils
     - lxml  # dep of `--txt-report`, `--cobertura-xml-report` & `--html-report`
-    - pytest < 8
```
webknjaz, line 158: I think `pytest` is still needed, just not capped. MyPy is linting the tests too.

.pre-commit-config.yaml
```diff
@@ -155,8 +155,7 @@ repos:
...
-    - Sphinx >= 5.3.0, < 6
+    - Sphinx >= 5.3.0
```
webknjaz, line 158:
```suggestion
    - pytest
    - Sphinx >= 5.3.0
```

.pre-commit-config.yaml
bdraco: Yeah that makes more sense.... getting tired 💤

webknjaz (approved):

webknjaz:
@bdraco missed removing the conditional imports: https://app.codecov.io/github/aio-libs/multidict/commit/6197e510d9c11814524f1a48117cf2a45ba1a59a/blob/tests/conftest.py?flags%5B0%5D=pytest#L11.

bdraco:
I checked the main code base but missed the tests!  I’ll take a look when I get back back from finding dinner before I go to bed

PR #1007, opened 2024-09-09
Bump RTD to use Python 3.12 and ubuntu-24.04

bdraco:
[Bumping sphinx](https://github.com/aio-libs/multidict/pull/999) is currently blocked because 8.x required Python 3.10+

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -150,7 +150,7 @@ jobs:
     strategy:
       matrix:
         pyver:
-        - 3.13
+        - 3.13-dev
         - 3.12
         - 3.11
         - "3.10"
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -150,7 +150,7 @@ jobs:
     strategy:
       matrix:
         pyver:
-        - 3.13-dev
+        - 3.13
         - 3.12
         - 3.11
         - "3.10"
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-cd.yml
```diff
@@ -150,7 +150,7 @@ jobs:
     strategy:
       matrix:
         pyver:
-        - 3.13
+        - 3.13-dev
```
bdraco, line 153:
```suggestion
        - 3.13
```

.readthedocs.yml
```diff
@@ -3,10 +3,10 @@
 version: 2
 
 build:
-  os: ubuntu-22.04
+  os: ubuntu-24.04
   tools:
     python: >-
-      3.9
+      3.12
```
webknjaz, line 9: N.B. In the past, I'd keep this at the lowest possible value to accommodate docs contributors running this on various versions. However, nowadays, I sometimes make it have the version that matches GHA.

webknjaz: @bdraco this is something I'd like to have in the future: https://github.com/ansible/awx-plugins/blob/8ba56cc/.readthedocs.yaml — the same entry-points/automations being called @ RTD and GHA. That project doesn't use make, but I wanted to show the general idea.

bdraco: That does looks like a nicer / more maintainable design.  I'm still wrapping my head around all the GHAs between all the projects.  Its a lot to take in, and I'm playing catch up.

PR #1008, opened 2024-09-09
Bump Python to 3.12 for linters

bdraco:
Needed for sphinx 8+, and the spell checker will fail with sphinx 8+ on python < 3.10

unblocks https://github.com/aio-libs/multidict/pull/999


bdraco:
merged into dependabot/pip/sphinx-8.0.2

PR #1009, opened 2024-09-10
📦 Move packaging to PEP 517 in-tree backend

webknjaz:
This essentially allows the cythonization opt-out be controlled by the `pure-python` PEP 517 config setting that can be passed to the corresponding build frontends via their respective CLIs.


webknjaz:
This is me cleaning up uncommitted things lying around on disk. It needs some love. **Do not merge.**

IIRC, the thing that I was trying to figure out last was generating gcov coverage, looking into https://gcovr.com. It supposedly needs to be compiled twice (unlike what was needed for Cython) with some non-optimized flags, but I haven't gotten it to work. If anybody has ideas — plz document them here.

PR #1010, opened 2024-09-10
Remove compat code for Python older than 3.8 from tests

bdraco:
https://github.com/aio-libs/multidict/pull/1005#issuecomment-2339416121
Remove pre Python 3.8+ compat code from tests

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -3,11 +3,12 @@ from __future__ import annotations
 import argparse
 import pickle
 from dataclasses import dataclass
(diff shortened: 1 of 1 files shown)


tests/conftest.py
webknjaz, line 1: @bdraco this missed the `PY_38_AND_BELOW` constant and dependent blocks FYI

bdraco: Thanks. Done in https://github.com/aio-libs/multidict/pull/1020

Issue #1014, opened 2024-09-18
Support for the free-threaded build of CPython 3.13


lysnikolaou:
## Long story short

Hey all! 👋

I took the liberty of looking into adding support for the 3.13 free-threaded build of CPython. This aims to be a tracking issue for all work necessary to support it, for which I'm going to open a PR shortly.

- [ ] Set up CI for free-threading
- [ ] Move away from APIs that use borrowed references (`PyList_GetItem`, `PyDict_GetItem` and friends) under the free-threaded build
- [ ] Audit the C extension module for thread-safety issues
  - [ ] There currently only seem to be issues with `pair_list_global_version`. If there's something else that I might've missed, please let me know.
- [ ] Mark the C extension module as thread-safe with `PyUnstable_Module_SetGIL`
- [ ] Build wheels for the free-threaded build.

Hope this all sounds good! Let me know in case there's something that doesn't seem right.

ngoldbaum:
It looks like all the checkmarks are ticked, can this be closed?

lysnikolaou:
I think it can be closed, but I'll let someone from the core team push the button.

lysnikolaou:
Folks, would it be possible to do a release so that free-threaded wheels get published on PyPI?

PR #1015, opened 2024-09-18
Implement support for the free-threaded build of CPython 3.13

lysnikolaou:
## What do these changes do?

Implement support for the free-threaded build of CPython 3.13!

In more detail, this PR does the following:

- Add a CI job that runs the tests under the free-threaded build.
- Enable building free-threaded wheels.
- Add `pythoncapi_compat.h` and move away from all the borrowed reference APIs.
- Mark extension module as thread-safe
- Add locking around `pair_list_global_version`.

There shouldn't be any.

## Related issue number

Related to #1014.

## Checklist

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -204,7 +204,7 @@ jobs:
         allow-prereleases: true
     - name: Setup Python ${{ matrix.pyver }}
       if: endsWith(matrix.pyver, '-freethreading')
-      id: python-install
+      id: python-install-freethreading
       uses: deadsnakes/action@v3.2.0
       with:
         python-version: 3.13-dev
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -196,14 +196,14 @@ jobs:
         path: dist
 
     - name: Setup Python ${{ matrix.pyver }}
-      if: ${{ !endsWith(matrix.pyver, "-freethreading") }}
+      if: ${{ !endsWith(matrix.pyver, '-freethreading') }}
       id: python-install
       uses: actions/setup-python@v5
       with:
         python-version: ${{ matrix.pyver }}
         allow-prereleases: true
     - name: Setup Python ${{ matrix.pyver }}
-      if: endsWith(matrix.pyver, "-freethreading")
+      if: endsWith(matrix.pyver, '-freethreading')
       id: python-install
       uses: deadsnakes/action@v3.2.0
       with:
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -150,7 +150,7 @@ jobs:
     strategy:
       matrix:
         pyver:
-        - "3.13-freethreading"
+        - 3.13-freethreading
         - 3.13
         - 3.12
         - 3.11
@@ -165,9 +165,9 @@ jobs:
         - os: windows
           no-extensions: Y
         - os: macos
-          pyver: "3.13-freethreading"
+          pyver: 3.13-freethreading  # this is still tested within cibuildwheel
         - os: windows
-          pyver: "3.13-freethreading"
+          pyver: 3.13-freethreading  # this is still tested within cibuildwheel
         include:
         - pyver: pypy-3.8
           no-extensions: Y
@@ -200,15 +200,14 @@ jobs:
         path: dist
 
     - name: Setup Python ${{ matrix.pyver }}
-      if: ${{ !endsWith(matrix.pyver, '-freethreading') }}
-      id: python-install
+      if: >-
+        !endsWith(matrix.pyver, '-freethreading')
       uses: actions/setup-python@v5
       with:
         python-version: ${{ matrix.pyver }}
         allow-prereleases: true
     - name: Setup Python ${{ matrix.pyver }}
       if: endsWith(matrix.pyver, '-freethreading')
-      id: python-install-freethreading
       uses: deadsnakes/action@v3.2.0
       with:
         python-version: 3.13-dev
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -150,6 +150,7 @@ jobs:
     strategy:
       matrix:
         pyver:
+        - "3.13-freethreading"
         - 3.13
         - 3.12
         - 3.11
@@ -195,11 +196,20 @@ jobs:
         path: dist
 
     - name: Setup Python ${{ matrix.pyver }}
+      if: ${{ !endsWith(matrix.pyver, "-freethreading") }}
       id: python-install
       uses: actions/setup-python@v5
       with:
         python-version: ${{ matrix.pyver }}
         allow-prereleases: true
+    - name: Setup Python ${{ matrix.pyver }}
+      if: endsWith(matrix.pyver, "-freethreading")
+      id: python-install
+      uses: deadsnakes/action@v3.2.0
+      with:
+        python-version: 3.13-dev
+        nogil: true
+
     - name: Get pip cache dir
       id: pip-cache
       run: |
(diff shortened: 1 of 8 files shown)


bdraco:
The changes look good to me and I didn't see any ref count issues, however I am by no means an expert on the C API

bdraco (approved):

bdraco:
Assuming the CI passes (might take a while to run since there is a lot of activity right now), I'd like another maintainer who has more experience with the C API to take a look before merging.

bdraco:
Thanks @lysnikolaou

bdraco:
It looks like one of the existing actions unrelated to this PR is using an old upload-artifact.  It would be nice if github told us which one.

lysnikolaou:
Yeah, I saw that as well. Opened #1016 for it.

bdraco:
Looks like its almost there 👍

lysnikolaou:
Looks like it's good to go on my side!

bdraco (approved):

bdraco:
LGTM, but I'd still like another set of 👀 https://github.com/aio-libs/multidict/pull/1015#issuecomment-2358761183 before merging

lysnikolaou:
Friendly ping here!

CHANGES/1015.feature.rst
```diff
@@ -0,0 +1 @@
+Implemented support for the free-threaded build of CPython 3.13 -- by :user:`lysnikolaou`.
```
webknjaz, line 1: I think we need another change note of the `packaging` type saying
```rst
Started publishing wheels made for the free-threaded build of CPython 3.13 -- by :user:`lysnikolaou`.
```

@bdraco WDYT?

.github/workflows/ci-cd.yml
```diff
@@ -163,6 +164,10 @@ jobs:
           no-extensions: Y
         - os: windows
           no-extensions: Y
+        - os: macos
+          pyver: "3.13-freethreading"
```
webknjaz, line 168:
```suggestion
          pyver: 3.13-freethreading  # this is still tested within cibuildwheel
```

.github/workflows/ci-cd.yml
```diff
@@ -163,6 +164,10 @@ jobs:
...
+        - os: windows
+          pyver: "3.13-freethreading"
```
webknjaz, line 170:
```suggestion
          pyver: 3.13-freethreading  # this is still tested within cibuildwheel
```

.github/workflows/ci-cd.yml
```diff
@@ -150,6 +150,7 @@ jobs:
     strategy:
       matrix:
         pyver:
+        - "3.13-freethreading"
```
webknjaz, line 153: let's follow the existing convention of not adding unnecessary quoting
```suggestion
        - 3.13-freethreading
```

.github/workflows/ci-cd.yml
```diff
@@ -195,11 +200,20 @@ jobs:
         path: dist
 
     - name: Setup Python ${{ matrix.pyver }}
+      if: ${{ !endsWith(matrix.pyver, '-freethreading') }}
```
webknjaz, line 203:
```suggestion
      if: >-
        !endsWith(matrix.pyver, '-freethreading')
```

webknjaz (commented):
Here's my CI/CD feedback. I can't say I'm experienced with C-API, however. Still, I think we can merge this and start a release process to see if more rare architectures also succeed building the new wheels. It should be enough to see if cibuildwheel passes in those jobs that only run on release, since it also runs some testing. We can then abort the release once we know. Or we could just attempt a regular release.

CHANGES/1015.feature.rst
bdraco: LGTM

multidict/_multilib/pair_list.h
```diff
@@ -53,8 +53,22 @@ typedef struct pair_list {
...
 static uint64_t pair_list_global_version = 0;
+#ifdef Py_GIL_DISABLED
+static PyMutex global_version_mutex;
+#endif
 
+#ifdef Py_GIL_DISABLED
+static inline
+uint64_t NEXT_VERSION(void)
+{
+    PyMutex_Lock(&global_version_mutex);
```
asvetlov, line 64: I think that the mutex is not needed here.

Straight loose increment is enough, the version number is an auxiliary thing.
If the global counter slightly desynchronized between threads for short time period -- it is ok, no existing code is broken.

.github/workflows/ci-cd.yml
```diff
@@ -195,11 +200,20 @@ jobs:
...
       id: python-install
       uses: actions/setup-python@v5
       with:
         python-version: ${{ matrix.pyver }}
         allow-prereleases: true
+    - name: Setup Python ${{ matrix.pyver }}
+      if: endsWith(matrix.pyver, '-freethreading')
+      id: python-install-freethreading
```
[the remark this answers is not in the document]
lysnikolaou: Actually, the `python-install` id is not used anywhere anymore, so I assume that a new version of `codecov` does not need this. I think we can remove these.

multidict/_multilib/pair_list.h
lysnikolaou: Mutex removed. Thanks for the help here!

lysnikolaou:
Thanks for the reviews everyone! I've addressed all of the feedback points.

bdraco:
It looks like we don't have any benchmarks for this repository so can't check the performance impact. I guess I'll need to write some.

.github/workflows/ci-cd.yml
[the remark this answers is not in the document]
lysnikolaou: Oh got it. Unfortunately, `deadsnakes/action` does not have a `python-version` output. Would it make sense to hardcode `3.13` for now?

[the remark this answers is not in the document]
webknjaz: We could have a new step like
```yaml
- name: Compute runtime Python version
  id: python-install
  run: |
    from os import environ
    from pathlib import Path
    from sys import version_info

    FILE_APPEND_MODE = 'a'

    python_version_str = ".".join(map(str, sys.version_info[:3]))

    with Path(environ['GITHUB_OUTPUT']).open(
            mode=FILE_APPEND_MODE,
    ) as outputs_file:
        print(f'python-version={python_version_str}', file=outputs_file)
```
that would work universally.

.github/workflows/ci-cd.yml
```diff
@@ -195,11 +200,19 @@ jobs:
...
       uses: actions/setup-python@v5
       with:
         python-version: ${{ matrix.pyver }}
         allow-prereleases: true
+    - name: Setup Python ${{ matrix.pyver }}
+      if: endsWith(matrix.pyver, '-freethreading')
+      uses: deadsnakes/action@v3.2.0
+      with:
+        python-version: 3.13-dev
+        nogil: true
```
webknjaz, line 214:
```suggestion
        nogil: true
    - name: Compute runtime Python version
      id: python-install
      run: |
        from os import environ
        from pathlib import Path
        from sys import version_info

        FILE_APPEND_MODE = 'a'

        python_version_str = ".".join(map(str, sys.version_info[:3]))

        with Path(environ['GITHUB_OUTPUT']).open(
                mode=FILE_APPEND_MODE,
        ) as outputs_file:
            print(f'python-version={python_version_str}', file=outputs_file)
      shell: python
```

webknjaz: Although, we should probably extend that with a free-threading flag or something.

.github/workflows/ci-cd.yml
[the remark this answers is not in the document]
webknjaz: Committed that. Marking this thread as resolved.

webknjaz (approved):

bdraco:
It looks like the testing fails because there is no cffi wheel for free-threaded 3.13 so it tries to build it and since libffi isn't installed in the cibuildwheel env, it fails

bdraco (approved):

PR #1016, opened 2024-09-18
Upgrade actions/[upload|download]-artifact and other GHA dependencies

lysnikolaou:
## What do these changes do?

- Upgrade `actions/upload-artifact` to v4 since v2 is deprecated and v3 is scheduled to be deprecated in November.

No.

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -187,7 +187,7 @@ jobs:
         workflow-artifact-name: >-
           ${{ needs.pre-setup.outputs.dists-artifact-name }}
     - name: Download distributions
-      if: ${{ !endsWith(matrix.pyver, '-dev') }}
+      if: ${{ !endsWith(matrix.pyver, '-dev') && true || false }}
       uses: actions/download-artifact@v4
       with:
         pattern: ${{ needs.pre-setup.outputs.dists-artifact-name }}*
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -95,7 +95,7 @@ jobs:
         }
         >> "${GITHUB_OUTPUT}"
     - name: Upload built artifacts for testing
-      uses: actions/upload-artifact@v3
+      uses: actions/upload-artifact@v4
       with:
         if-no-files-found: error
         name: ${{ needs.pre-setup.outputs.dists-artifact-name  }}
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -180,7 +180,7 @@ jobs:
 
     steps:
     - name: Retrieve the project source from an sdist inside the GHA artifact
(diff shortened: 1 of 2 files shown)


.github/workflows/reusable-build-wheel.yml
```diff
@@ -42,7 +42,7 @@ jobs:
     timeout-minutes: ${{ inputs.qemu && 47 || 14 }}
     steps:
     - name: Retrieve the project source from an sdist inside the GHA artifact
-      uses: re-actors/checkout-python-sdist@release/v1
+      uses: re-actors/checkout-python-sdist@187f55296b0f54d88259aaaf99af32ad3647d3bc  # Current main that uses download-artifact@v4
```
webknjaz, line 45:
```suggestion
      uses: re-actors/checkout-python-sdist@release/v2
```

.github/workflows/ci-cd.yml
```diff
@@ -180,15 +180,15 @@ jobs:
 
     steps:
     - name: Retrieve the project source from an sdist inside the GHA artifact
-      uses: re-actors/checkout-python-sdist@release/v1
+      uses: re-actors/checkout-python-sdist@187f55296b0f54d88259aaaf99af32ad3647d3bc  # Current main that uses download-artifact@v4
```
webknjaz, line 183:
```suggestion
      uses: re-actors/checkout-python-sdist@release/v2
```

webknjaz (commented):
Also, look into the yarl PRs where I did this properly (and the follow-ups!). This needs unique names for the artifacts.

lysnikolaou:
@webknjaz Thanks a lot for the review and the help! I think we're good to go on this now.

.github/workflows/ci-cd.yml
```diff
@@ -180,17 +180,18 @@ jobs:
...
+      uses: re-actors/checkout-python-sdist@release/v2
       with:
         source-tarball-name: >-
           ${{ needs.build-pure-python-dists.outputs.sdist-filename }}
         workflow-artifact-name: >-
           ${{ needs.pre-setup.outputs.dists-artifact-name }}
     - name: Download distributions
-      if: ${{ !endsWith(matrix.pyver, '-dev') && true || false }}
-      uses: actions/download-artifact@v3
+      if: ${{ !endsWith(matrix.pyver, '-dev') }}
```
webknjaz, line 190: This change doesn't seem necessary. Best submit it separately.

webknjaz (commented):
I don't see any immediate problems, reviewing from phone. But I'd rather double-check from the laptop later..

.github/workflows/ci-cd.yml
lysnikolaou: Yeah, I though that it could go in since it's a simple one-line improvement. Removed.

bdraco (approved):

bdraco:
Thanks @lysnikolaou

PR #1020, opened 2024-10-06
Remove missed compat code for Python older than 3.8 from tests

bdraco:
https://github.com/aio-libs/multidict/pull/1010#discussion_r1788762690


webknjaz (approved):

bdraco:
Looks like we need to exclude 3.8 wheels before we can merge this

bdraco:
Oh its actually still needed for py3.8.  We have to wait on this one

PR #1029, opened 2024-10-16
Add codspeed benchmarks

bdraco:
We are planning on merging free-threading support in https://github.com/aio-libs/multidict/pull/1015 but have no good way to test this does not regress performance. Add benchmarks ahead of merge so we can get a test in the CI


bdraco:
RTD build failed for some reason and I can't get it to run so doing another PR https://github.com/aio-libs/multidict/pull/1030

PR #1033, opened 2024-10-30
Upgrade codspeed to 3.0.0

bdraco:
SSIA


bdraco:
We need to drop python 3.8 first

bdraco:
replaced by https://github.com/aio-libs/multidict/pull/1031

Issue #965, opened 2024-04-11
Case-insensitivity of keys doesn't extend to set operations


Dreamsorcerer:
Would be nice to have this work correctly:
```
>>> d = multidict.CIMultiDict()
>>> d["CONTENT-LENGTH"] = "4"
>>> "Content-Length" in d.keys()
True
>>> {"Content-Length"} & d.keys()
set()  # Should get a result here
>>> {"CONTENT-LENGTH"} & d.keys()
{'CONTENT-LENGTH'}
```

I tried to do this in aiohttp to check if any of a set of headers is present.

asvetlov:
it seems that the feature request requires case insensitive set implemented first.
In turn, it raises additional questions:
What is the type of `CISet() & set()`? Should it be *ci set* or bare *set*? The same for `set() & CISet()`?

One possible solution could be returning CI Set for all binary operations if one of two arguments is also CI Set.

asvetlov:
It seems that we cannot guarantee that CI Set is always returned; the type of `a & b` is `type(a)` for sets at least for their ABCs:
```
>>> frozenset(['a']) & {'b'}
frozenset()
>>> {'b'} & frozenset(['a'])
set()
>>> 
```

Dreamsorcerer:
I think that can be changed by ussing `__rand__`, can't it?

But, I don't think it's necessary to have a CISet here, the main thing is just to ensure that the keys are returned. So, adding `__add__`/`__or__` to KeysView and returning a set which contains the items in `other` which case-insensitively match keys in the KeysView.

asvetlov:
Fixed by #1038

PR #1038, opened 2024-12-03
Properly support set operations for case insensitive multidict views

asvetlov:
For #965


tests/test_multidict.py
```diff
@@ -823,3 +823,14 @@ def test_keys__repr__(self, cls: Type[CIMultiDict[str]]) -> None:
...
+    @pytest.mark.xfail(reason="CI sets are not implemented yet")
+    def test_keys_case_insensitive_union(
+        self,
+        cls: Type[CIMultiDict[str]],
+    ) -> None:
+        d = cls([("KEY", "one"),])
+
+        assert "key" in d.keys()
+        assert d.keys().isdisjoint({"key"})
+        assert {"key"} & d.keys() == {"KEY"}
```
Dreamsorcerer, line 836: I think it may be safer to return the values from the set, rather than the keys:
```suggestion
        assert {"key"} & d.keys() == {"key"}
```

This makes it predictable what the exact values will be in the set, if a developer wants to get them afterwards (e.g. it means they are normalised to whatever the developer is looking for).

asvetlov: Good suggestion, thanks!

multidict/_multidict_py.py
```diff
@@ -27,6 +27,11 @@
 else:
     from typing_extensions import Self
 
+if sys.version_info >= (3, 10):
+    from types import NotImplementedType
+else:
+    from typing import Any as NotImplementedType
```
webknjaz, line 33: Try
```suggestion
    NotImplementedType = object
```
to see if it works.

multidict/_multidict_py.py
```diff
@@ -530,6 +568,14 @@ def _replace(self, key: str, value: _V) -> None:
 class CIMultiDict(MultiDict[_V]):
     """Dictionary with the support for duplicate case-insensitive keys."""
 
+    def _key(self, key: str) -> str:
+        if isinstance(key, str):
+            if getattr(type(key), '__is_istr__', False):
+                return key
+            else:
+                return istr(key)
```
Dreamsorcerer, line 576: I'm not clear I understand this part? Is there expected to be some other custom types implemented by users?

Seems to me like if the condition is true, then the type is probably istr, in which case it's be the same to do:
```suggestion
            return key
```

multidict/_multidict_py.py
Dreamsorcerer: Typeshed actually types it as a subclass of Any, so it might make more sense to keep it like this. Regardless, we'll be dropping the code end of this year, so not worth spending time on.

Dreamsorcerer:
@asvetlov Please double check my last commit. That fixes the type errors.

multidict/_multidict_py.py
```diff
@@ -160,6 +169,38 @@ def __repr__(self) -> str:
...
+    def __or__(self, other: Iterable[_T]) -> Union[set[Union[str, _T]], NotImplementedType]:
+        ret: set[Union[str, _T]] = set(self)
+        try:
+            it = iter(other)
+        except TypeError:
+            return NotImplemented
+        for key in it:
+            if not isinstance(key, str):
+                ret.add(key)
+                continue
```
Dreamsorcerer, line 194: As the runtime behaviour was incorrect in this case, probably want another test to verify any C code works correctly too.

multidict/_multidict_py.py
```diff
@@ -178,6 +179,104 @@ def __repr__(self) -> str:
         body = ", ".join(lst)
         return f"<{self.__class__.__name__}({body})>"
 
+    def __and__(self, other: Iterable[str]) -> set[str]:
```
Dreamsorcerer, line 182: All of these str parameters should be changed to object, as they are type safe with any type. This should also resolve the unreachable errors in mypy.
```suggestion
    def __and__(self, other: Iterable[object]) -> set[str]:
```

multidict/_multidict_py.py
asvetlov: Dropped

multidict/_multidict_py.py
asvetlov: tests are added

multidict/_multidict_py.py
asvetlov: done, thanks for the suggestion

asvetlov:
C implementation for KeysView is ready.
ItemsView requires a little more work.

asvetlov:
Hmm, the implementation is slightly incorrect, it always should prefer overlapping keys from the left operand.
`md.keys() & {'key'}` should select keys from multidict but `{'key'} & md.keys()` should select from the set.

asvetlov:
Now Python impl and tests are good.

Dreamsorcerer:
> `md.keys() & {'key'}` should select keys from multidict but `{'key'} & md.keys()` should select from the set.

Are you sure about that? I think we discussed this before somewhere, and I suggested as a developer that if you are intersecting with `'key'`, it would be less expected that you get something that is not `'key'`.

e.g. I can imagine code like this:
```
if (md.keys() & {"key"}) == {"key"}:
# Or
if ({"key"} & md.keys()) == {"key"}:
```

I don't think think the order should matter here, it should always produce a result from the user's set. If it produces `'Key'` from the CI dict, then the above code would never work.

Dreamsorcerer:
The above could also be safer if we had a CISet that was returned in this case and made the equality comparison safer regardless. But, that might be overkill to create a whole new set class.

asvetlov:
> > `md.keys() & {'key'}` should select keys from multidict but `{'key'} & md.keys()` should select from the set.
> 
> Are you sure about that? I think we discussed this before somewhere, and I suggested as a developer that if you are intersecting with `'key'`, it would be less expected that you get something that is not `'key'`.
> 
> e.g. I can imagine code like this:
> 
> ```
> if (md.keys() & {"key"}) == {"key"}:
> # Or
> if ({"key"} & md.keys()) == {"key"}:
> ```
> 
> I don't think think the order should matter here, it should always produce a result from the user's set. If it produces `'Key'` from the CI dict, then the above code would never work.

`if ({"key"} & md.keys()) == {"key"}:` will try `set.__and__(md)` first, and it succeeds because `set()` could accept any iterable as the right argument.
In turn, `if (["key"] & md.keys()) == {"key"}:` first tries `list.__and__(md)` which returns `NotImplemented`. Than Python applies `md.__rand__(list)` which succeeds.

I really don't want to rely on this subtle logic. It is better to establish that the left argument controls that keys are used if they are present in both left and right arguments.

If multidict operators are called, they make sure that the comparison is performed on identities, it works well with case insensitivity. But the type is established by the left arg.

Dreamsorcerer:
> It is clear that `{'k1', 'k2'} - md.keys()` should return a subset from the left argument, and `md.keys() - {'k1', 'k2'}` returns a subset of `md` keys, isn't it?

Well, that's because the right side is never expected to have anything returned.

I still think the `&` can be pretty confusing if the keys in the user set are not included, but I'm not 100% convinced my suggestion is the correct one, so I won't argue it any further.

asvetlov:
I had both variants implemented, and I've found that the PR's version is less controversial (but neither is 100% perfect, sure).

asvetlov:
Regarding performance, I honestly don't care about the `xor` operation.
I suggest committing as-is, if somebody needs better numbers we could do it in a separate PR.
XOR is an unpleasant operation in terms of algorithmic complexity anyway. I had a variant with adding all items from the left, iterating over the right, discarding overlapping items and adding missing.
But it either requires a temporary dict/set, or has O(N**2) complexity :(

bdraco:
XOR is going to be a rare enough operation that I don't think its a problem.

The speed up to the OR and AND operations are likely going to outweigh any loss to XOR performance for most cases.

asvetlov:
Let's make an agreement that the performance of the items view is not very important (while it could be improved, sure).
The operations are supported because ItemsView ABC is declared to support them.
In the practical sense, I don't recall any application of this facility in aiohttp world.

bdraco:
> Let's make an agreement that the performance of the items view is not very important (while it could be improved, sure). The operations are supported because ItemsView ABC is declared to support them. In the practical sense, I don't recall any application of this facility in aiohttp world.

I looked though aiohttp and I didn't find any of these type of operations in aiohttp hot paths so I think its fine.

The other changes you made in the past few days are likely to be much more impactful in improving performance.

Let's merge this, and I can do a new release of multidict later today, update it in aiohttp and we can get a better idea of what the performance impacts are.

If I'm wrong and we have these operations somewhere in an aiohttp hot path, we can revisit.  I don't think thats the case however.

bdraco (approved):

bdraco:
Looks like there is a regression in the pure-python version https://github.com/aio-libs/multidict/issues/1111

bdraco:
tests in https://github.com/aio-libs/multidict/pull/1112

PR #1043, opened 2025-02-15
Test removing redundant code

Dreamsorcerer:
At first glance, this code looks redundant. Running the tests first to see if it catches any problems.

### DIFF
--- a/multidict/_multidict_base.py
+++ b/multidict/_multidict_base.py
@@ -59,48 +59,24 @@ def _viewbaseset_richcmp(view, other, op):
 def _viewbaseset_and(view, other):
     if not isinstance(other, Iterable):
         return NotImplemented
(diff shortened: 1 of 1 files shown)


multidict/_multidict_base.py
```diff
@@ -95,10 +83,6 @@ def _viewbaseset_sub(view, other):
 def _viewbaseset_xor(view, other):
     if not isinstance(other, Iterable):
         return NotImplemented
-    if isinstance(view, Set):
```
bdraco, line 98: I think only the first condition is duplicate ?

Dreamsorcerer: It's more that it checks if it's a set and then converts it to a set. But, this seems to be required to work properly...

PR #1045, opened 2025-02-16
Fix __contains__ with unexpected types

Dreamsorcerer:
Found this while adding inline type annotations for mypy.

### DIFF
--- a/CHANGES/1045.bugfix.rst
+++ b/CHANGES/1045.bugfix.rst
@@ -1,1 +1,1 @@
-Fixed `__contains__` throwing an exception for non strings.
+Fixed ``__contains__`` throwing an exception for non strings.
--- a/CHANGES/1045.bugfix.rst
+++ b/CHANGES/1045.bugfix.rst
@@ -1,1 +1,1 @@
-Fixed ``in`` checks throwing an exception instead of returning ``False`` when testing non-strings.
+Fixed ``in`` checks throwing an exception instead of returning :data:`False` when testing non-strings.
--- a/CHANGES/1045.bugfix.rst
+++ b/CHANGES/1045.bugfix.rst
@@ -1,1 +1,1 @@
-Fixed ``__contains__`` throwing an exception for non strings.
+Fixed ``in`` checks throwing an exception instead of returning ``False`` when testing non-strings.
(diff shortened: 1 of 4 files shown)


CHANGES/1045.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixed `__contains__` throwing an exception for non strings.
```
webknjaz, line 1: RST?

CHANGES/1045.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixed ``__contains__`` throwing an exception for non strings.
```
webknjaz, line 1: Not sure how to best rephrase it. I think it might give people more context if this was saying something like "There was this specific exception raise via `in` checks, and it is no longer happening". Or "The `in` check used to raise exception X when non-string value presence was being checked. It is now fixed to return false."
Ideas?

tests/test_multidict.py
```diff
@@ -224,6 +224,23 @@ def test__iter__(
...
+    ) -> None:
+        d = cls([("key", "one"), ("key2", "two"), ("key", 3)])
+
+        assert list(d) == ["key", "key2", "key"]
+
+        assert "key" in d
+        assert "key2" in d
+
+        assert "foo" not in d
+        assert 42 not in d
```
webknjaz, line 242: Can we also try the opposite code path?
```suggestion
        assert 42 not in d
        assert not (42 in d)
```
(this will probably need a noqa comment)

Dreamsorcerer: If you want, they are fundamentally the same though (i.e. it won't test anything different in our code):
https://docs.python.org/3/reference/expressions.html#not-in

webknjaz: Ah, right. I didn't realize there's no separate dunder method for that. Marking as resolved.

CHANGES/1045.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixed ``in`` checks throwing an exception instead of returning ``False`` when testing non-strings.
```
webknjaz, line 1: Is this just in multidict instances or proxies too?

webknjaz:
```suggestion
Fixed ``in`` checks throwing an exception instead of returning :data:`False` when testing non-strings.
```

asvetlov (approved):
Nice catch!

PR #929, opened 2024-01-20
Fix SEGFAULTs under Python 3.12 by using public C-API

charles-dyfis-net:
## What do these changes do?

Fix build issues (observed personally with clang 16.0.6 as wrapped by nixpkgs when building multidict via [poetry2nix](https://github.com/nix-community/poetry2nix), and reported by @psss to also happen on gcc-14) related to changes to the `_PyArg_Parser` structure by moving away from this internal interface to use a slower, fully-supported interface instead.

## Are there changes in behavior for the user?

This change has a performance penalty.

## Related issue number

Trivial modification to the work done in #909.
Fixes #926.

## Checklist

- [X] I think the code is well written
- [ ] Unit tests for the changes exist
- [ ] Documentation reflects the changes

Issue #926, opened 2024-01-18
Unable to `pip install multidict` with `gcc-14`


psss:
## Long story short

With the latest Fedora Rawhide which freshly contains `gcc-14` I'm not able to install `multidict`.

## Expected behaviour

Package successfully installed.

## Actual behaviour

I see the following error:

```
gcc -fno-strict-overflow -Wsign-compare -DDYNAMIC_ANNOTATIONS_ENABLED=1 -DNDEBUG -fexceptions -fcf-protection -fexceptions -fcf-protection -fexceptions -fcf-protection -fPIC -I/tmp/venv/include -I/usr/include/python3.12 -c multidict/_multidict.c -o build/temp.linux-x86_64-cpython-312/multidict/_multidict.o -O2 -std=c99 -Wall -Wsign-compare -Wconversion -fno-strict-aliasing -pedantic
multidict/_multidict.c: In function ‘multidict_getall’:
multidict/_multidict.c:458:37: error: initialization of ‘int’ from ‘void *’ makes integer from pointer without a cast [-Wint-conversion]
  458 |     static _PyArg_Parser _parser = {NULL, _keywords, "getall", 0};
      |                                     ^~~~
multidict/_multidict.c:458:37: note: (near initialization for ‘_parser.initialized’)
multidict/_multidict.c:458:43: error: initialization of ‘const char *’ from incompatible pointer type ‘const char * const*’ [-Wincompatible-pointer-types]
  458 |     static _PyArg_Parser _parser = {NULL, _keywords, "getall", 0};
      |                                           ^~~~~~~~~
multidict/_multidict.c:458:43: note: (near initialization for ‘_parser.format’)
multidict/_multidict.c:458:54: error: initialization of ‘const char * const*’ from incompatible pointer type ‘char *’ [-Wincompatible-pointer-types]
  458 |     static _PyArg_Parser _parser = {NULL, _keywords, "getall", 0};
      |                                                      ^~~~~~~~
multidict/_multidict.c:458:54: note: (near initialization for ‘_parser.keywords’)
multidict/_multidict.c: In function ‘multidict_getone’:
multidict/_multidict.c:503:37: error: initialization of ‘int’ from ‘void *’ makes integer from pointer without a cast [-Wint-conversion]
  503 |     static _PyArg_Parser _parser = {NULL, _keywords, "getone", 0};
      |                                     ^~~~
multidict/_multidict.c:503:37: note: (near initialization for ‘_parser.initialized’)
[...]

PR #1070, opened 2025-03-18
Implement a custom parser for METH_FASTCALL | METH_KEYWORDS protocol

asvetlov:
The story is:

Back in the day, multidict used the `METH_VARARGS | METH_KEYWORDS` calling protocol for its methods.
Later, the library switched to `METH_FASTCALL | METH_KEYWORDS` protocol.
The new method is significantly faster than previously used, especially if arguments are passed as positional. 

Unfortunately, parsing FASTCALL arguments is not a trivial task; and Python doesn't provide a public parser API.
Thus, `multidict` used to utilize a private `_PyArg_ParseStackAndKeywords` / `_PyArg_UnpackKeywords`.
The downside is obvious: a private API doesn't guarantee to work in future Python versions.
After a segfault with Python 3.12 (https://github.com/aio-libs/multidict/issues/926), https://github.com/aio-libs/multidict/pull/929 patch was applied to switch back to `METH_VARARGS | METH_KEYWORDS` as a temporary fix.

While making a generic FASTCALL parser is challenging, multidict doesn't need it, honestly. The library's methods are called with two arguments (`.add()`), the second arg often has a default value (`.get()`). A parser for such a case is simple; the PR provides it. The parser doesn't depend on any non-public function, and it is compatible with any future Python versions as FASTCALL is a fixed-in-stone calling protocol.

### DIFF
--- a/tests/test_mutable_multidict.py
+++ b/tests/test_mutable_multidict.py
@@ -318,6 +318,37 @@ class TestMutableMultiDict:
 
         assert {"key" + str(SIZE - 1): SIZE - 1} == d
 
+    def test_update(
+        self,
+        case_sensitive_multidict_class: type[CIMultiDict[Union[str, int]]],
+    ) -> None:
+        d = case_sensitive_multidict_class()
+        assert d == {}
+
+        d.update([("key", "one"), ("key", "two")], key=3, foo="bar")
+        assert d != {"key": "one", "foo": "bar"}
+        assert 4 == len(d)
+        itms = d.items()
+        # we can't guarantee order of kwargs
+        assert ("key", "one") in itms
+        assert ("key", "two") in itms
+        assert ("key", 3) in itms
+        assert ("foo", "bar") in itms
+
+        other = case_sensitive_multidict_class(bar="baz")
+        assert other == {"bar": "baz"}
+
+        d.update(other)
+        assert ("bar", "baz") in d.items()
+
+        d.update({"foo": "moo"})
+        assert ("foo", "moo") in d.items()
+
+        d.update()
+        assert 5 == len(d)
+
+        with pytest.raises(TypeError):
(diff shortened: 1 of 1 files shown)


bdraco:
Very nice.  I don't think we have any coverage for passing incorrect arguments.  Not sure if we need that though.

bdraco:
Testing

- [x] run on production
- [x] run the tests I wrote in https://github.com/aio-libs/multidict/pull/1072 against it (not sure it actually matters)
- [x] checkout aiohttp master and run the tests against it
- [ ] checkout aiohttp 3.11 and run the tests against it

bdraco (approved):

multidict/_multilib/parser.h
```diff
@@ -0,0 +1,147 @@
...
+}
+
+
+
+/* Parse FASTCALL|METH_KEYWORDS arguments as two args,
+the first arg is mandatory and the second one is optional.
+If the second arg is not passed it remains NULL pointer.
+
+The parser accepts three forms:
+1. all posiotional args,
```
bdraco, line 31:
```suggestion
1. all positional args,
```

multidict/_multilib/parser.h
```diff
@@ -0,0 +1,147 @@
...
+                  PyObject *kwnames,
+                  Py_ssize_t minargs,
+                  const char* arg1name,
+                  PyObject **arg1,
+                  const char* arg2name,
+                  PyObject **arg2
+)
+{
+    assert(maxargs>=1);
+    assert(maxargs<=2);
```
bdraco, line 48: `maxargs` ?

bdraco:
Tested on a second production system ✅

multidict/_multilib/parser.h
asvetlov: Or is it better to keep it as-is, these assertions exist for checking that the function was called with a correct compile-time constant.

PR #1074, opened 2025-03-19
Tune coverage setup

asvetlov:
The PR reduces false positives in coverage reports

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,3 +12,14 @@ free-threaded-support = true
 
 [tool.cibuildwheel.linux]
 before-all = "yum install -y libffi-devel || apk add --upgrade libffi-dev || apt-get install libffi-dev"
+
+
+[tool.coverage.run]
+branch = true
+source_pkgs = ['multidict']
+
+[tool.coverage.report]
+exclude_also = [
+    # @overload or protocol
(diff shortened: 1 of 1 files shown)


asvetlov:
@webknjaz, if you'll fix the coverage report, I'll be happy with it.
Now, the tool is broken, even though it is synced with others.
Merged red PRs bother me.

P.S. Sorry, I don't understand the current codecov config. As for me, it looks overcomplicated.
Why should I care about the tests coverage? What is mypy coverage, and why should I care about it, especially in a strict mode?

asvetlov:
@webknjaz if you are talking about reverting back `.coveragerc` instead of moving this part of config into `pyproject.toml` -- I'm ok with it.

Also, could you please explain why https://app.codecov.io/gh/aio-libs/multidict/blob/master/multidict%2F_abc.py#L21 is marked as partially covered? From my understanding, this line should be just ignored.

webknjaz:
> Why should I care about the tests coverage?

https://nedbatchelder.com/blog/201908/dont_omit_tests_from_coverage.html

webknjaz:
@asvetlov file an issue with specific things that you think might need to be improved and I can try to find time to make better changes to the config.

PR #1077, opened 2025-03-19
Improve coverage

asvetlov:


.coveragerc
```diff
@@ -33,3 +33,5 @@ source =
   .
 source_pkgs =
   multidict
+omit =
+  tests/gen_pickles.py
```
webknjaz, line 37: We should probably handle this differently. I've been thinking of refactoring that in a way that wouldn't compromise coverage. There's two PRs exploring this and an issue. But I think that after all, I'd look into delegating this to a pytest plugin.

```suggestion
  # FIXME: migrate to a snapshotting plugin. See #922, #924 and #938.
  tests/gen_pickles.py
```

tests/test_circular_imports.py
```diff
@@ -51,7 +51,7 @@ def _discover_path_importables(
         if pkg_dir_path.parts[-1] == "__pycache__":
             continue
 
-        if all(Path(_).suffix != ".py" for _ in file_names):
+        if all(Path(_).suffix != ".py" for _ in file_names):  # pragma: no cover
```
webknjaz, line 54: Make motivation obvious through code comments:

```suggestion
        if all(Path(_).suffix != ".py" for _ in file_names):  # It's never empty # pragma: no branch
```

webknjaz (changes requested):

.coveragerc
asvetlov: I'd rather suggest moving the generator into `./tools` folder.

webknjaz: If a proper pytest plugin is used, we won't even need a separate script. It'd be better integrated.

webknjaz: But regardless of that, the ommission deserves a comment with justification.

multidict/_multidict_base.py
```diff
@@ -73,59 +76,43 @@ def _viewbaseset_richcmp(
...
-    if isinstance(view, Set):
-        view = set(iter(view))
-    if isinstance(other, Set):
-        other = set(iter(other))
-    if not isinstance(other, Set):
-        other = set(iter(other))
-    return view & other
+    lft = set(iter(view))
+    rgt = set(iter(other))
+    return lft & rgt  # type: ignore[return-value]
```
Dreamsorcerer, line 85: Why are we getting a type error here now?

asvetlov: Good question.

Dreamsorcerer (approved):
I noticed those oddities earlier, but was too busy to spend any time checking if it was safe to simplify, thanks.

multidict/_multidict_base.py
Dreamsorcerer: > got "set[Sequence[object]]"

That doesn't sound right to me..

asvetlov: Because it is applicable to both KeysView and ItemsView.
ItemsView returns `tuple[str, T]`.

Honestly, I don't think that this file should have type annotations.
It is never called from python but from C Extension.
mypy check is useless here; it doesn't analyze the whole path, and it has no idea what types were used by C Extension during the call.

Dreamsorcerer: OK, KeysView[str] is going to make an iterator of strs, while ItemsView[str, object] is going to make an iterator of tuple[str, object]. The common supertype of those is Iterator[Sequence[object]] (as str is a sequence itself).

Dreamsorcerer, line 85:
```suggestion
    lft: set[object] = set(iter(view))
    rgt = set(iter(other))
    return lft & rgt
```

Should solve it.

multidict/_multidict_base.py
```diff
@@ -73,59 +76,43 @@ def _viewbaseset_richcmp(
...
-    if isinstance(view, Set):
-        view = set(iter(view))
-    if isinstance(other, Set):
-        other = set(iter(other))
-    if not isinstance(other, Set):
-        other = set(iter(other))
-    return view - other
+    lft = set(iter(view))
+    rgt = set(iter(other))
+    return lft - rgt  # type: ignore[return-value]
```
Dreamsorcerer, line 105:
```suggestion
    lft: set[object] = set(iter(view))
    rgt = set(iter(other))
    return lft - rgt
```

Dreamsorcerer: We could technically be smarter about this with overload and TypeVars, but not sure how much benefit it'd bring.
e.g.
```
@overload
def _viewbaseset_and(
    view: KeyView[str], other: set[T]
) -> set[Union[str, T]]:
    ...
@overload
def _viewbaseset_and(
    view: ItemsView[str, T], other: set[U]
) -> set[Union[tuple[str, T], U]]:
    ...
```

Dreamsorcerer:
We could technically be smarter about this with overload and TypeVars, but not sure how much benefit it'd bring.
e.g.
```
@overload
def _viewbaseset_and(
    view: KeyView[str], other: set[T]
) -> set[Union[str, T]]:
    ...
@overload
def _viewbaseset_and(
    view: ItemsView[str, T], other: set[U]
) -> set[Union[tuple[str, T], U]]:
    ...
```

tests/test_circular_imports.py
```diff
@@ -51,7 +51,7 @@ def _discover_path_importables(
         if pkg_dir_path.parts[-1] == "__pycache__":
             continue
 
-        if all(Path(_).suffix != ".py" for _ in file_names):
+        if all(Path(_).suffix != ".py" for _ in file_names):  # It's never empty # pragma: no branch
```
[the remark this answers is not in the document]
asvetlov: Done

multidict/_multidict_base.py
```diff
@@ -73,59 +77,43 @@ def _viewbaseset_richcmp(
...
+    lft = set(iter(view))
+    rgt = set(iter(other))
+    return lft - rgt
 
 
 def _viewbaseset_xor(
-    view: set[object], other: object
-) -> Union[set[object], NotImplementedType]:
+    view: _ViewArg, other: object
+) -> Union[set[Sequence[object]], NotImplementedType]:
```
Dreamsorcerer, line 111: I don't think you want to add Sequence here. If other is tuple[int], then that return type is not correct.

Dreamsorcerer: I'm not sure why mypy is allowing it though...

Dreamsorcerer: Nevermind, you're right as these particular methods only return items that are in the left side.

asvetlov: Technically, Any is the best type. I don't want typing here; it doesn't make any value and doesn't help mypy to find the misusage

Dreamsorcerer: The return type is probably less relevant, but the parameter types help to check for and catch mistakes such as #1045.

multidict/_compat.py
```diff
@@ -10,5 +10,6 @@
 if USE_EXTENSIONS:
     try:
         from . import _multidict  # type: ignore[attr-defined]  # noqa: F401
-    except ImportError:
+    except ImportError:  # pragma: no branch
```
webknjaz, line 13: Looks like this doesn't work for some reason. Maybe, I was wrong, and it does indeed need a “no cover”. This has to be checked because I was commenting from mobile earlier w/ no way to verify. I suppose coveragepy might not treat this as a branch, after all.

asvetlov: updated

multidict/_multidict_base.py
```diff
@@ -34,34 +38,26 @@ def _abc_valuesview_register(view_cls: type[object]) -> None:
 
 
 def _viewbaseset_richcmp(
-    view: set[object], other: object, op: Literal[0, 1, 2, 3, 4, 5]
+    view: _ViewArg, other: object, op: Literal[0, 1, 2, 3, 4, 5]
 ) -> Union[bool, NotImplementedType]:
+    if not isinstance(other, Set):
+        return NotImplemented  # type: ignore[no-any-return]
```
webknjaz, line 44: I don't understand what requires this type ignore. @Dreamsorcerer do you know? Or should the return type switch to something like
```suggestion
) -> Union[bool, type[ImplementedType]]:
    if not isinstance(other, Set):
        return NotImplemented
```
?

webknjaz: Oh, so it must be `Any as NotImplementedType` that's at fault.

webknjaz: Can we replace that import with `NotImplementedType = object`? Since we don't run MyPy against Python 3.9 anymore (which I still don't like, though), it wouldn't matter for the time being.

asvetlov: Please hold on.
I feel this file will go away in a few days.

Dreamsorcerer: It's the NotImplemented (not the type annotation). It seems to vary in some Python versions which is confusing, some Python versions it's reported as Any, and others it's not.

Regardless, that's not part of this PR, I already put that it. I don't think it's fixable at this point, but maybe it will change in a future Python/mypy release..

webknjaz (approved):
🛳️

PR #1078, opened 2025-03-19
Don't create Python exception if key is not found but default value is provided

asvetlov:


bdraco:
Another nice optimization! Will test it on production later today.

bdraco:
Testing

- [x] run on production 1
- [ ] run on production 2
- [ ] checkout aiohttp master and run the tests against it
- [ ] checkout aiohttp 3.11 and run the tests against it

bdraco (approved):

PR #1079, opened 2025-03-19
Add benchmarks for views and reprs

asvetlov:


bdraco:
The test variability on extend seems to be the cost of the realloc

Screenshot 2025-03-19 at 10 03 46 AM
Screenshot 2025-03-19 at 10 03 57 AM

asvetlov:
@bdraco agree.
Please note, I didn't touch this test. 
Let's ignore it as a false positive

bdraco:
> @bdraco agree. Please note, I didn't touch this test. Let's ignore it as a false positive

Sorry for the lack of clarity, its definitely attributable to variability in the benchmark and not this change.

bdraco (approved):

PR #1081, opened 2025-03-19
Implement __repr__ for C Extension classes in C

asvetlov:


CHANGES/1081.feature
```diff
@@ -0,0 +1,4 @@
+
```
webknjaz, line 1:
```suggestion
```

CHANGES/1081.feature
```diff
@@ -0,0 +1,4 @@
...
+Implement ``__repr__`` for C Extension classes in C.
```
webknjaz, line 2: Don't forget that the change log isn't supposed to look like a TODO-list or a Git log. It's supposed to reference things that have already changed compared to the previous release. Therefore, the use of imperative mood is incompatible with this document type.
The past tense or present simple works best, when explaining *to the end-users* how the changes affect *them*. This is your audience, not contributors.
```suggestion
Implemented :py:meth:`~object.__repr__` interface in the C Extension-based :py:class:`~multidict.Multidict` implementation.
```

asvetlov: I've applied past tense.

Sorry, I disagree. with the other proposed changes.  A link to `__repr__` doesn't add too much for the reader. A link to `MultiDict` is not complete, the PR modifies reprs for all `multidict` classes except `istr`. Let's keep the text simple.

CHANGES/1081.feature
webknjaz, line 1: As a rule of thumb, always check if the preview looks consistent with the surrounding change log via RTD previews: https://multidict--1081.org.readthedocs.build/en/1081/changes/#to-be-included-in-vrelease-if-present.

CHANGES/1081.feature
webknjaz: Marking as resolved.

CHANGES/1081.feature
```diff
@@ -0,0 +1,3 @@
+Implemented ``__repr__`` for C Extension classes in C.
```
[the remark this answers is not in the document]
asvetlov: Ok, thanks!

PR #1084, opened 2025-03-20
Always return new references in C Extension

asvetlov:
Borrowed references to Python objects are dangerous.
The PR constantly uses a new-reference strategy and newer returns a borrowed ref.


asvetlov:
Hmm, very strange. INCREF/DECREF cannot be so slow!

asvetlov:
Ok, the approach doesn't work for an item finding.
Internal `pair_list.h` utilities shouldn't incref/decref

PR #1085, opened 2025-03-20
Mitigate too high standard deviation in benchmarks

asvetlov:
Benchmarks for `md.add()` and `md.extend()` had a standard deviation ~ 500%.
It leads to false reports because the test execution time was very unstable.

The PR reduces stddev for these tests back to ~10%


asvetlov:
Perf change is expected; each affected test is executed 100 times now.

PR #1086, opened 2025-03-20
Optimize list comparison

asvetlov:


CHANGES/1084.misc
```diff
@@ -0,0 +1,2 @@
+
+Refactoring of ``pair_list`` implementation.
```
webknjaz, line 2: This should reflect the impact the end-users would observe having updated to the version including the change. Also, the `-ing` form is not a good fit here because it should describe what has already changed, not what's in the process of changing.

asvetlov: Please feel free to reword

asvetlov: Note, the change doesn't affect any end-user's code.

asvetlov: Updated

webknjaz: Yes, that's why it's “misc”. But still, putting very vague generic statements is not really helpful in the change log document. Being more specific could make it useful, though.

PR #1089, opened 2025-03-20
Add benchmarks for creating MultiDict objects

### LINKED ISSUE #1088: Add benchmarks for creating various MultiDict instances
py-spy shows we spent a bit of time creating these objects

Image

Image

https://github.com/aio-libs/aiohttp/blob/7b79f0c02e91fd172aa0d75a3d8315f0717b5c2d/aiohttp/payload.py#L168
https://github.com/aio-libs/aiohttp/blob/7b79f0c02e91fd172aa0d75a3d8315f0717b5c2d/aiohttp/client.py#L1138
https://github.com/aio-libs/aiohttp/blob/7b79f0c02e91fd172aa0d75a3d8315f0717b5c2d/aiohttp/client.py#L1141

bdraco:
closes #1088

### DIFF
--- a/tests/test_multidict_benchmarks.py
+++ b/tests/test_multidict_benchmarks.py
@@ -4,7 +4,7 @@ from typing import Dict, Union
 
 from pytest_codspeed import BenchmarkFixture
 
-from multidict import CIMultiDict, MultiDict, istr
+from multidict import CIMultiDict, MultiDict, MultiDictProxy, istr
 
 # Note that this benchmark should not be refactored to use pytest.mark.parametrize
 # since each benchmark name should be unique.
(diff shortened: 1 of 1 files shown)


bdraco:
already looks quite performant.  Interesting that it shows up on the `py-spy` in the issue

tests/test_multidict_benchmarks.py
```diff
@@ -410,3 +410,56 @@ def test_multidict_repr(benchmark: BenchmarkFixture) -> None:
     @benchmark
     def _run() -> None:
         repr(md)
+
+
+def test_create_empty_multidict(benchmark: BenchmarkFixture) -> None:
+    @benchmark
+    def _run() -> None:
+        MultiDict()
```
webknjaz, line 418: Try depending on `any_multidict_class` instead of importing directly.

webknjaz, line 418: Would something like this work?
```suggestion
def test_create_empty_multidict(
    any_multidict_class: Type[MutableMultiMapping[str]],
    benchmark: BenchmarkFixture,
) -> None:
    benchmark(lambda: any_multidict_class())
```

bdraco: Yeah I think so

bdraco: I'll do a followup in a bit after I finish digging though the latest profiles

webknjaz: Thanks! I figured it'd be useful to expand the tests in this way due to built-in parametrization.

bdraco: https://github.com/aio-libs/multidict/pull/1092

PR #1091, opened 2025-03-20
Add benchmarks for copying an existing CIMultiDict

bdraco:
followup for #1088 for the other cases currently in use in aiohttp

https://github.com/aio-libs/aiohttp/blob/7b79f0c02e91fd172aa0d75a3d8315f0717b5c2d/aiohttp/client.py#L1139 https://github.com/aio-libs/aiohttp/blob/7b79f0c02e91fd172aa0d75a3d8315f0717b5c2d/aiohttp/client.py#L1138

These two blocks above look ripe for optimization


bdraco:
https://github.com/aio-libs/multidict/blob/2de02b5a478a084cf97e47d4a6260212749f1d47/multidict/_multidict.c#L353 seems to hold true today

PR #1092, opened 2025-03-20
Use `case_insensitive_multidict_class` and `any_multidict_class` for benchmarks

bdraco:
https://github.com/aio-libs/multidict/pull/1089#discussion_r2006532223

### DIFF
--- a/CHANGES/1092.contrib
+++ b/CHANGES/1092.contrib
@@ -1,0 +1,1 @@
+Started running benchmarks for the pure Python implementation and the C implementation -- by :user:`bdraco`.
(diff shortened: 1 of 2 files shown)


bdraco:
There are a few more than can be collapsed if this works ok

webknjaz:
Why exclude istr, though? We also have `case_insensitive_str_class` that might be useful here.

bdraco:
> Why exclude istr, though? We also have `case_insensitive_str_class` that might be useful here.

We can use `case_insensitive_multidict_proxy_class` for the `istr` ones

webknjaz:
@bdraco I updated the required Codecov checks. Rebase when you can.

bdraco:
Adding all the combination is making the benchmarks a bit slower.  codspeed supports xdist so we could add that later if it bothers anyone

webknjaz:
Why is it slower, though? You're saying the tests are serial, right?

webknjaz (approved):
Might be worth having a contrib note.

bdraco:
> Why is it slower, though? You're saying the tests are serial, right?

Yes. The increase in quantity of benchmarks running serially

PR #1094, opened 2025-03-21
Run benchmarks with pytest-xdist to reduce run time

bdraco:
https://github.com/aio-libs/multidict/pull/1092#issuecomment-2741904289


bdraco:
The benchmarks are too small to use xdist because the jitter ends up too high on the GH runners

PR #1096, opened 2025-03-21
Implement a part of multidict python helpers in C

asvetlov:


webknjaz (approved):

bdraco:
> Hmm, I didn't expect such performance gain. The C implementation is dumb; it could be optimized, but I don't care much about the performance of views operations except iteration. @bdraco @Dreamsorcerer, if you guys have a different opinion, please share it. I could optimize more, but it takes time, sure.

I don't think its worth the time to optimize views more.

The most common operations are likely:

- fetch via subscript
- set
- `get`, `getall`
- Creating new MultiDict objects

The optimizations you already did to the miss cases for `get` and `getall` are likely to have the most real world impact.

PR #1097, opened 2025-03-22
Use istr as keys in CIMultiDict

asvetlov:
Currently, keys of CIMultiDict are stored as-is, e.g. `CIMultiDict({'k': 'v'})` has `str` key.

But it is slightly incorrect, CIMultiDict should store `istr` keys.

### DIFF
--- a/multidict/_multidict_py.py
+++ b/multidict/_multidict_py.py
@@ -80,8 +80,9 @@ class _Iter(Generic[_T]):
 
 
 class _ViewBase(Generic[_V]):
-    def __init__(self, impl: _Impl[_V]):
+    def __init__(self, impl: _Impl[_V], keyfunc: Callable[[str], str]):
         self._impl = impl
+        self._keyfunc = keyfunc
 
     def __len__(self) -> int:
         return len(self._impl._items)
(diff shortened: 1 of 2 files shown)


asvetlov:
The regression is ok

bdraco:
> The regression is ok

I think it’s fine as it’s a rare case.

will test on production later today

bdraco:
Testing

- [x] aiohttp master
- [ ] aiohttp 3.11
- [ ] HA production 1
- [ ] HA production 2

CHANGES/1097.bugfix
```diff
@@ -0,0 +1,6 @@
+Rewritten :class:`multidict.CIMultiDict` and it proxy to always return
```
bdraco, line 1:
```suggestion
Rewrote :class:`multidict.CIMultiDict` and it proxy to always return
```

bdraco (approved):

bdraco:
Tested on production and as above. All seems to be working as expected.

Looked through the code changes and didn't see any issues stand out

asvetlov:
Thanks for the review, @bdraco

PR #1112, opened 2025-03-31
Fix creating CIMultiDict from MultiDict not making keys case-insentive

### LINKED ISSUE #1111: test_init_headers_MultiDict failing on PyPy
https://github.com/aio-libs/aiohttp/actions/runs/14181568999/job/39728702548?pr=10654

Likely a regression in the pure-python version

bdraco:
When `MultiDict` was passed to `CIMultiDict` the keys need to be converted
fixes #1111

side note `_multidict_extend` doesn't appear to use `name`

### DIFF
--- a/multidict/_multidict.c
+++ b/multidict/_multidict.c
@@ -78,7 +78,7 @@ _multidict_extend(MultiDictObject *self, PyObject *arg,
     }
 
     if (arg != NULL) {
-        if (CIMultiDict_CheckExact(arg) || (!CIMultiDict_CheckExact(self) && MultiDict_CheckExact(arg))) {
+        if (MultiDict_CheckExact(arg) || CIMultiDict_CheckExact(arg)) {
             list = &((MultiDictObject*)arg)->pairs;
             if (pair_list_update_from_pair_list(&self->pairs, used, list) < 0) {
                 goto fail;
--- a/multidict/_multidict.c
+++ b/multidict/_multidict.c
@@ -347,10 +347,20 @@ ret:
 static inline PyObject *
 multidict_repr(MultiDictObject *self)
 {
-    PyObject *name = PyObject_GetAttrString((PyObject*)Py_TYPE(self), "__name__");
-    if (name == NULL)
+    int tmp = Py_ReprEnter((PyObject *)self);
+    if (tmp < 0) {
         return NULL;
+    }
+    if (tmp > 0) {
+        return PyUnicode_FromString("...");
+    }
+    PyObject *name = PyObject_GetAttrString((PyObject *)Py_TYPE(self), "__name__");
+    if (name == NULL) {
+        Py_ReprLeave((PyObject *)self);
+        return NULL;
+    }
     PyObject *ret = pair_list_repr(&self->pairs, name, true, true);
+    Py_ReprLeave((PyObject *)self);
     Py_CLEAR(name);
     return ret;
 }
--- a/multidict/_multidict.c
+++ b/multidict/_multidict.c
@@ -78,7 +78,7 @@ _multidict_extend(MultiDictObject *self, PyObject *arg,
     }
 
     if (arg != NULL) {
-        if (MultiDict_CheckExact(arg) || CIMultiDict_CheckExact(arg)) {
+        if (CIMultiDict_CheckExact(arg) || (!CIMultiDict_CheckExact(self) && MultiDict_CheckExact(arg))) {
             list = &((MultiDictObject*)arg)->pairs;
             if (pair_list_update_from_pair_list(&self->pairs, used, list) < 0) {
                 goto fail;
--- a/multidict/_multidict.c
+++ b/multidict/_multidict.c
@@ -83,7 +83,7 @@ _multidict_extend(MultiDictObject *self, PyObject *arg,
             if (pair_list_update_from_pair_list(&self->pairs, used, list) < 0) {
                 goto fail;
             }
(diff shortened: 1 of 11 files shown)


tests/test_multidict.py
```diff
@@ -1189,3 +1191,45 @@ def test_items_case_insensitive_isdisjoint(
...
+    assert start_as_cimd["key2"] == "value2"
+    converted_to_ci = CIMultiDict(start_as_md)
+    assert converted_to_ci.get("key") == "value1"
+    assert converted_to_ci["key"] == "value1"
+    assert converted_to_ci.get("key2") == "value2"
+    assert converted_to_ci["key2"] == "value2"
+
+
+@pytest.mark.xfail(reason="issue #1111", strict=True)
+def test_pure_python_convert_multidict_to_cimultidict_eq() -> None:
```
Dreamsorcerer, line 1217: Shouldn't these be run on both implementations with a fixture?

bdraco: It doesn't fail on the C version

multidict/_multidict.c
```diff
@@ -78,7 +78,7 @@ _multidict_extend(MultiDictObject *self, PyObject *arg,
     }
 
     if (arg != NULL) {
-        if (MultiDict_CheckExact(arg) || CIMultiDict_CheckExact(arg)) {
+        if (CIMultiDict_CheckExact(arg) || (!CIMultiDict_CheckExact(self) && MultiDict_CheckExact(arg))) {
```
bdraco, line 81: This isn't efficient because it ends up falling through to `PyMapping_Items`

multidict/_multidict_py.py
```diff
@@ -628,7 +631,10 @@ def _extend(
     ) -> None:
         if arg:
             if isinstance(arg, (MultiDict, MultiDictProxy)):
-                items = arg._impl._items
+                if self._ci and not arg._ci:
+                    items = [(self._title(k), k, v) for _, k, v in arg._impl._items]
+                else:
+                    items = arg._impl._items
```
bdraco, line 637: Should this be making a copy?

bdraco:
Feel free to adjust this PR.

I'm not sure this regression is worth yanking 6.3.0 or if this this is even a real use case ?

multidict/_multidict_py.py
bdraco: Let's check https://github.com/aio-libs/multidict/pull/1113

bdraco: Yeah thats a problem

multidict/_multidict_py.py
```diff
@@ -254,7 +254,7 @@ def __xor__(self, other: Iterable[_T]) -> set[Union[tuple[str, _V], _T]]:
         except TypeError:
             return NotImplemented
         ret: set[Union[tuple[str, _V], _T]] = self - rgt
-        ret |= (rgt - self)
+        ret |= rgt - self
```
bdraco, line 257:
```suggestion
        ret |= (rgt - self)
```

multidict/_multidict_py.py
```diff
@@ -425,7 +425,7 @@ def __xor__(self, other: Iterable[_T]) -> set[Union[str, _T]]:
         except TypeError:
             return NotImplemented
         ret: set[Union[str, _T]] = self - rgt  # type: ignore[assignment]
-        ret |= (rgt - self)
+        ret |= rgt - self
```
bdraco, line 428:
```suggestion
        ret |= (rgt - self)
```

tests/test_multidict.py
webknjaz: But it may gain the same bug over time. I think the value of using the parametrized fixtures is in making sure that the behavior is the same across implementations. And this would help making sure that any regression fixes don't contribute to diverging behavior.

multidict/_multidict.c
asvetlov: Please no, I prefer the proper fix.

multidict/_multilib/pair_list.h
```diff
@@ -1132,23 +1132,45 @@ static inline int
...
 {
     Py_ssize_t pos;
+    Py_hash_t hash;
+    PyObject *identity = NULL;
+    bool recalc_identity = list->calc_ci_indentity != other->calc_ci_indentity;
 
     for (pos = 0; pos < other->size; pos++) {
         pair_t *pair = other->pairs + pos;
+        if (recalc_identity) {
+            identity = pair_list_calc_identity(list, pair->key);
```
asvetlov, line 1142: The fix recalculates identity and its hash if the type of self does not equal the other's.

asvetlov (approved):
All right, I think now the PR is good.

multidict/_multidict_py.py
```diff
@@ -628,9 +631,13 @@ def _extend(
     ) -> None:
         if arg:
             if isinstance(arg, (MultiDict, MultiDictProxy)):
-                items = arg._impl._items
+                if self._ci != arg._ci:
```
webknjaz, line 634: it's customary to use identity checks for booleans
```suggestion
                if self._ci is not arg._ci:
```

tests/test_multidict.py
bdraco: It makes sense to change it now that it's a positive test instead of an xfail

multidict/_multidict.c
bdraco: Thanks for finishing it up 👍

tests/test_multidict.py
bdraco: Adjusted to test all versions

PR #1113, opened 2025-04-01
Add tests for creating a MultiDict from an existing one with new pairs

bdraco:


CHANGES/1113.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixed the pure Python mutating the original :class:`multidict.MultiDict` when creating a new :class:`multidict.CIMultiDict` from an existing one when keyword arguments are also passed -- by :user:`bdraco`.
```
bdraco, line 1:
```suggestion
Fixed the pure Python version mutating the original :class:`multidict.MultiDict` when creating a new :class:`multidict.CIMultiDict` from an existing one when keyword arguments are also passed -- by :user:`bdraco`.
```

PR #1116, opened 2025-04-01
Reformat

asvetlov:


webknjaz (changes requested):
Add a contrib change note, then. And a blame ignore update in a follow-up.

webknjaz:
This PR doesn't only contain formatting but a contributor-facing configuration change. And yes, blame ignore should be in a follow up.

Makefile
```diff
@@ -30,7 +30,7 @@ lint: black-check isort-check
 	python -Im pre_commit run --all-files --show-diff-on-failure
 
 fmt:
-	black -t py35 $(SRC)
+	black -t py39 $(SRC)
```
asvetlov:
```suggestion
	black -t py35 $(SRC)
```

asvetlov:
Are you talking about `black --py39 ...` option?
Ok, I've reverted it back -- it doesn't change anything for the project.

bdraco:
might be nice to format the C code with something like https://github.com/ratgdo/esphome-ratgdo/blob/05967a5cb1e27341e55917188c79ca80a610da5f/.pre-commit-config.yaml#L10 as well (no preferences on the settings, other than to not have to think about it)

webknjaz (approved):

PR #1121, opened 2025-04-03
Fix memory leak on objects passed to `init`, `update`, or `extend`

### LINKED ISSUE #1117: Memory leak on 6.3.0, 6.3.1
Since using `[multidict==6.3.0](https://github.com/aio-libs/multidict/releases/tag/v6.3.0)` we are having a memory leak on `yarl._url`.

We are using `aiohttp==3.9.5` and installing deps `yarl==1.18.3` that has `multidict >= 4.0`

Here you can see a memray screenshot

Image

Image

We get rid of the leak downgrading again to `multidict==6.2.0`

bdraco:
fixes #1117

### DIFF
--- a/CHANGES/1121.bugfix.rst
+++ b/CHANGES/1121.bugfix.rst
@@ -1,0 +1,1 @@
+Resolved a memory leak by ensuring proper reference count decrementation -- by :user:`asvetlov` and :user:`bdraco`.
--- a/docs/spelling_wordlist.txt
+++ b/docs/spelling_wordlist.txt
@@ -26,6 +26,7 @@ Ctrl
 cython
 Deprecations
 deallocation
+decrementation
 dev
 dict
 docstrings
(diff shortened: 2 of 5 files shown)


asvetlov:
Good finding.
Let's apply a little different fix.
`_multidict_extend_parse_args()` should incref all returned values; I suggest you don't change it.
But the caller should decref.
This is the standard practice in Python, let's keep it.

asvetlov:
Please take a look at the current PR's state.
It fixes the bug but avoids reference stealing.
Borrowed references are dangerous, in CPython itself we are trying to get rid of them everywhere

bdraco:
Thanks! Thats much better and safer. 🚀 

Will clean it up, add changelog, and get it shipped

bdraco:
If you get a chance, https://github.com/aio-libs/yarl/pull/1456 could use a followup review.

PR #1127, opened 2025-04-04
Start building armv7l wheels

bdraco:
## What do these changes do?

Copy the wheel build changes from `yarl` to use `armv7l` which also splits the `manylinux` and `musllinux` jobs which means the release process will be a bit faster

~~Note that the `manylinux` `armv7l` wheels might still not build and we will than have to add the exclude like we do for `yarl` since the previous `manylinux` image for `armv7l` did not have a working `cffi` package. In theory this has been fixed in https://github.com/pypa/cibuildwheel/releases/tag/v2.23.0~~ yarl released fine with manylinux now

`armv7l` wheels

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -124,9 +124,24 @@ jobs:
         - ubuntu
         - windows
         - macos
+        tag:
+        - ''
+        - 'musllinux'
+        exclude:
+        - os: windows
+          tag: 'musllinux'
+        - os: macos
+          tag: 'musllinux'
+        - os: ubuntu
+          tag: >-
+           ${{
+           (github.event_name != 'push' || !contains(github.ref, 'refs/tags/'))
+           && 'musllinux' || 'none'
+           }}
     uses: ./.github/workflows/reusable-build-wheel.yml
     with:
       os: ${{ matrix.os }}
+      tag: ${{ matrix.tag }}
       wheel-tags-to-skip: >-
         ${{
         (github.event_name != 'push' || !contains(github.ref, 'refs/tags/'))
@@ -135,7 +150,8 @@ jobs:
         *-musllinux_*
         *-win32
         pp*'
-        || ''
+        || (matrix.tag == 'musllinux') && '*-manylinux_* pp*'
+        || '*-musllinux_* pp*'
         }}
       source-tarball-name: >-
         ${{ needs.build-pure-python-dists.outputs.sdist-filename }}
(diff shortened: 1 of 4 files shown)


.github/workflows/reusable-build-wheel.yml
```diff
@@ -75,8 +81,10 @@ jobs:
     - name: Upload built artifacts for testing and publishing
       uses: actions/upload-artifact@v4
       with:
-        name: >-
-          ${{ inputs.dists-artifact-name }}-${{ inputs.os }}-${{ inputs.qemu }}
+        name: ${{ inputs.dists-artifact-name }}-
+           ${{ inputs.os }}-
+           ${{ inputs.qemu }}-
+           ${{ inputs.tag }}
```
webknjaz, line 87: @bdraco I recently discovered that this would break because of asterisks: https://github.com/ansible/pylibssh/pull/676#issuecomment-2656574936.

And so I ended up with a solution where I compute a hash of all the inputs: [ansible/pylibssh@`90e86aa` (#676)](https://github.com/ansible/pylibssh/pull/676/commits/90e86aa39174723effb05bf151360d9448880f84#diff-3826dffee70012c1e07ee3d63a4cf5ba7ed82c17bbac68145432a97e71dbade7R95-R112).
Then, I append it: [ansible/pylibssh@`90e86aa` (#676)](https://github.com/ansible/pylibssh/pull/676/commits/90e86aa39174723effb05bf151360d9448880f84#diff-3826dffee70012c1e07ee3d63a4cf5ba7ed82c17bbac68145432a97e71dbade7R213).

Plz migrate that solution here instead.

.github/workflows/ci-cd.yml
```diff
@@ -124,9 +124,24 @@ jobs:
...
+        - 'musllinux'
+        exclude:
+        - os: windows
+          tag: 'musllinux'
+        - os: macos
+          tag: 'musllinux'
+        - os: ubuntu
+          tag: >-
+           ${{
+           (github.event_name != 'push' || !contains(github.ref, 'refs/tags/'))
```
webknjaz, line 138: There's a more elegant shortcut in GHA nowadays:
```suggestion
           (github.event_name != 'push' || github.ref_type != 'tag')
```

.github/workflows/reusable-build-wheel.yml
bdraco: Thanks. Will adjust when time permits.

bdraco: migrated in 656ee41af45ce42929346360b9aa78d6615a9f42

PR #1128, opened 2025-04-05
Replace pair_list with hash table

asvetlov:


asvetlov:
The PR is more-or-less done but I'd like to make some polishing and self-review later.
Please don't merge it, I'll be only partially available next month, I cannot commit that I'll have enough time for fixing issues when the new version will be released.

Careful testing is appreciated!

New multidict is close to Python's dict except for multiple keys, of course.

It starts from the empty hashtable, which grows by a power of 2 starting from 8: 8, 16, 32, 64, 128, ...
The amount of items is 2/3 of the hashtable size (1/3 of the table is never allocated).

The table is resized if needed, and bulk updates (`extend()`, `update()`, and constructor calls) pre-allocate many items at once, reducing the amount of potential hashtable resizes.

Item deletion puts DKIX_DUMMY special index in the hashtable. In opposite to the standard `dict`, DKIX_DUMMY is never replaced with an index of the new entry except by hashtable indices rebuild. It allows to keep the insertion order for multiple equal keys.

The iteration for operations like `getall()` is a little tricky. The next index calculation could return the already visited index before reaching the end. To eliminate duplicates, the code marks already visited entries by `entry->hash = -1`. `-1` hash is an invalid hash value that could be used as a marker. After the iteration finishes, all marked entries are restored.
Double iteration over the indices still has O(1) amortized time, it is ok.

`.add()`, `val = md[key]`, `md[key] = val`, `md.setdefault()` all have O(1).
`.getall()` / `.popall()` have O(N) where N is the amount of returned items.
`.update()` / `extend()` have O(N+M) where N and M are amount of items in the left and right arguments (we had quadratic time here before). 

`.copy()` is super fast; construction multidict from another multidict could be optimized to reuse `.copy()` approach.

The performance is okay, multidict creation is slightly slower because the hashtable should be recalculated and the indices table rebuilds, but all other operations are faster.

Multidict is still slightly slower than the regular dict because I don't want to use the private API for accessing internal Python structures, the most notable is the string's hash.

TODO: 
1. Optimize `MultiDict(md)` construction.
2. GC untrack the multidict if all keys/values are not tracked. In aiohttp, values are usually untracked `str` instanced.

Open question: Should we use `HT_PERTURB_SHIFT` in the index calculation? It is crucial for storing integers where `hash(num) == num`, but I doubt if it decreases the number of collisions for `str` keys. During the work on PR I saw many times when `idx = next(idx)` didn't change the current `idx` or it was like `1, 2, 1, 3` with a relatively high chance of duplication. 
It may not be worth changing. I have no idea how to prove this except by calculating the collision statistics on a large enough number of keys.

If anybody wants to play with the code, two things could help debugging.
1. Uncomment `# CFLAGS = ["-O0", "-g3", "-UNDEBUG"]` line in `setup.py` to make a debugging build. It enables asserts and `ASSERT_CONSISTENT(...)` self-consistency check.
2. `_ht_dump(...)` function prints the current hashtable structure in a form useful for analyzing the internal structure.

Please feel free to experiment with and ask any questions.

multidict/_multidict_py.py
```diff
@@ -455,15 +454,204 @@ def _identity(self, key: str) -> str:
...
+    def new(cls, log2_size: int, entries: list[Optional[_Entry[_V]]]) -> Self:
+        size = 1 << log2_size
+        usable = (size << 1) // 3
+        if log2_size < 10:
+            indices = cls.PREALLOCATED_INDICES[log2_size].__copy__()
+        elif log2_size < 16:
+            indices = array("h", (-1 for i in range(size)))
+        elif log2_size < 32:
+            indices = array("l", (-1 for i in range(size)))
+        else:  # pragma: no cover  # don't test huge multidicts
```
webknjaz, line 514: Why not?

Vizonex:
https://app.codecov.io/gh/aio-libs/multidict/pull/1128/blob/multidict/_multidict_py.py#L819
@asvetlov I wrote a test that passes through line 819 I wonder this was the last thing needed for code-coverage other than that the think I can think of that you didn't was is adding a summary of the changes made in the changes folder 
You'll have to put this down into `test_update.py` since I tried deleting mine and then attempted forking this one to make changes which didn't work.
```python
def test_update_with_second_md(any_multidict_class: _MD_Classes) -> None:
    obj1 = any_multidict_class()
    obj2 = any_multidict_class([("a", 2)])
    obj1.update(obj2)
    assert obj1 == obj2
```

multidict/_multidict_py.py
asvetlov: Because of memory: a multidict with 2^32 items takes more than 128 GiB, it is much more than the memory size of the test box.

asvetlov:
@Vizonex thanks!

multidict/_multidict_py.py
Dreamsorcerer: Is it cheatable? e.g. If we used a log2_size==32, but actually only passed 1 or 2 items. Or is the empty allocated array already too big?

multidict/_multidict_py.py
```diff
@@ -455,15 +454,204 @@ def _identity(self, key: str) -> str:
...
+        # lst = [(i, ix, perturb)]
+        while ix != -1:
+            perturb >>= 5
+            i = (i * 5 + perturb + 1) & mask
+            ix = indices[i]
+            # lst += [(i, ix, perturb)]
+        # if len(lst) > 7:
+        #     print("find_empty_slot", hash_, mask)
+        #     for line in lst:
+        #         print('  ', *line)
```
Dreamsorcerer: Looks like something needs cleaning up.

Dreamsorcerer:
I'm not too familiar with this, so I'll leave it with the others. Certainly looking forward to the performance improvement though.

multidict/_multilib/hashtable.h
```diff
@@ -0,0 +1,1930 @@
...
+fail:
+    Py_XDECREF(identity);
+    return -1;
+}
+
+
+static inline int
+md_get_all(MultiDictObject *md, PyObject *key, PyObject **ret)
+{
+    *ret == NULL;
```
bdraco: Assignment was likely intended here

multidict/_multilib/hashtable.h
```diff
@@ -0,0 +1,1930 @@
...
+            || pos2 >= other->keys->nentries) {
+            return 1;
+        }
+        entry_t *entry1 = lft_entries + pos1;
+        if (entry1->identity == NULL) {
+            pos1++;
+            continue;
+        }
+        entry_t *entry2 = rht_entries +pos2;
+        if (entry1->identity == NULL) {
```
bdraco:
```suggestion
        if (entry2->identity == NULL) {
```

multidict/_multilib/hashtable.h
bdraco:
```suggestion
    *ret = NULL;
```

multidict/_multilib/hashtable.h
```diff
@@ -0,0 +1,1930 @@
...
+        ret = -1;
+        PyErr_SetString(PyExc_RuntimeError,
+                        "MultiDict is changed during iteration");
+        goto cleanup;
+    }
+
+    entry_t *entries = htkeys_entries(finder->md->keys);
+    if (!finder->first) {
+        htkeysiter_next(&finder->iter);
+        finder->first = false;
```
bdraco: something seems off here

multidict/_multilib/hashtable.h
```diff
@@ -0,0 +1,1930 @@
...
+            if (_md_update(md, hash, identity, key, value) < 0) {
+                goto fail;
+            }
+            Py_CLEAR(identity);
+            Py_CLEAR(key);
+        } else {
+            int tmp = _md_add_with_hash_steal_refs(md, hash, identity,
+                                                   key, Py_NewRef(value));
+            if (tmp < 0) {
+                Py_DECREF(value);
```
bdraco, line 1381: TODO: - note to self - need to double check this is safe and steal_refs didn't already take ownership

multidict/_multilib/hashtable.h
```diff
@@ -0,0 +1,1930 @@
...
+    finder->version = md->version;
+    finder->md = md;
+    finder->identity = identity;
+    finder->hash = _unicode_hash(identity);
+    if (finder->hash == -1) {
+        return -1;
+    }
+    htkeysiter_init(&finder->iter, finder->md->keys, finder->hash);
+    finder->first = true;
+    return -0;
```
bdraco: Why -0?

multidict/_multilib/hashtable.h
```diff
@@ -0,0 +1,1930 @@
...
+        goto fail;
+    }
+
+    if (md_init_finder(md, identity, &finder) < 0) {
+        assert(PyErr_Occurred());
+        goto fail;
+    }
+
+    int tmp;
+    PyObject *value = NULL;;
```
bdraco:
```suggestion
    PyObject *value = NULL;
```

multidict/_multilib/hashtable.h
```diff
@@ -0,0 +1,1930 @@
...
+    Py_CLEAR(key);
+    if (ret == NULL) {
+        return NULL;
+    }
+
+    htkeysiter_t iter;
+    htkeysiter_init(&iter, md->keys, entry->hash);
+
+    for(; iter.index != pos; htkeysiter_next(&iter)) {
+    }
```
bdraco, line 1123: Is there a guarantee this will visit all slots and not loop forever?

multidict/_multilib/hashtable.h
```diff
@@ -0,0 +1,1930 @@
...
+    while (entry->identity == NULL) {
+        pos->pos += 1;
+        if (pos->pos >= md->keys->nentries) {
+            goto cleanup;
+        }
+        entry += 1;
+    }
+
+    if (pidentity) {
+        *pidentity = Py_NewRef(entry->identity);;
```
bdraco:
```suggestion
        *pidentity = Py_NewRef(entry->identity);
```

multidict/_multilib/hashtable.h
```diff
@@ -0,0 +1,1930 @@
...
+        }
+        entry_t *entry = entries + iter.index;
+        if (hash != entry->hash) {
+            continue;
+        }
+        int tmp = _str_cmp(entry->identity, identity);
+        if (tmp < 0) {
+            goto fail;
+        }
+        if (tmp == 0) {
```
bdraco, line 554: this was confusing but looks correct

multidict/_multilib/hashtable.h
bdraco:
```suggestion
    } else {
        finder->first = false;
```

I think this was the intent ?

multidict/_multilib/hashtable.h
asvetlov: Where were my eyes!

multidict/_multilib/hashtable.h
asvetlov: `_md_add_with_hash_steal_refs` can fail only if hashtable resize fails, in this case the function doesn't steal the reference to value

multidict/_multilib/hashtable.h
asvetlov: A typo. I'll fix it, thanks.

asvetlov: done

multidict/_multilib/hashtable.h
asvetlov: Let me please refresh my memory; maybe the `first` flag is not required at all.

I recall having issues with infinite hashtable lookup, but it was properly fixed in another place.

bdraco: https://github.com/aio-libs/multidict/pull/1128#discussion_r2138153887

multidict/_multilib/hashtable.h
asvetlov: We already know the index of the last element.
We should find a slot in hashtable that points to the last index.
The slot is definitely in the HT because there is no way to add HT entry without grabbing a free slot for it; `_md_add_with_hash_steal_refs` does it atomically.
Strictly speaking, the operation is atomic only if the GIL is held. Free threading could cause a crash in multidict if the same object is modified by different threads concurrently. This bug, related to free-threading, exists in all published versions as well. I want to fix it eventually, but the proper free threading support is another story.

Returning to the question, we have a slot that points to the last element by definition. We know the hash of the last stored key. If we iterate over all slots using this hash, we will eventually reach the required index. The index is present in the iterated sequence by definition.

Very often it requires a single lookup, but in case of hash collisions or duplicated keys, multiple trials are needed.

multidict/_multidict_py.py
asvetlov: Yep, thanks!
This is the artifact of my debug sessions, hash-table is a little harder structure than plain list

multidict/_multidict_py.py
asvetlov: For slots, it is only 32GB for the index table itself. 
Still too big for my taste.
I have no idea how to cheat this function without significant performance loss or compromising readability.

multidict/_multilib/hashtable.h
asvetlov: Yes, the `first` flag is not required at all.

multidict/_multidict_py.py
Dreamsorcerer: I meant cheat the test, not change the function. But, sounds like it's not going to work.

Vizonex:
How is code coverage not hitting the same line I thought I had covered? I'm gonna have to re-patch it. This is confusing 😕

asvetlov:
Now the test coverage is better.
The only missed things are checks like `if identity == entry.identity: ...`
They are essential in case of hash collisions: two different strings have the same hash.
Unfortunately, I have no idea how to generate this pair of strings, especially for 8-byte hash and the python random hash seed.
I suggest marking these lines as `# pragma: no branch` and that's it.

Vizonex:
I love the changes being done to multidict, anything I can do this time to keep things moving forward?

bdraco:
yarl performance changes: https://codspeed.io/aio-libs/yarl/branches/dependabot%2Fpip%2Fmultidict-6.5.0

Vizonex:
> I triggered a manual dependabot run:
> 
> yarl performance changes: https://codspeed.io/aio-libs/yarl/branches/dependabot%2Fpip%2Fmultidict-6.5.0 aiohttp performance changes: https://codspeed.io/aio-libs/aiohttp/branches/dependabot%2Fpip%2F3.13%2Fmultidict-6.5.0

That's amazing I hope adding multidict to cython will further enhance aiohttp it as well

PR #1135, opened 2025-04-09
Fix ref leak when calling update

### LINKED ISSUE #1134: Memory leak updating CIMultiDict from mapping (present in far back as 5.2.0, likely introduced in 4.4.0)
> # Leak 2 (old leak)
> 
> This version seems to leak as far back as 5.2.0, maybe earlier
> 
> leak happens when updating a `CIMultiDict` from a `dict` or `CIMultiDict`
> 
> EDIT: I have not found a version that doesn't leak that I can still compile
> 
> ```python
> import os
> import gc
> import psutil
> import ctypes
> from multidict import CIMultiDict
> 
> def trim_memory() -> int:
>     gc.collect()
>     # libc = ctypes.CDLL("libc.so.6")
>     # return libc.malloc_trim(0)
> 
> 
> def get_memory_usage():
>     process = psutil.Process(os.getpid())
>     memory_info = process.memory_info()
>     return memory_info.rss / (1024 * 1024)
> 
> 
> def main():
>     for _ in range(10_000):
>         for _ in range(1000):
>             result = CIMultiDict()
>             headers = {f"X-Any-{i}": str(i) for i in range(1000)}
>             result.update(headers)
>         del result
>         del headers
>         trim_memory()
>         print(f"Memory usage: {get_memory_usage():.2f} MB")
> 
> 
> if __name__ == "__main__":
>     main()
> ```
> 
> ```
> % python3.9 leak2.py                                 
> multidict version: 6.3.2
> Memory usage: 37.47 MB
> Memory usage: 61.06 MB
> Memory usage: 84.12 MB
> ```
> 
> ```
> % python3.9 leak2.py                                 
> multidict version: 6.0.0
> Memory usage: 36.91 MB
> Memory usage: 60.98 MB
> Memory usage: 84.89 MB
> Memory usage: 108.28 MB
> Memory usage: 131.38 MB
> ``` 

 _Originally posted by @bdraco in [#1131](https://github.com/aio-libs/multidict/issues/1131#issuecomment-2787532499)_

bdraco:
fixes #1134

### DIFF
--- a/multidict/_multilib/pair_list.h
+++ b/multidict/_multilib/pair_list.h
@@ -1015,6 +1015,7 @@ _dict_set_number(PyObject *dict, PyObject *key, Py_ssize_t num)
         return -1;
     }
 
+    Py_DECREF(tmp);
     return 0;
 }
 


asvetlov (approved):

multidict/_multilib/pair_list.h
```diff
@@ -1015,6 +1015,7 @@ _dict_set_number(PyObject *dict, PyObject *key, Py_ssize_t num)
         return -1;
     }
 
+    Py_DECREF(tmp);
```
asvetlov, line 1018: I have no idea why I had thought that `PyDict_SetItem` steals a reference.

PR #1145, opened 2025-04-10
Run tests in debug mode in an additional CI job

bdraco:
### ✅ What This Does

- Adds **additional CI runners** to catch issues like [#1143](https://github.com/aio-libs/multidict/issues/1143).
- These new runners are designed to **supplement** the existing runners:
  - They do **not** track coverage.
  - They are intended for **quick validation**:
    - Ensure a debug build compiles.
    - Ensure the test suite passes.
    - Keep CI runtime impact minimal.

---

### 🔧 New Environment Variable

The `MULTIDICT_DEBUG_BUILD` environment variable is now available to enable a debug build.

PR #1146, opened 2025-04-10
Run tests in debug mode as well

bdraco:
Run tests in debug mode in an additional CI job (alternative)

alternative to https://github.com/aio-libs/multidict/pull/1145 that uses
the existing matrix


.github/workflows/ci-cd.yml
```diff
@@ -324,18 +344,25 @@ jobs:
...
+      if: >-
+        ${{ matrix.debug == 'Y' }}
+      run: >-
+        make clean
+      shell: bash
     - name: >-
         Self-install (from ${{
-          endsWith(matrix.pyver, '-dev')
+          (endsWith(matrix.pyver, '-dev') || matrix.debug == 'Y')
           && 'source'
```
webknjaz, line 356: I think we can just treat source as debug. Since these aren't compiled. Compiling them in non-debug mode is already exercised in the manylinux jobs. So we can just stick debug into the env vars here and the wheel variants would go through the `-O3` code paths. To achieve this, I'd maybe remove the wheels post download conditionally and the lookup step would only find sdists in such jobs.

asvetlov (commented):
Yeah, I like this variant more than #1145

README.rst
```diff
@@ -104,7 +104,12 @@ e.g.:
 Please note, the pure Python (uncompiled) version is about 20-50 times slower depending on
 the usage scenario!!!
 
+For extension development, set the ``MULTIDICT_DEBUG_BUILD`` environment variable to compile
+the extensions in debug mode:
 
+.. code-block:: bash
```
bdraco, line 110:
```suggestion
.. code-block:: console
```

https://github.com/aio-libs/multidict/pull/1145#discussion_r2037110459

docs/multidict.rst
```diff
@@ -442,3 +442,15 @@ Environment variables
 
       The pure-Python (uncompiled) version is roughly 20-50 times slower than
       its C counterpart, depending on the way it's used.
+
+.. envvar:: MULTIDICT_DEBUG_BUILD
+   
+   An environment variable that instructs the packaging scripts to compile
+   the C-extension based variant of :mod:`multidict` with debug symbols.
+   This is useful for debugging the C-extension code, but it will result in
+   a larger binary size and slower performance.
```
bdraco, line 451:
```suggestion
   a larger binary size and worse performance.
```

.github/workflows/ci-cd.yml
bdraco: Left the sdist in because it is needed to hash the requirements file.  It could be refactored more later in a future PR if desired.

PR #1151, opened 2025-04-12
Increment version to 6.4.4.dev0

bdraco:


asvetlov:
@bdraco perhaps you should run `git pull --prune` to not upload stale branches again.
If it was not you -- sorry, I don't know how to get the uploader nickname from the git history.

bdraco:
I don’t recall doing an unbound push, but if I did, apologies for that!

PR #1157, opened 2025-04-27
Shorten fixture names

asvetlov:
Long names don't fit the benchmarks UI; short versions are better and still readable.


webknjaz (changes requested):
Fixture names and identifiers are long for the sake of readability. Short abbreviations are more difficult to decode.

asvetlov:
The programming world is full of short names and abbreviations.
We all live with it without any problem.
If the fixture name takes half of the allowed line width -- it is a problem from my perspective.
Python has a long tradition for using shorteners: `len()` v.s. `length()`, `__init__()` v.s. `__initializer__()`, `dict` v.s. `dictionary`. 

All these short names are pretty readable and maintainable.

Moreover, I personally found that `case_sensitive` / `case_insensitive` pair gives me a higher chance of making a mistake than `ci` / `cs`.

asvetlov:
@bdraco @Dreamsorcerer guys, please share your opinion on the raised question, do you prefer short abbreviated names in tests or as-long-as-possible ones?

bdraco (commented):
I’ve definitely experienced this problem before, where the data becomes really difficult to read on smaller screens because of its length. I completely understand that everyone has different perspectives on the best way to handle it, and it’s unlikely we’ll get full alignment across the board.

Personally, given that I have limited time both on desktop and overall, I tend to favor changes that make it easier to contribute from mobile devices. From that perspective, I’m supportive of this change since it helps improve accessibility.

Dreamsorcerer:
I try to aim to be concise, but unambiguous. It's easy to go too far in either direction. The changes in this PR don't seem to make a significant difference to me, so I don't mind either way (I might err slightly to leaving the current state, just to avoid git blame noise).

webknjaz:
I agree with Nick's assessment of difficulties present on devices with insufficient screen estate. That's one of the reasons I tend to insist on short lines.

I also often use shorter identifiers in various UI labels (like GHA).

Shortcuts are sometimes acceptable. Like in cases of commonly known terms/abbreviations (MD is a medical doctor, right? Everybody knows that!).
Weird shortened words are fine in tightly scoped contexts (like in a function), where they can be deducted easily.

In this case, however, other contexts should be taken into account too.

The fixtures already contain one abbreviation (CI for continuous integration, huh? 😉). And that is already somewhat questionable. But at least it matches the actual class names. Adding a chain of abbreviations is definitely not readable. Sure, it might degrade some Æstetics a little bit.

But what about other places where readability counts? We cnt jst strt drppn lttrs bcs o' 1 sngl tbl on gh.

I think this would penalize maintainability long-term at the cost of a short-term satisfaction.

asvetlov:
Maybe short fixture names and comprehensive docstrings could provide a good enough compromise?
I'll add docstrings to the PR soon; after Spanish lockout my spare time is very limited :)

webknjaz:
The docstrings should always be verbose so that `pytest --collect-only --fixtures` is useful.

But that doesn't really fix the problem of a series of abbreviations being used in disconnected contexts because it adds mental overhead affecting when consuming the thing. Code is read much more often than it's being written, and for that reason it's important to optimize for the reading experience. The readers aren't just 1–2 people in the inner circle, and I was optimizing for inclusivity rather than creating entry barriers for potential contributors.

I'm in favor of making names shorter *where this doesn't compromise readability*. `py` and `ext` are generally fine. But dropping any context is a no-go because of mental overhead. When I see `cs`, I know that it's *obviously* a locale code fore Czech. I might've thought of something like Counter Strike if it was upper case. But any other meanings would never be my first guess, nor second or third even.

asvetlov:
We have an agreement for `py` and `ext`, it's excellent!
Regarding `cs` -- we can drop it entirely.
The public API has an implicit assumption that the absent prefix is for case-sensitivity (`MultiDict`) and that case-insensitive things are coded with a `ci` prefix: `CIMultiDict` and `CIMultiDictProxy`.
`ci` is backed in *our* domain already, it is a part of the library's public API and should not cause ambiguity in the library's context.

webknjaz:
`ci` is only fine when followed by`multidict` in source code. But pytest test IDs target different contexts and shouldn't use ambiguous contractions.

asvetlov:
@webknjaz are you kidding me?
What is the different context?
Multidict has literally 2,5 classes and one function, that's it. 
Where is the space for different mental contexts? Where is the space for ambiguity? The domain is so tiny that it can fit in any brain easily.
But long names don't fit on the screen, especially for mobile devices.
This is the point, nothing else.

We have a total of 58 contributors over the past 10 years, including the project admins. 95% of contributors made fewer than 10 commits throughout the entire history. I don't think anybody will suffer after the change.

PR #1167, opened 2025-05-21
Add Windows arm64 builds

### LINKED ISSUE #1166: Add support for Windows arm64 wheels
Now that github actions has added [windows arm64 runner support](https://github.blog/changelog/2025-04-14-windows-arm64-hosted-runners-now-available-in-public-preview/) for public repos (`windows-11-arm`) would there be any appetite for building windows arm64 wheels?

finnagin:
## What do these changes do?

This would add builds for and run tests on arm64 windows wheels

No

## Related issue number

Fixes #1166 

## Checklist

- [x] I think the code is well written (I hope so!)

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -6,6 +6,7 @@ on:
   merge_group:
   push:
     branches:
+    - arm64-runner
     - master
     - >-
       [0-9].[0-9]+
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -6,7 +6,6 @@ on:
   merge_group:
   push:
     branches:
-    - arm64-runner
     - master
     - >-
       [0-9].[0-9]+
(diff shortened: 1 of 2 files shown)


finnagin:
The earliest python version that actions/setup-python supports for windows arm64 is 3.11 so I skipped the arm64 wheel tests for 3.9 & 3.10 but CIBuildWheel does build wheels for those versions and they pass the tests that it runs. Does that seem reasonable to skip those tests?

asvetlov (changes requested):
@finnagin thanks for the PR!
How is ARM Windows different from x64?
From my understanding, API should be the same, at least for a subset af calls used by multidict.
Does it make sense to drop `windows-11-arm` from test job matrix but leave it in build-wheels job?
It doesn't waste additional CI time but provides the same level of test coverage since tests run as a part of the wheel building.

finnagin:
> Does it make sense to drop `windows-11-arm` from test job matrix but leave it in build-wheels job? 
> It doesn't waste additional CI time but provides the same level of test coverage since tests run as a part of the wheel building.

Sure! That makes sense to me. I can do that.

asvetlov (approved):

PR #1168, opened 2025-05-29
Fix failed spelling task

asvetlov:


.github/workflows/reusable-linters.yml
```diff
@@ -74,6 +74,7 @@ jobs:
         fail_ci_if_error: true
     - name: Install spell checker
       run: |
+        sudo apt update
         sudo apt install libenchant-2-dev
```
webknjaz, line 78:
```suggestion
```

asvetlov: Why?

webknjaz: It's an unnecessary artifact from the Travis CI times. It's preinstalled and just wastes CI minutes for no good reason. It's not present in any of my projects, and we've been removing it where there's leftovers lately.

Issue #1170, opened 2025-06-07
Giving multidict Cython Or PyCapsule Support?


Vizonex:
I feel very tempted to ask if I could make it so that Cython could actually support multidict rather than just simply importing it since I feel that performance may steadily improve if this approach was implemented. This way instead of calling `import multidict` we could do something like `cimport multidict` I would assume that doing so would allow the following things in aiohttp to run a bit faster.

- The `_http_writer.pyx` Cython Code could have direct access to istr as a physical CPython Type allowing this check to be a little bit smoother.
- As for `_http_parser.pyx` this could be pretty significant as many of the `multidict` calls could be replaced for C-API Calls. 

## Approach For Compiling
- As for compiling I'm looking to try and approach this the same way numpy would where a call like `_import_array()` could be implemented to import the Capsule.  
-  There could be an extra module that is specifically responsible for finding the header files at compile-time so that users who want to use multidict within cython can easily compile their code alongside with it.

If anybody is confused about what a capsule is I have a [link to it's documentation](https://docs.python.org/3.9/extending/extending.html#using-capsules)

PR #1171, opened 2025-06-08
Implement C-API Capsule For Future CPython & Cython Support

Vizonex:
## What do these changes do?

This Pull Request Features a brand new C-API CPython Capsule that will allow for the implementation of enabling `multidict` to work with Cython a bit more fluently but also allow CPython Developers to intract with `multidict` as well.
I have plans to make it so that `multidict` can have a mini distutils-like setup for finding the header files it needs for compiling with other CPython & Cython extensions similar to what __numpy__ does except we don't need to link any libraries.

This was mainly an idea and after being given the ok that someone would look into it that I was excited enough to spend 3 hours writing it and making sure it was as readable as possible. I have a bit of experience in the field of writing Cython and CPython extensions and doing this felt like a treat to me.

I was up a little bit later than usual so if I forgot something feel free to let me know. 
I'm not planning to push everything right away and I would like to takes these features 
slowly one step at a time to ensure I didn't miss anything. 

A Later Pull request plans to include cython bindings and a mini distutils tool like I mentioned earlier.

On the development side this doesn't change much except adds in an experimental Capsule.
## Related issue number

- #1170 

## Checklist

### DIFF
--- a/multidict/_multilib/capsule.h
+++ b/multidict/_multilib/capsule.h
@@ -1,0 +1,302 @@
+#ifndef __MULTIDICT_CAPSULE_H__
+#define __MULTIDICT_CAPSULE_H__
+
+#include "Python.h"
+#include "dict.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+
+// This C-API Provides Comptability to CPython & Cython.
+
+typedef struct  _multidict_capi_s {
+    PyTypeObject *_IStrType;
+
+    PyTypeObject *_MultiDictType;
+    PyTypeObject *_CIMultiDictType;
+    PyTypeObject *_MultiDictProxyType;
+    PyTypeObject *_CIMultiDictProxyType;
+
+    // multidict_copy 
+    PyObject* (*_MultiDict_Copy)(MultiDictObject* self);
+
+    // multidict_items
+    PyObject* (*_MultiDict_Items)(MultiDictObject* self);
+
+    // multidict_tp_iter 
+    PyObject* (*_MultiDict_Iter)(MultiDictObject* self);
+    
+    // multidict_keys
+    PyObject* (*_MultiDict_Keys)(MultiDictObject* self);
+
+    // multidict_values
+    PyObject* (*_MultiDict_Values)(MultiDictObject* self);
+
+    // multidict_reduce
+    PyObject* (*_MultiDict_Reduce)(MultiDictObject* self);
+
+    // multidict_repr
+    PyObject* (*_MultiDict_Repr)(MultiDictObject* self);
+
+    // multidict_update
+    PyObject* (*_MultiDict_Update)(MultiDictObject* self, PyObject* args, PyObject* kwargs);
+
+    // multidict_proxy_copy(MultiDictProxyObject *self)
+    PyObject* (*_MultiDictProxy_Copy)(MultiDictProxyObject* self);
+    
+} PyMultiDict_CAPI;
+
+
+static PyMultiDict_CAPI* MultiDict_CAPI;
+
+
+// https://docs.python.org/3.9/extending/extending.html#using-capsules
+
+// Most inline functions have a different calling style and need a different approch.
+
+/************************ istr *************************/
+
+#define PyIStrType MultiDict_CAPI->_IStrType
+
+#define PyIStr_New(args) PyIStrType->tp_new(PyIStrType, args, NULL)
+
+// NOTE: IStr_CheckExact(state, obj) , IStr_Check(state, obj) are already defined for us...
+
+
+
+#define PyMultiDictType MultiDict_CAPI->_MultiDictType
+
+
+
+/********************* MultiDict *********************/
+
+
+#define MultiDict_Len(self) \
+    pair_list_len(&self->pairs)
+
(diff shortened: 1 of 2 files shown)


bdraco:
Thanks so much for this PR, it's great to see you contributing!

We used to have quite a bit of Python overhead in the header parser due to jumping back into the Python API to add headers to `multidict`, but we resolved that by building a list in Cython and passing the result to `multidict` at the end:

- https://github.com/aio-libs/aiohttp/blob/8795da3db71b866bd034b552f5410596d49309b9/aiohttp/_http_parser.pyx#L393  
- https://github.com/aio-libs/aiohttp/blob/8795da3db71b866bd034b552f5410596d49309b9/aiohttp/_http_parser.pyx#L421

Your change would avoid that extra copy, which is nice, but unfortunately that's the only spot in `aiohttp` I can think of where it would help. That makes it hard to justify the long-term maintenance cost of exposing and stabilizing a new ABI in `multidict`.

The timing is also a bit tricky. https://github.com/aio-libs/multidict/pull/1128 is in the works, aiming to change lookups from O(n) to O(1). If we were to consider introducing a stable ABI, it would have to wait until that lands and proves itself performance-wise.

That said, we really appreciate you digging into performance and submitting this. If you're interested in more optimization opportunities, we’d love to chat. Feel free to drop by [#aio-libs on Matrix](https://matrix.to/#/#aio-libs:matrix.org). There are still a few hotspots on [codspeed](https://codspeed.io/aio-libs) worth exploring.

Even if this PR doesn’t move forward, please don’t let that discourage you. We’d love to see more contributions from you in the future!

multidict/_multilib/capsule.h
```diff
@@ -0,0 +1,302 @@
...
+    PyObject* (*_MultiDict_Keys)(MultiDictObject* self);
+
+    // multidict_values
+    PyObject* (*_MultiDict_Values)(MultiDictObject* self);
+
+    // multidict_reduce
+    PyObject* (*_MultiDict_Reduce)(MultiDictObject* self);
+
+    // multidict_repr
+    PyObject* (*_MultiDict_Repr)(MultiDictObject* self);
```
asvetlov, line 41: Should we expose `tp_repr` slot?
I doubt if it is a hot path in any code; `PyObject_Repr()` should call the slot in a performant enough way.

Also, I'm not sure about `__reduce__`, but I think that `.extend()` is worth being listed here.

multidict/_multilib/capsule.h
```diff
@@ -0,0 +1,302 @@
...
+    // multidict_update
+    PyObject* (*_MultiDict_Update)(MultiDictObject* self, PyObject* args, PyObject* kwargs);
+
+    // multidict_proxy_copy(MultiDictProxyObject *self)
+    PyObject* (*_MultiDictProxy_Copy)(MultiDictProxyObject* self);
+    
+} PyMultiDict_CAPI;
+
+
+static PyMultiDict_CAPI* MultiDict_CAPI;
```
asvetlov, line 52: I have a bad feeling about mixing two-phase init, `mod_state`, and this static variable.

asvetlov, line 52: I have a bad feeling about mixing two-phase init, `mod_state`, and this static variable.

multidict/_multilib/capsule.h
```diff
@@ -0,0 +1,302 @@
...
+
+
+#define PyMultiDictType MultiDict_CAPI->_MultiDictType
+
+
+
+/********************* MultiDict *********************/
+
+
+#define MultiDict_Len(self) \
```
asvetlov, line 76: Is it the beginning of a public C API?
If yes, it shouldn't use macros with direct access to the multidict internals.
Say, after merging hashtable-based solution this api will be broken if a third-party compiles exactly this version but imports HT-based multidict library version.

Did I miss something?

asvetlov (commented):
Thanks for the PR.
It opens an exciting field of improvements; I love to have something like this eventually.

multidict/_multilib/capsule.h
Vizonex: Now that I think about it PyObject_Repr() could be a better approach to this.

multidict/_multilib/capsule.h
Vizonex: Currently I started waiting for that after being told about the hash-table solution so I put this on hold, however when it's done I plan to start redoing stuff shouldn't take me that long as I found writing this all to be a lot quicker than I expected.

Vizonex: @asvetlov When the new hash-table is finished I have a few questions that could help me with revising this more
- Should I look into only writing C-API Capsules for all functions that require functions that are only written in _multidict.c or should I try hooking all the major components that do not have Vector-Call Styled approaches?
- Should I start looking at this more from a CPython API Approach for the Dictionaries? 
- Would Looking at python's `datetime.h` be a good reference for me to use or do you know of any better approaches for it?
- Should the Capsule just carry the `mod_state` as opposed to the Capsule having to carry all of these types itself?

Vizonex: @asvetlov I also have an idea since I have a lot of time on my hands, I'm gonna close this pull request, help you finish the hash-table end and then after that I can restart this again.

PR #1174, opened 2025-06-11
Drop mypy coverage

### LINKED ISSUE #1122: Drop typing coverage
1. It raises a lot of false positives that cannot be suppressed. Basically, there is no way to make 100% type coverage by definition.
2. It misleads, a reviewer should open the report and explicitly filter by specific flag to get the 'real' coverage numbers.

Typing coverage is a curious feature, but unfortunately, it doesn't work as expected, let's just drop it.

asvetlov:
Closes #1122

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -2,7 +2,7 @@
 
 codecov:
   notify:
-    after_n_builds: 2  # The number of test matrix+lint jobs uploading coverage
+    after_n_builds: 1  # The only test-summary job uploads the coverage
     wait_for_ci: false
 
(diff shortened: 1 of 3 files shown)


webknjaz (changes requested):
I still think that recording type preciseness is useful. That hasn't changed 🤷‍♂️

asvetlov:
@webknjaz is it useful for what? 
Please elaborate.
Also, please answer the questions asked in #1122

bdraco:
This was switched to coveralls to solve the red ci issue

PR #1179, opened 2025-06-19
Setup pre-commit hook for auto-formating both Python and C code

### LINKED ISSUE #1123: Setup pre-commit hook for auto-formating both Python and C parts
The Python part could be the standard ruff.
Regarding C code, I personally prefer  CPython formatting style.
It is not very formal, but I think we can setup https://zed0.co.uk/clang-format-configurator/ to get a config which is at least very close to CPython style.

The only challenge is figuring out the meaning of these dozens of options.
              
              might be nice to format the C code with something like https://github.com/ratgdo/esphome-ratgdo/blob/05967a5cb1e27341e55917188c79ca80a610da5f/.pre-commit-config.yaml#L10 as well (no preferences on the settings, other than to not have to think about it)

_Originally posted by @bdraco in https://github.com/aio-libs/multidict/issues/1116#issuecomment-2770124116_

asvetlov:
Closes #1123

### DIFF
--- a/CHANGES/1123.contrib
+++ b/CHANGES/1123.contrib
@@ -1,1 +1,1 @@
-Setup pre-commit hook for auto-formating both Python and C code.
+Setup pre-commit hook for auto-formatting both Python and C code.


asvetlov:
Benchmarks for pure python tests are not stable after merging hashtable-based impl but this is another issue.

webknjaz:
Urgh.. Too bad that it's now not possible to put a commit into blame ignore.

PR #1181, opened 2025-06-19
Update .git-blame-ignore-revs

asvetlov:


webknjaz:
This is a bad idea as it hides configuration changes. Only automatic formatting is acceptable to ignore like this.

PR #1185, opened 2025-06-21
New MultiDict C-API functions And Added Documentation

Vizonex:
## What do these changes do?

First of all, thank you for giving something better to work with I really appreciate it. 
This Adds a few new api calls mainly belonging to Multidict's api I had a proposed idea for an iterator but I think I might delete it and call it a day. I wasn't lazy at all about it in fact it encouraged me to keep going. I was pretty creative about exposing some functions such as MultiDicts updating each other or using Sequence and Mapping so that they could be utilized as shortcuts if needed. 

Only the Good stuff is coming, I even wrote doxygen comments into each functions so that users would have a much better time using the C-API. Writing the pytest stuff is on my todo list I haven't done it yet but maybe sometime tomorrow I will finish this. 

## Related issue number

- #1183 
- #1178 
- #1171 

## Checklist

### DIFF
--- a/multidict/__init__.py
+++ b/multidict/__init__.py
@@ -20,7 +20,6 @@ __all__ = (
     "upstr",
     "istr",
     "getversion",
-    "get_include",
 )
 
 __version__ = "6.5.1.dev0"
--- a/multidict/__init__.py
+++ b/multidict/__init__.py
@@ -5,6 +5,7 @@ multidict. It behaves mostly like a dict but it can have
 several values for the same key.
 """
 
+import pathlib
 from typing import TYPE_CHECKING
 
 from ._abc import MultiMapping, MutableMultiMapping
@@ -20,6 +21,7 @@ __all__ = (
     "upstr",
     "istr",
     "getversion",
+    "get_include",
 )
 
 __version__ = "6.5.1.dev0"
@@ -57,3 +59,15 @@ else:
 
 
 upstr = istr
+
+# Inspired by Numpy
+
+
+def get_include() -> str:
+    """
+    Get multidict headers for compiling
+    multidict with other C Extensions or
+    cython code
+    """
+
+    return str(pathlib.Path(__file__).parent)
--- a/multidict/__init__.py
+++ b/multidict/__init__.py
@@ -5,7 +5,6 @@ multidict. It behaves mostly like a dict but it can have
 several values for the same key.
 """
 
-import pathlib
 from typing import TYPE_CHECKING
 
 from ._abc import MultiMapping, MutableMultiMapping
--- a/multidict/__init__.py
+++ b/multidict/__init__.py
@@ -59,15 +59,3 @@ else:
 
 
 upstr = istr
-
-# Inspired by Numpy
-
-
-def get_include() -> str:
-    """
-    Get multidict headers for compiling
-    multidict with other C Extensions or
-    cython code
-    """
-
-    return str(pathlib.Path(__file__).parent)
--- a/multidict/_multilib/capsule.h
+++ b/multidict/_multilib/capsule.h
@@ -101,7 +101,7 @@ MultiDict_SetDefault(void* state_, PyObject* self, PyObject* key,
         _invalid_type();
         return NULL;
     }
-    return md_set_default(self, key, _default);
+    return md_set_default((MultiDictObject*)self, key, _default);
 }
 
 /// @brief Remove all items where key is equal to key from d.
@@ -118,7 +118,7 @@ MultiDict_Del(void* state_, PyObject* self, PyObject* key)
         _invalid_type();
         return -1;
     }
-    if (md_del(self, key) < 0) {
+    if (md_del((MultiDictObject*) self, key) < 0) {
         PyErr_SetObject(PyExc_KeyError, key);
         return -1;
     }
@@ -141,45 +141,6 @@ MultiDict_Version(void* state_, PyObject* self)
     return md_version((MultiDictObject*)self);
 }
 
-// MultiDict_CreatePosMarker & MultiDict_Next Took inspiration from PyDict_Next
-// it's implementations are held subject to being changed.
-
-/// @brief Creates a new positional marker for a multidict to iterate
-/// with when being utlizied with `MultiDict_Next`
-/// @param state_ the module state
-/// @param self the multidict to create a positional marker for
-/// @param pos the positional marker to be created
-/// @return 0 on sucess, -1 on failure along with `TypeError` exception being
-/// thrown
-static int
-MultiDict_CreatePosMarker(void* state_, PyObject* self, md_pos_t* pos)
-{
-    mod_state* state = (mod_state*)state_;
-    if (MultiDict_Check(state, self) <= 0) {
-        _invalid_type();
-        return -1;
-    }
-    md_init_pos((MultiDictObject*)self, pos);
-    return 0;
-}
-
-// I'll fill out this documentation once I figure out if users want this...
-
-static int
-MultiDict_Next(void* state_, PyObject* self, md_pos_t* pos,
-               PyObject** identity, PyObject** key, PyObject** value)
-{
-    mod_state* state = (mod_state*)state_;
-    if (MultiDict_Check(state, self) <= 0) {
-        _invalid_type();
-        return -1;
-    } else if (pos == NULL) {
-        PyErr_SetString(PyExc_ValueError,
-                        "positional marker cannot be left NULL");
-        return -1;
-    }
-    return md_next((MultiDictObject*)self, &pos, identity, key, value);
-}
 
 /// @brief Determines if a certain key exists a multidict object
 /// @param state_ the module state
@@ -214,8 +175,10 @@ MultiDict_GetOne(void* state_, PyObject* self, PyObject* key)
     }
     PyObject* ret = NULL;
     if (md_get_one((MultiDictObject*)self, key, &ret) < 0) {
+        return NULL;
+    } else if (ret == NULL){
         PyErr_SetObject(PyExc_KeyError, key);
-    };
+    }
     return ret;
 }
 
@@ -236,9 +199,9 @@ MultiDict_Get(void* state_, PyObject* self, PyObject* key)
     }
     PyObject* ret = NULL;
     if ((md_get_one((MultiDictObject*)self, key, &ret) < 0)) {
-        Py_RETURN_NONE;
+        return NULL;
     }
-    return ret;
+    return (ret == NULL) ? Py_None : ret;
 }
 
 /// @brief Return a list of all values for *key* if *key* is in the
@@ -258,8 +221,12 @@ MultiDict_GetAll(void* state_, PyObject* self, PyObject* key)
     }
     PyObject* ret = NULL;
     if (md_get_all((MultiDictObject*)self, key, &ret) < 0) {
-        PyErr_SetObject(PyExc_KeyError, key);
+        return NULL;
     };
+    if (ret == NULL){
+        PyErr_SetObject(PyExc_KeyError, key);
+        return NULL;
+    }
     return ret;
 }
 
@@ -299,10 +266,16 @@ MultiDict_PopOne(void* state_, PyObject* self, PyObject* key)
         _invalid_type();
         return NULL;
     }
-    PyObject* ret;
+    PyObject* ret = NULL;
     if (md_pop_one((MultiDictObject*)self, key, &ret) < 0) {
         PyErr_SetObject(PyExc_KeyError, key);
-    };
+        return NULL;
+    }
+
+    if (ret == NULL){
+        PyErr_SetObject(PyExc_KeyError, key);
+        return NULL;
+    }
     return ret;
 }
 
@@ -320,10 +293,14 @@ MultiDict_PopAll(void* state_, PyObject* self, PyObject* key)
         _invalid_type();
         return NULL;
     }
-    PyObject* ret;
+    PyObject* ret = NULL;
     if (md_pop_all((MultiDictObject*)self, key, &ret) < 0) {
-        PyErr_SetObject(PyExc_KeyError, key);
+        return NULL;
     };
+    if (ret == NULL){
+        PyErr_SetObject(PyExc_KeyError, key);
+        return NULL;
+    }
     return ret;
 }
 
--- a/multidict/_multilib/capsule.h
+++ b/multidict/_multilib/capsule.h
@@ -83,7 +83,7 @@ multidict_new_capsule(mod_state *state)
     PyObject *ret =
         PyCapsule_New(capi, MultiDict_CAPSULE_NAME, multidict_capsule_destructor);
     if (ret == NULL) {
-        capsule_free(capi);
+        multidict_capsule_free(capi);
     }
     return ret;
 }
--- a/multidict/_multilib/capsule.h
+++ b/multidict/_multilib/capsule.h
@@ -446,20 +446,20 @@ MultiDict_Equals(void* state_, PyObject* self, PyObject* other)
 /// the Heap Normally you won't be needing to call this
 /// @param capi The Capsule Object To Free
 static void
-multidict_capsule_free(MultiDict_CAPI* capi)
+capsule_free(MultiDict_CAPI* capi)
 {
     PyMem_Free(capi);
 }
 
 static void
-multidict_capsule_destructor(PyObject* o)
+capsule_destructor(PyObject* o)
 {
     MultiDict_CAPI* capi = PyCapsule_GetPointer(o, MultiDict_CAPSULE_NAME);
-    multidict_capsule_free(capi);
+    capsule_free(capi);
 }
 
 static PyObject*
-multidict_new_capsule(mod_state* state)
+new_capsule(mod_state* state)
 {
     MultiDict_CAPI* capi =
         (MultiDict_CAPI*)PyMem_Malloc(sizeof(MultiDict_CAPI));
@@ -490,9 +490,9 @@ multidict_new_capsule(mod_state* state)
     capi->MultiDict_Equals = MultiDict_Equals;
 
     PyObject* ret = PyCapsule_New(
-        capi, MultiDict_CAPSULE_NAME, multidict_capsule_destructor);
+        capi, MultiDict_CAPSULE_NAME, capsule_destructor);
     if (ret == NULL) {
-        multidict_capsule_free(capi);
+        capsule_free(capi);
     }
     return ret;
 }
--- a/multidict/_multilib/capsule.h
+++ b/multidict/_multilib/capsule.h
@@ -93,7 +93,7 @@ MultiDict_Clear(void* state_, PyObject* self)
 /// @param _default the default value to have inserted
 /// @return default on sucess, NULL on failure
 PyObject*
-Multidict_SetDefault(void* state_, PyObject* self, PyObject* key,
+MultiDict_SetDefault(void* state_, PyObject* self, PyObject* key,
                      PyObject* _default)
 {
     mod_state* state = (mod_state*)state_;
@@ -472,7 +472,7 @@ multidict_new_capsule(mod_state* state)
     capi->MultiDict_New = MultiDict_New;
     capi->MultiDict_Add = MultiDict_Add;
     capi->MultiDict_Clear = MultiDict_Clear;
-    capi->MultiDict_SetDefault = Multidict_SetDefault;
+    capi->MultiDict_SetDefault = MultiDict_SetDefault;
     capi->MultiDict_Del = MultiDict_Del;
     capi->MultiDict_Version = MultiDict_Version;
     capi->MultiDict_Contains = MultiDict_Contains;
--- a/multidict/_multilib/capsule.h
+++ b/multidict/_multilib/capsule.h
@@ -12,7 +12,6 @@ extern "C" {
 #include "hashtable.h"
 #include "state.h"
 
-
 inline static void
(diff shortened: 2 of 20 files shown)


Vizonex:
After I sleep I will need to investigate This error I get when running pytest 
```
(multidict) PS ...> pytest tests/test_capi.py
==================================================================== test session starts =====================================================================
platform win32 -- Python 3.10.18, pytest-8.4.0, pluggy-1.6.0
codspeed: 3.2.0 (disabled, mode: walltime, timer_resolution: 100.0ns)
rootdir: [...]
configfile: pytest.ini
plugins: cov-6.1.0, codspeed-3.2.0
collected 3 items

tests\test_capi.py Windows fatal exception: access violation
 ```
 The other 2 tests work just fine except for md_new(0).

**Comment by Vizonex** [comment]
@asvetlov I solved the problem with `MultiDict_New` Adding a reference count by 1 stops the segfault.

**Comment by Vizonex** [comment]
When I'm done testing the C stuff I will redo the Cython API in the same format as testcapi to keep things organized.

File: multidict/multidict_api.h [hunk 1]
```diff
@@ -22,60 +23,348 @@ typedef struct {
        if the client is compiled with older multidict_api.h header
     */
 
-    void *state;
+    void* state;
```
asvetlov, line 26: I'm curious, why does the diff exist?
Did you run `pre-commit run --all-files` first?
I've chosen the style without taking too much care of it; it could be tuned if needed. Too many options to twist!
Anyway, the main benefit of an autoformatter is that all the code is in the same style, even if it sometimes looks a little worse than we want.
If you want to tune `.clang-format` -- your are welcome, please create a separate PR for this.

multidict/multidict_api.h
```diff
@@ -22,60 +23,348 @@ typedef struct {
...
+    int (*MultiDict_Clear)(void* state, PyObject* self);
+
+    PyObject* (*MultiDict_SetDefault)(void* state, PyObject* self,
+                                      PyObject* key, PyObject* _default);
+
+    int (*MultiDict_Del)(void* state, PyObject* self, PyObject* key);
+    uint64_t (*MultiDict_Version)(void* state, PyObject* self);
+
+    // proposed but IDK yet...
+    // static int MultiDict_CreatePosMarker(void* state, PyObject* self,
```
asvetlov, line 42: What should this function do?

multidict/multidict_api.h
```diff
@@ -22,60 +23,348 @@ typedef struct {
...
+    // md_pos_t* pos)
+
+    // static int MultiDict_Next(void* state, PyObject* self,
```
asvetlov, line 45: I'm still not sure what API should be; let's implement the function later to simplify the review

multidict/multidict_api.h
Vizonex: I just ran the reformatter.

multidict/__init__.py
```diff
@@ -57,3 +59,15 @@
...
+
+
+def get_include() -> str:
+    """
+    Get multidict headers for compiling
+    multidict with other C Extensions or
+    cython code
+    """
+
+    return str(pathlib.Path(__file__).parent)
```
asvetlov: We have two options here.
1. Drop the method, document `import multidict, include_path=multidict.__path__` as I did in `testcapi/setup.py`
2. Keep the function but return bare `__path__`; `pathlib` usage is redundant here

asvetlov (commented):
Looks good, thanks!
Please let me carefully re-read the source again a little later before the merging.

multidict/multidict_api.h
Vizonex: @asvetlov I was trying to recreate the PyDict_Next Iterator but I didn't think people would exactly approve of it so I was waiting for someone to review before I decided to either throw it out or keep it in.

multidict/__init__.py
Vizonex: @asvetlov There was an issue when I tried compiling it but it may have been a configuration problem on my end so met let me go back and revert this change.

Vizonex:
@asvetlov I'm going to finish writing all the tests should we worry about patching up the bugs now or later?

Vizonex:
@asvetlov Doesn't seem like I can merge this is there anything I'm missing that you would suggest I add before I start making more things later so that I don't overburden anyone?

multidict/multidict_api.h
webknjaz: This is because there are two chains of commits due to the merge commit produced by `git pull`. One has to understand git well enough to avoid such cases. In general, newbies get into this trap when they try to collaborate on the same remote branch with multiple writers. A proper rebase would fix it.

Vizonex: > This is because there are two chains of commits due to the merge commit produced by `git pull`. One has to understand git well enough to avoid such cases. In general, newbies get into this trap when they try to collaborate on the same remote branch with multiple writers. A proper rebase would fix it.

I actually had a few hiccups with pre-commit coming in but it hasn't caused me extreme problems at all. In fact this pr is mostly done all tests that I've done from my computer were all successful. 

The only thing I'm waiting on now is to merge this chunk in and then open a new pull request but with some more functions my plan is to do all of this in order. I discovered that doing everything in small chunks is a lot more effective than trying to tackle on big thing all together. In all it keeps everything more organized. 

Currently My todo list is as follows.

- [X] MultiDict
- istr 
- MultiDictProxy
- The CI classes will come last so that we can plan it out more carefully
- Documentation on how to use and compile the C-API so that noobs are not confused about how to approach it.
- Cython branch will be setup once the C-API is done but I will keep it under the same format as the C-API and the module, the setup will be configured to same way as `testcapi` only this time I will name it `testcythonapi`. 
- Optimizing the Aiohttp http-parser with the new cython api after the next multidict release has been introduced with these features is the endgame.

Vizonex:
While I wait let me fill out the documentation for these new functions we added to the capi since I figured out how to do so. 
https://www.sphinx-doc.org/en/master/usage/domains/c.html

multidict/_multilib/capsule.h
```diff
@@ -39,38 +47,404 @@ MultiDict_New(void *state_, int prealloc_size)
...
 
+/// @brief Clears a multidict object and removes all it's entries
+/// @param state_ the module state pointer
+/// @param self the multidict object
+/// @return 0 if success otherwise -1 , will raise TypeError if MultiDict's
+/// Type is incorrect
+static int
+MultiDict_Clear(void* state_, PyObject* self)
+{
+    // TODO: Macro for repeated steps being done?
```
asvetlov, line 79: Maybe the macro is a good idea, please feel free to try

multidict/_multilib/capsule.h
```diff
@@ -39,38 +47,404 @@ MultiDict_New(void *state_, int prealloc_size)
...
+static int
+MultiDict_Del(void* state_, PyObject* self, PyObject* key)
+{
+    mod_state* state = (mod_state*)state_;
+    if (MultiDict_Check(state, self) <= 0) {
+        _invalid_type();
+        return -1;
+    }
+    if (md_del((MultiDictObject*)self, key) < 0) {
+        PyErr_SetObject(PyExc_KeyError, key);
```
asvetlov, line 122: This line converts all errors (for example, `TypeError` if passed `key` is not a string) to `KeyError`; it is not good.

Better to just return a result of `md_del()` as is; `md_del()` already sets `KeyError` if the key is not found.

multidict/_multilib/capsule.h
```diff
@@ -18,18 +18,26 @@ _invalid_type()
     PyErr_SetString(PyExc_TypeError, "self should be a MultiDict instance");
 }
 
-static PyTypeObject *
-MultiDict_GetType(void *state_)
+/// @brief Gets the multidict type object
```
asvetlov, line 21: Regarding Doxigen comments -- I'm not sure if we need it.

We should either set up a Doxygen docs generator, as we already do for Sphinx docs, or we should drop these comments.

Without an automatic generator/checker, the comments eventually will go out of sync with the code itself; misleading comments are harmful.

Taking into account that we use Sphinx already, maybe better to document C API in sphinx also?

Aha, I see the comment that you already started Sphinx docs. In this case, Doxigen comments don't make sense.

multidict/_multilib/capsule.h
```diff
@@ -39,38 +47,404 @@ MultiDict_New(void *state_, int prealloc_size)
...
+static PyObject*
+MultiDict_GetOne(void* state_, PyObject* self, PyObject* key)
+{
+    mod_state* state = (mod_state*)state_;
+    if (MultiDict_Check(state, self) <= 0) {
+        _invalid_type();
+        return NULL;
+    }
+    PyObject* ret = NULL;
+    if (md_get_one((MultiDictObject*)self, key, &ret) < 0) {
```
asvetlov, line 176: `md_get_one()` is designed to quickly figure out if the key exists or not without analyzing Python exception.

`PyErr_ExceptionMatches()` + `PyErr_Clear()` is much slower than `if (ret == NULL)`.

I  think `MultiDict_GetOne()` should follow [`PyDict_GetItemRef`](https://docs.python.org/3/c-api/dict.html#c.PyDict_GetItemRef). 
As I see, now `md_get_one` returns `-1` if an exception was raised, `0` otherwise. Maybe better return `0` if not found, `1` if found. Let's do it in a separate PR against the master branch; the fix is not directly related to CAPI itself.

multidict/_multilib/capsule.h
```diff
@@ -39,38 +47,404 @@ MultiDict_New(void *state_, int prealloc_size)
...
+
+/// @brief Return the **first** value for *key* if *key* is in the
+/// dictionary, else None.
+/// @param state_ the module state pointer
+/// @param self the multidict object
+/// @param key the key to get one item from
+/// @return returns a default value on success,  NULL is returned and raises
+/// `TypeError` on failure
+static PyObject*
+MultiDict_Get(void* state_, PyObject* self, PyObject* key)
```
asvetlov, line 192: Do we need `MultiDict_Get()` if `GetOne()`/`GetAll()` also provided?
In python API, `MultiDict.get()` exists because multidict implements `collections.abc.Mapping` interface.
In C, we do not have such a contract.
If the function doesn't exist, nobody should learn how the function works.

multidict/_multilib/capsule.h
```diff
@@ -39,38 +47,404 @@ MultiDict_New(void *state_, int prealloc_size)
...
+    if (MultiDict_Check(state, self) < 0) {
+        _invalid_type();
+        return NULL;
+    }
+    PyObject* ret = NULL;
+    if (md_get_all((MultiDictObject*)self, key, &ret) < 0) {
+        return NULL;
+    };
+    if (ret == NULL) {
+        PyErr_SetObject(PyExc_KeyError, key);
```
asvetlov, line 226: The same note as for `GetOne()`

multidict/_multilib/capsule.h
```diff
@@ -39,38 +47,404 @@ MultiDict_New(void *state_, int prealloc_size)
...
+        return NULL;
+    }
+    PyObject* ret = NULL;
+    if (md_pop_one((MultiDictObject*)self, key, &ret) < 0) {
+        PyErr_SetObject(PyExc_KeyError, key);
+        return NULL;
+    }
+
+    if (ret == NULL) {
+        PyErr_SetObject(PyExc_KeyError, key);
```
asvetlov, line 275: Again, we shouldn't raise an exception, but maybe modify `md_pop_one()` to return `0` or `1` depending on the key's existence.

multidict/_multilib/capsule.h
```diff
@@ -39,38 +47,404 @@ MultiDict_New(void *state_, int prealloc_size)
...
+}
+
+/// @brief Remove and return a value from the dictionary.
+/// @param state_ the module state pointer
+/// @param self the multidict object
+/// @param key the key to remove
+/// @return corresponding value on success or None, otherwise raises
+/// `TypeError` or `KeyError` and returns NULL
+static PyObject*
+MultiDict_Pop(void* state_, PyObject* self, PyObject* key)
```
asvetlov, line 239: `MultiDict.pop()` exists because `collections.abc.MutableMapping` requires it; in C API `MultiDict_PopOne()` + `MultiDict_PopAll()` should be enough.

multidict/_multilib/capsule.h
```diff
@@ -39,38 +47,404 @@ MultiDict_New(void *state_, int prealloc_size)
...
+    if (MultiDict_Check(state, self) <= 0) {
+        _invalid_type();
+        return NULL;
+    }
+    PyObject* ret = NULL;
+    if (md_pop_all((MultiDictObject*)self, key, &ret) < 0) {
+        return NULL;
+    };
+    if (ret == NULL) {
+        PyErr_SetObject(PyExc_KeyError, key);
```
asvetlov, line 300: The same as above, don't set an exception;

multidict/_multilib/capsule.h
```diff
@@ -39,38 +47,404 @@ MultiDict_New(void *state_, int prealloc_size)
...
+        PyErr_SetString(PyExc_TypeError,
+                        "self should be a MultiDict instance");
+        return -1;
+    }
+    if (MultiDict_Check(state, other) <= 0) {
+        PyErr_SetString(PyExc_TypeError,
+                        "other should be a MultiDict instance");
+        return -1;
+    }
+    return md_update_from_ht(
```
asvetlov, line 362: `md_update_from_ht` is not enough; if `update` is not zero `md_post_update()` should be called to remove possible duplicates.  Otherwise, multidict is in a broken state.
`md_update_from_ht()` doesnt call `md_post_update()` because the API allows something like `md.update(arg, **kwargs)`; it leads to `md_update_from_ht()` + `md_update_from_dict()` + `md_post_update()`.

I'm not sure what 'update' API should be. We could expose all four `md_update_from_ht`, `md_update_from_dict`, `md_update_from_seq`, `md_post_update`, but it is error-prone if a user forgets to call post-update.

The single `int MultiDict_Extend(PyObject* self, PyObject * arg, PyObject **dict, bool update)` is not perfect because the implementation should check the type of `arg` to figure out what to do; the check is not for free in this case; it basically looks like stairs from `_multidict.c:_multidict_extend()`.

On the other hand, we have `MultiDict.update()` and `MultiDict.extend()` in Python API.

Maybe, it would be better to provide `MultiDict_ExtendFromMultiDict(..., bool stack)`  and write in the docs that the function works like `md.update()` if `stack` is non-zero, `md.extend()` otherwise? I'm not a native speaker, I don't know what word could be better than `stack` in our case. Or `MultiDict_UpdateFromMultiDict(..., bool extend)`? Anyway, double `update` looks clumsy in public API, doesn't it? 
The internal namings for `md_update_from_ht(..., bool update)` are not ideal, but they are not visible to users. I'd like to have a public API better designed.

`MultiDict_ExtendFromMultiDict()` should call `md_post_update()` if we do the updation, not extension.

By this, a user cannot implement `md.update(arg, **kwargs)`. I think it is a very rare operation, especially in C code. We could live with this limitation.

multidict/_multilib/capsule.h
```diff
@@ -39,38 +47,404 @@ MultiDict_New(void *state_, int prealloc_size)
...
+static int
+MultiDict_UpdateFromDict(void* state_, PyObject* self, PyObject* kwds,
+                         bool update)
+{
+    mod_state* state = (mod_state*)state_;
+    if (MultiDict_Check(state, self) <= 0) {
+        _invalid_type();
+        return -1;
+    }
+    return md_update_from_dict((MultiDictObject*)self, kwds, update);
```
asvetlov, line 382: Note: `md_update_from_dict` accepts only `dict` objects but fails on classes derived from `dict`.
We should either relax the check or reflect it in our docs.
In Python interface, all non-dict mappings are handled as `md_update_from_seq(arg.items())`

multidict/_multilib/capsule.h
```diff
@@ -39,38 +47,404 @@ MultiDict_New(void *state_, int prealloc_size)
...
+/// TypeError
+static int
+MultiDict_Equals(void* state_, PyObject* self, PyObject* other)
+{
+    mod_state* state = (mod_state*)state_;
+    if (MultiDict_Check(state, self) <= 0) {
+        _invalid_type();
+        return -1;
+    }
+    if (PyMapping_Check(other)) {
```
asvetlov, line 419: `MultiDict_Check` should go first. `PyMapping_Check` succeeds for multidict objects, which leads to a correct but slow path.

multidict/_multilib/capsule.h
```diff
@@ -39,38 +47,404 @@ MultiDict_New(void *state_, int prealloc_size)
...
+
+/// @brief Checks to see if a multidict matches another dictionary or multidict
+/// object
+/// @param state_ the module state pointer
+/// @param self the multidict object
+/// @param other the corresponding object to check against
+/// @return 1 if true, 0 if false, -1 if failue occured follwed by raising a
+/// TypeError
+static int
+MultiDict_Equals(void* state_, PyObject* self, PyObject* other)
```
asvetlov, line 412: I doubt if the function should be exposed; standard Python containers (`dict`, `list`, `tuple`, `set`) have no comparison functions, `PyObject_RichCompare()` does its job well.
The only exception is `str`, which has a rich set of comparison functions to work with objects, ascii and utf8 strings.

multidict/multidict_api.h
asvetlov: > Cython branch will be setup once the C-API is done but I will keep it under the same format as the C-API and the module, the setup will be configured to same way as testcapi only this time I will name it testcythonapi.
Minor: `testcapi` could have two modules, one for C api test helpers and another one for Cython. Maybe we can even provide the same API by both wrappers to run the same api tests for both implementations, like we run the same tests suite for pure python and C extensions.

testcapi/testcapi/_api.c
```diff
@@ -26,7 +36,7 @@ static PyObject *
 md_new(PyObject *self, PyObject *arg)
 {
     mod_state *state = get_mod_state(self);
-    return MultiDict_New(state->capi, 0);
+    return Py_NewRef(MultiDict_New(state->capi, 0));
```
asvetlov, line 39: This is suspicious.
`MultiDict_New` should return a new reference; double incref is not needed.

testcapi/testcapi/_api.c
```diff
@@ -13,6 +14,15 @@ get_mod_state(PyObject *mod)
     return state;
 }
 
+
+#define PyBool_As_CBool(obj) \
+    PyObject_IsTrue(obj) ? true: false
+
+#define RETURN_NULL_OR_NEWREF(ITEM) \
+    PyObject* REF = ITEM; \
+    return (REF != NULL) ? Py_NewRef(REF) : NULL
```
asvetlov, line 23: Why newref here? It looks like a memory leak generation.

multidict/_multilib/capsule.h
asvetlov: Also, should we use `bool` type for the flag? CPython uses `int` in public API, maybe we should do the same?
In the internal API it could stay `bool`, that's fine.

asvetlov (commented):
Hey, I've left a bunch of comments.

Let's process them.

asvetlov:
Sorry, I've merged the PR by accident; I'll revert the merge.
PR is not ready yet.

testcapi/testcapi/_api.c
Vizonex: My reasoning behind this was that not doing so would cause the program to segfault.

Vizonex:
> I've reverted the merge into `aio-libs:capi` but I have no idea how to change the status from 'merged' back to 'in-progress', sorry.


@asvetlov  It's fine I'll try continuing where I left off in a new branch by just re-forking everything off again.

multidict/_multilib/capsule.h
Vizonex: I will I moved to #1190 but I am making sure to read over all your suggestions real quick before I move on...

multidict/_multilib/capsule.h
Vizonex: > Aha, I see the comment that you already started Sphinx docs. In this case, Doxigen comments don't make sense.

I documented everything as I was writing everything so that moving things around would be less painful to move but I think we still need a readable format to have in the api header file incase some user decides to hover over the function and wonder what it's doing immediately. Msvc's C/C++ Extension seems to autogenerate these comments and allows me to see the definitions when I hover my mouse of them. However, I am willing to remove delete or reformat them all since I'm done moving the current sphinx documentation over.

multidict/_multilib/capsule.h
Vizonex: Alright let me make sure it has that in #1190 then...

testcapi/testcapi/_api.c
Vizonex: Removing Py_NewRef here triggers Segfault when running in `pytest tests\test_capi.py`
```
tests\test_capi.py Windows fatal exception: access violation
```
Perhaps maybe I should move NewRef into `Multidict_New` instead?

Vizonex: Or Maybe Py_INCREF?

multidict/_multilib/capsule.h
Vizonex: > As I see, now `md_get_one` returns `-1` if an exception was raised, `0` otherwise. Maybe better return `0` if not found, `1` if found. Let's do it in a separate PR against the master branch; the fix is not directly related to CAPI itself.

@asvetlov Sure, I would be willing to work on that when I'm done making fixes into this branch shouldn't be too difficult to port something like that into #1190 and another branch

PR #1186, opened 2025-06-22
Run CI against all PRs, not only for PR with master base branch

asvetlov:
When @Vizonex created PR against capi working branch, these PRs didn't run CI tests.

If I understand correctly how GitHub works, it is because of the current filters for `pull_request` event.

The proposal drops the filters for `pull_request` event, all PRs will be tested regardless of the base branch name.

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -12,10 +12,6 @@ on:
     tags:
     - v*
   pull_request:
-    branches:
-    - master
-    - >-
-      [0-9].[0-9]+
   schedule:
   - cron: 0 6 * * *  # Daily 6AM UTC build
 


webknjaz (approved):
Yes, that's right. In projects set up better, it's also forbidden to pollute the upstream with topic branches in favor of forks. This makes maintaining such automations much more straightforward.

webknjaz:
This PR could use a contrib change note, FWIW.

asvetlov:
I'd prefer to skip changenotes, the PR doesn't change the contribution process but adds a reasonable check that should be applied anyway on merging the base branch into the master.

asvetlov:
> In projects set up better, it's also forbidden to pollute the upstream with topic branches in favor of forks. This makes maintaining such automations much more straightforward.

Good idea, I've created 'Enforce forks' ruleset to prevent the creation of new branches.
Existing feature branches will continue working as before until we merge and delete them.

asvetlov:
> This change log section is more about news for contributors, including the infra changes.

Doesn't it overcompicate the process? Even CPython has no such category for NEWs entries, despite CPython is a huge project with very many contributors. `numpy` has no such category as well.

I'm curious, what python projects have a section for notes for contributors?

webknjaz:
Yeah, it's sometimes seen as controversial. I've been showing this approach to many maintainers. Most like it. Some think it should be in a separate document.
But that's why it's optional. You still have to use your best judgment. However, having this option is nice and I argue that it allows occasional contributors who don't follow every commit still see notable updates.
I came up with this while setting up Towncrier in a bunch of projects and realized that the current defaults are rather misleading. One other category is packaging that is designed to communicate infra changes important to downstreams.

webknjaz:
My primary motivation was to split out things that otherwise get dumped as misc since it's a catch-all.

webknjaz:
Also note that since contrib/packaging/misc aren't for typical end-users, they are rendered at the end of the release section.

webknjaz:
Just remembered that pytest is one of the bigger projects where I implemented this: https://pytest--13552.org.readthedocs.build/en/13552/changelog.html#contributor-facing-changes

PR #1187, opened 2025-06-22
Slightly rewrite internal C API to reflect the modern dict API

asvetlov:
1. `md_get_one()`, `md_get_all()`, `md_pop_one()`, `md_pop_all()` now return `1` if key exists in multidict, `0` otherwise.
2. `md_set_default()` now also returns `1` if key exists in multidict, `0` if key doesn't exist and the default was iserted into the object.
3. All functions return `-1` in case of error.

The PR doesn't affect any public API; it is a preparation step for introducing multidict's public C API.


Vizonex:
@asvetlov I'll make sure that I merge these items into #1190 incase and start making changes as well as macros for the capsule api.

PR #1188, opened 2025-06-22
Add .codspeed folder to .gitignore

asvetlov:
The folder is created by tests running; it should never be added to the git tree.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -87,4 +87,4 @@ TAGS
 # pyenv
 .python-version
 
-.codspeed
+/.codspeed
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -86,3 +86,5 @@ TAGS
 
 # pyenv
 .python-version
+
+.codspeed
```
webknjaz: If it's in the root, let's be more specific
```suggestion
/.codspeed
```

asvetlov: Thanks!

PR #1190, opened 2025-06-22
Add More Multidict Functions C-API 

Vizonex:
## What do these changes do?

The summary of #1185 should suffice but basically the pr was accidently merged then deleted luckily I had a backup of the original code on hand so that we can resume where we left off.

Changes mainly will involve everything explained in #1185 but given with some better details and other suggestions from other contributors will go into this one.

## Related issue number

- #1185 it was accidently merged and then deleted, this will be the resumption to that pr to try and remedy that problem...

## Checklist


CHANGES/1190.feature.rst
```diff
@@ -0,0 +1 @@
+Added Edits for the C-API -- by :user:`Vizonex` and :user:`asvetlov`.
```
asvetlov: This file is not needed, changes should be generated from #1183

docs/multidict.rst
```diff
@@ -454,3 +454,231 @@ Environment variables
...
+   :retval NULL: if object fails to import
+
+
+.. c:function:: static inline PyTypeObject* MultiDict_GetType(MultiDict_CAPI* api)
+   
+   Obtains the Multidict TypeObject.
+
+   :param api: Python Capsule Pointer
+   :returns: return A CPython `PyTypeObject` is returned as a pointer
+   :retval NULL: if object fails to be obtained
```
asvetlov:
```suggestion
   :retval NULL: if an exception was raised
```

docs/multidict.rst
```diff
@@ -454,3 +454,231 @@ Environment variables
...
+   
+   Simillar to the way `Numpy <https://numpy.org/doc/stable/reference/c-api/index.html>`_ and `Datetime <https://docs.python.org/3/c-api/datetime.html>`_ work. multidict utilizes a 
+   `Python Capsule <https://docs.python.org/3/c-api/capsule.html#>`_ to expose the CAPI to other low level projects. 
+
+   .. c:var:: void* state;
+
+   Carries the module state to help lookup types and other objects.
+
+
+.. c:function:: static inline MultiDict_CAPI* MultiDict_Import()
```
asvetlov: CPython doc never mentions `static` and `inline` C attributes; I think we should remove them too

multidict/_multilib/hashtable.h
```diff
@@ -814,7 +814,7 @@ md_get_one(MultiDictObject *md, PyObject *key, PyObject **ret)
         if (tmp > 0) {
             Py_DECREF(identity);
             *ret = Py_NewRef(entry->value);
-            return 1;
+            return 0;
```
asvetlov: Please revert changes in `hashtable.h`

multidict/_multilib/capsule.h
```diff
@@ -39,61 +47,404 @@ MultiDict_New(void *state_, int prealloc_size)
...
-                     PyObject *default_, PyObject **result)
+/// @brief If key is in the dictionary  its the first value.
+/// If not, insert key with a value of default and return default.
+/// @param state_ the module state pointer
+/// @param self the MultiDict object
+/// @param key the key to insert
+/// @param _default the default value to have inserted
+/// @return default on sucess, NULL on failure
+PyObject*
+MultiDict_SetDefault(void* state_, PyObject* self, PyObject* key,
```
asvetlov, line 106: The signature should follow `md_set_default`, as it is in the base branch now

testcapi/setup.py
```diff
@@ -9,7 +9,6 @@
         "testcapi._api",
         ["testcapi/_api.c"],
         include_dirs=multidict.__path__,
-        extra_compile_args=["-O0", "-g3", "-UNDEBUG"],
```
asvetlov, line 12: Please revert

asvetlov (changes requested):
https://github.com/aio-libs/multidict/pull/1185#pullrequestreview-2948131619 review has a lot of comments that are not applied to this PR yet.

Could you please address it, or should I move these comments to the PR?

multidict/_multilib/hashtable.h
Vizonex: alright one sec...

Vizonex:
@asvetlov I'm wondering if I should just go back to what I had yesterday and then make whatever changes I should've made from that point on or is it too late to turn around?

Vizonex:
I'm going to pause for now and pick up again tomorrow but with a new branch this has just been mentally taxing. I think I spent 7 hours just trying to fix things unfortunately and it lead me to not seeing anything being commented.

multidict/_multilib/capsule.h
```diff
@@ -12,24 +12,35 @@ extern "C" {
...
+    if (MultiDict_Check(((mod_state*)STATE), (SELF)) <= 0) {         \
+        PyErr_Format(PyExc_TypeError,                                \
+                     #SELF " should be a MultiDict instance not %s", \
+                     Py_TYPE(SELF)->tp_name);                        \
+        return ON_FAIL;                                              \
+    }
 
-static PyTypeObject *
-MultiDict_GetType(void *state_)
+#define __MULTIDICT_RETURN_IF_FOUND(FUNC, SELF, KEY, ITEM) \
```
asvetlov: Underlying `md_*()` functions should return correct values, `1` if found and `0` if not found. 
Please revert `hashtable.h` and drop the macros.
If your tests will find some inconsistency, it should be  fixed (preferable by a separate PR)

Vizonex: > Please revert `hashtable.h` and drop the macros. 

@asvetlov I dropped the macros should I make sure the functions are actually returning 0, 1 or not yet? I would assume you want to revert again to #1187 am I correct or incorrect?

> Underlying `md_*()` functions should return correct values, `1` if found and `0` if not found. If your tests will find some inconsistency, it should be fixed (preferable by a separate PR).

Sorry if I am confused about this part, I would assume you meant for me to change `md_get_one`, `md_get_all` , `md_pop_one`, `md_pop_all`, am I correct about that?

asvetlov: Please check `testcapi._capi.md_set_default` test helper from `aio-libs:capi`. The idea is: the function accepts `multidict`, `key`, and `default`. It returns `(value, found)` tuple to explicitly check everything.
Clumsy but handy for our tests.

I suggest do the same for other test helpers where it is applicable.

asvetlov: Sorry for confusion, my initial comment was about `md_*` functions from hashtable.h, test helpers with the same name should have a little different signature as I described above; you cannot pass `PyObject**` as an argument but you can return a tuple with multiple values, it is a natural way for Python.

Vizonex: @asvetlov looking at your code I see what you might want me to try and change I like one function/approch since it removes the burden just a little bit 
```c
static int
check_nargs(const char *name, Py_ssize_t nargs, Py_ssize_t required)
{
    if (nargs != required) {
        PyErr_Format(PyExc_TypeError,
                     "%s should be called with %d arguments, got %d",
                     name,
                     required,
                     nargs);
        return -1;
    }
    return 0;
}
```
I'll probably merge this one. I see what you said about using tuples for our tests.

```c
PyObject *val = PyBool_FromLong(ret);
if (val == NULL) {
      Py_CLEAR(result);
      return NULL;
 }
 return PyTuple_Pack(2, result, val);
 ```

do you think `get_all`, `get_one`, `pop_one` and `pop_all` could be applied the same way or are we just talking about `setdefault`?

Comment (asvetlov) [reply multidict/_multilib/capsule.h]: Please apply the same approach to all functions you mentioned.

And yes, I think `check_nargs()` is good enough, in tests we don't need comprehensive error texts.
But the check is essential to prevent python crash because of incorrect number of arguments passing accidentally.

Comment (Vizonex) [reply multidict/_multilib/capsule.h]: let me make sure hashtable.h also matches the one shown here https://github.com/aio-libs/multidict/blob/capi/multidict/_multilib/hashtable.h so that I can't screw this up again. 
I think to help prevent me from messing up I'll try to be as verbose as I can about what I am doing.

Comment (Vizonex) [reply multidict/_multilib/capsule.h]: I'll do the same with https://github.com/aio-libs/multidict/blob/capi/multidict/_multidict.c since it's also changed from what I have

Comment (Vizonex) [reply multidict/_multilib/capsule.h]: @asvetlov One change I am making however is that the argument count check should come before obtaining the state so that failure can result in a much faster exit
```c
static PyObject *
md_type(PyObject *self, PyObject *const *args, Py_ssize_t nargs)
{
    if (check_nargs("md_type", nargs, 0) < 0) {
        return NULL;
    }
    mod_state *state = get_mod_state(self);
    return Py_NewRef(MultiDict_GetType(state->capi));
}
```
another idea would be a macro or function that just gets the capi out of the box I wrote one right here but I'll let you determine if we will use it or not...
```c
static inline MultiDict_CAPI* 
get_capi(PyObject* mod){
    return get_mod_state(mod)->capi;
}

// OR

#define GET_CAPI(mod) get_mod_state(mod)->capi
```

asvetlov: I'm ok with both proposals. For the the helper please use `get_capi()` function.
Inline functions are slightly safe than macroses where they are applicable

Vizonex: I'll be redoing the tests that now return tuple objects. But I think the test_capi.py functions should be kept in alphabetical order.

Vizonex: I will also be checking for misses so I had to make a tweak this not upstream yet
```c
static PyObject* handle_reuslt(int ret, PyObject* result){
    if (ret < 0) {
        return NULL;
    }
    // Test if we missed 
    if (ret == 0){
        return PyTuple_Pack(2, Py_False, Py_None);
    }
    assert(result != NULL);
    PyObject *val = PyBool_FromLong(ret);
    if (val == NULL) {
        Py_CLEAR(result);
        return NULL;
    }
    return PyTuple_Pack(2, result, val);
}
```

Vizonex: You forgot to put a pointer into the null check in md_get_all() so I did it for you, I'll do the same for the rest

Before
```c
    md_finder_cleanup(&finder);
    Py_DECREF(identity);
    return ret != NULL;
 ```
 https://github.com/aio-libs/multidict/blob/2b653ea7de88070b5a1f89f3e45828b36d5275c7/multidict/_multilib/hashtable.h#L871
 
After
```c
    md_finder_cleanup(&finder);
    Py_DECREF(identity);
    return *ret != NULL;
 ```

Comment (asvetlov) [reply multidict/_multilib/capsule.h]: Great, thanks!

I'll review tomorrow, ok.

Regarding alphabetical order -- I support it but sometimes better to tweak the rule.

For example, we have 'pop' group. PopOne first, then PopAll, then PopItem.
The order is good for docs reading I believe.

Also, complex functions for multidict updates should go last.

If we have an agreement, then let's keep the order in all multidict_api.h, capsule.h, and _capi.c test helper + test_capi.py

I suggest not touching hashtable.h file, the functions are ordered as bottoms reuse upper functions, it helps with the file reading

If you want please feel free to reorder. Otherwise, I can live with existing state.

Comment (Vizonex) [reply multidict/_multilib/capsule.h]: Another problem I just figured out about is that in the function Multidict_New the Real reason it's tirggering Segfaults is likely due to weakref possibly being unmanaged.

Comment (asvetlov) [reply multidict/_multilib/capsule.h]: I'll take a look at MD_New() . The function should return an object with PY_REFCNT(o) == 1.
If it is 0 -- we have an error.
Sorry, I'm out ofy laptop and cannot check it right now

Comment (Vizonex) [reply multidict/_multilib/capsule.h]: @asvetlov After calling md_new(0) it doesn't crash until it reaches the GC I don't know a whole lot about gc weakrefs but I tracked the segfault to this line. 
https://github.com/aio-libs/multidict/blob/2b653ea7de88070b5a1f89f3e45828b36d5275c7/multidict/_multidict.c#L468
This is a pretty weird bug I don't think its well documented.

**Comment by Vizonex** [comment]
Tomorrow Or Later I will see about tracking down the gc bug.

File: multidict/_multilib/capsule.h (hunk 7 above)
Comment (asvetlov) [reply multidict/_multilib/capsule.h]: What Python version? What is your reproducer?
Since you didn't apply `pre-commit` before pushing, GitHub CI cannot help me to see the error.
I highly recommend installing `pre-commit` locally and running `pre-commit install`, the next `git commit` runs pre commit hooks automatically

Comment (Vizonex) [reply multidict/_multilib/capsule.h]: > What Python version? What is your reproducer? Since you didn't apply `pre-commit` before pushing, GitHub CI cannot help me to see the error. I highly recommend installing `pre-commit` locally and running `pre-commit install`, the next `git commit` runs pre commit hooks automatically

3.10.18 I got it from this unique little [repo](https://github.com/adang1345/PythonWindows) way back when trying to hunt down vulnerability patches to 3.9.

**Comment by Vizonex** [comment]
@asvetlov I'm back. I woke up. I ran pre-commit install inside my env.

**Comment by Vizonex** [comment]
@asvetlov When I run pre-commit run --all-files here's what I get 
```
Validate GitHub Workflows.......................................................Passed
Check GitHub Workflows set timeout-minutes......................................Passed
Validate ReadTheDocs Config.....................................................Passed
yamllint........................................................................Failed
- hook id: yamllint
- exit code: 1

.clang-format
  1:4       error    wrong new line character: expected \n  (new-lines)

.github/workflows/ci-cd.yml
  1:4       error    wrong new line character: expected \n  (new-lines)

.github/FUNDING.yml
  1:4       error    wrong new line character: expected \n  (new-lines)

.github/dependabot.yml
  1:4       error    wrong new line character: expected \n  (new-lines)

.codecov.yml
  1:4       error    wrong new line character: expected \n  (new-lines)

.pre-commit-config.yaml
  1:4       error    wrong new line character: expected \n  (new-lines)

.readthedocs.yml
  1:4       error    wrong new line character: expected \n  (new-lines)

.github/workflows/reusable-cibuildwheel.yml
  1:4       error    wrong new line character: expected \n  (new-lines)

.github/workflows/auto-merge.yml
  1:4       error    wrong new line character: expected \n  (new-lines)

.github/config.yml
  1:4       error    wrong new line character: expected \n  (new-lines)

.yamllint
  1:4       error    wrong new line character: expected \n  (new-lines)

.github/workflows/codeql.yml
  1:4       error    wrong new line character: expected \n  (new-lines)

.github/workflows/reusable-linters.yml
  1:4       error    wrong new line character: expected \n  (new-lines)

ruff check......................................................................Passed
ruff format.....................................................................Passed
there must be not top-level `__init__.py` in `tests/`.......(no files to check)Skipped
changelog filenames.........................................(no files to check)Skipped
Changelog files should use a non-broken :user:`name` role.......................Passed
clang-format....................................................................Passed
MyPy, for Python 3.13...........................................................Passed
MyPy, for Python 3.11...........................................................Passed
```

**Comment by Vizonex** [comment]
Ok so it appears all I have left is documentation according to the linter this should be an easy one to fix seems I had some spelling errors when I was writing late at night.

**Comment by Vizonex** [comment]
@asvetlov  Workflow seems to be ignore testcapi but I'm sure you'll find a way to get it to run.

**Comment by asvetlov** [comment]
Sorry, this evening I was busy preparing the 6.5.1 bugfix release.
Please let me review tomorrow.

File: testcapi/testcapi/_api.c [hunk 8]
```diff
@@ -27,83 +34,246 @@ check_nargs(const char *name, Py_ssize_t nargs, Py_ssize_t required)
     return 0;
 }
 
+// Took the most repetative part and put it right here to help
+// you can get rid of this comment before the pr is finished,
+// Using a function was less confusing here than a macro - Vizonex
+
+static PyObject *
+handle_reuslt(int ret, PyObject *result)
```
asvetlov:
```suggestion
handle_result(int ret, PyObject *result)
```

Vizonex:
@asvetlov I'll leave the rest to you now, now that I've replaced the `GC_New` with `tp_alloc` All we need now is for the workflows to run the capi tests and will be all set to merge and then work on the rest of the c-api.

asvetlov:
Ok.
The only thing that I'm not sure about is `MultiDict_Equals`. 
In CPython, PyObject_RichCompare is used everywhere in public API, only PyUnicode has own comparison functions.
I doubt if we need `MultiDict_Equals`; a shorter API is better. `PyObject_RichCompare` should have almost the same performance as `MultiDict_Equals`; its usage is relatively unfrequent, I guess.

Vizonex:
@asvetlov I'm going to see about resolving a few conflicts today with this repo and `test_multidict_benchmarks` and then maybe after that we can see about merging and then we can work on adding in the rest of the api soon. Know that I will not be available on Saturday I may have a ton of stuff happening that day. After that, we should be fine.

Vizonex:
@asvetlov I'll make sure your done with your benchmarks pr and then I will proceed.

Vizonex:
@asvetlov Sounds good let me sync what I need to sync

Vizonex:
@asvetlov Is there anything left I need to go over before this gets merged or referenced?

Vizonex:
To do so I am writing a new function called MultiDict_CheckOrException and it will return a boolean I guess we could use an underscore at the start to label it not public?

CHANGES/1190.feature.rst
Vizonex: I just removed it.

docs/multidict.rst
Vizonex: Just made those changes.

multidict/_multilib/hashtable.h
Vizonex: @asvetlov Made sure that my script of `hashtable.h` matches with yours now. I guess merging with your branch fixed it for me. https://github.com/aio-libs/multidict/pull/1190/commits/5f5be17fd91bd03142e03c4f941805e933a6c01d

docs/multidict.rst
Vizonex: Here is the commit https://github.com/Vizonex/multidict/commit/a29309cce4f66e788fc939a21cf3e1962604092e

multidict/_multilib/capsule.h
Vizonex: I converted the macro to a function and it looks script in my eyes feels slightly better than with the macro now
```c
static int
MultiDict_CheckOrException(void* state_, PyObject* self)
{
    if (!(MultiDict_Check(((mod_state*)state_), self))) {
        PyErr_Format(PyExc_TypeError,
                     "object should be a MultiDict instance not %s",
                     Py_TYPE(self)->tp_name);
        return 0;
    }
    return 1;
}
```
Let me know if you want me to add this to the actual C-API Module

asvetlov:
Yesterday I did cherry-pick but forgot to push my changes.

Vizonex:
> Yesterday, I cherry-picked but forgot to push my changes. Please excuse me, it was evening.

It's fine whatever I was supposed to go to got cancelled so I came back to this sooner than expected. Seems I have a handful of conflicts to resolve today.

Vizonex:
I meant to reopen this. My bad...

Vizonex:
Let me add some new stuff  I'm throwing in...

Vizonex:
Last but not least will be CIMultiDict and CIMultiDictProxy. I will do these ones later tomorrow unless someone wants to beat me to it which I fully encourage.

Vizonex:
All I have left now is documentation, doing that would be as simple as copy and paste with a few edits.

Vizonex:
@asvetlov Shouldn't be too hard to add in since I got done with everything else now.

Vizonex:
```c
/* allocates `md_pos`, calls `md_init_pos()`, and returns the allocated structure.
The rettype is void* to hide implementation details */
void* MultiDict_InitPos(void* state, PyObject *self){
    __MULTIDICT_VALIDATION_CHECK(self, state, NULL);
    md_pos_t* pos = PyMem_Malloc(sizeof(md_pos_t));
    if (pos == NULL){
        PyErr_NoMemory();
        return NULL;
    }
    md_init_pos((MultiDictObject*)self, pos);
    return (void*)pos;
}

/* returns the next key/value pair, moves pos, returns -1 for error, 0 for end of iteration, 1 otherwise */
int MultiDict_Next(void* state, PyObject *self, void *pos, PyObject **key, PyObject **value){
    __MULTIDICT_VALIDATION_CHECK(self, state, -1);
    // Maybe a better error like typeerror could be used but for now
    // NoMemory is probably a best case scenario for why it's NULL 
    if (pos == NULL){
        PyErr_NoMemory();
        return NULL;
    }
    return md_next((MultiDictObject*)self, (md_pos_t*)pos, NULL, key, value);
}

/* rettype seems redundant as the impl never fails; 
plus the current implementation doesn't need 'md'.
I have a feeling that 'md' will be required for locks 
in free-threaded build, plus a reservation for telling 
about potential error is a good thing. */
int MultiDict_FreePos(void* state, PyObject *self, void *pos){
    __MULTIDICT_VALIDATION_CHECK(self, state, -1);
    if (pos != NULL){
        if (((md_pos_t*)pos)->version != ((MultiDictObject*)self)->version) {
            PyErr_SetString(PyExc_RuntimeError,
                        "MultiDict is changed during cleanup after iteration");
            return -1;
          
        }
        PyMem_Free(pos);
    }
    return 0;
}
```
Something like this could work. Thought about using `FreePos()` to ensure that the marker and multidict did not change before being released.

Vizonex:
I'm going to sepreate it via functions to ensure that minimal writing is required since I just finished making all the 4 different multidict types
```c
static void* 
__capi_alloc_pos_object(){
    md_pos_t* pos = PyMem_Malloc(sizeof(md_pos_t));
    if (pos == NULL){
        PyErr_NoMemory();
        return NULL;
    }
    return (void*)(pos);
}
```
Something like this would be more optimal over being called upon by 4 different functions.

Vizonex:
While I was looking how the new iterator I think I'm going to see if our http-writer pyx module could be moved to C so that it has a lower-level footprint. I'll make it into possibly a separate repo but it will utilize the new tech we worked on.

asvetlov:
> While I was looking how the new iterator I made I think I'm going to see if the http_writer.pyx module from aiohttp could be moved to C so that it has a lower-level footprint with only one function ever getting called. I'll make it into possibly a separate repo temporarily but it will utilize the new tech we worked on.

Maybe, let's see. Actually, Cython which utilizes multidict C API could be good enough

Vizonex:
@asvetlov I think I'm going to move some of the iterator functions over to something like `MultidictIter` since it comes with everything by default, this way users will be able to utilize the `MultidictIter` Object for using `Next()` and users will not have to Call to Free the position marker when iteration is over.

Vizonex:
Can't wait to see 6.7 be released. Do we still want to add cython to it or should be cram that into 6.8?

Vizonex:
Seems I added Cython in by accident but feel free to let me know if it shouldn't be autogenerated. A Handful of edits will likely fix that.

Vizonex:
I added in my Cython script generator in because the `__init__.pxd` script it generates leaked in by accident. I have now added the tool that I've used to generate it unless anybody has any better ideas on how the cython portion should be maintained.

Vizonex:
@asvetlov I wanted to provide cython support by the next update and I was wondering if I should go and document that portion of code or try moving away from trying to autogenerate that portion of code.

Vizonex:
I still haven't figured out a way to get the workflows to run the C-API tests yet. Does anybody have any input on how it should be approched?

Vizonex:
@asvetlov If I'm going to be honest maybe we should have the testcapi module in the multidict library as an underscore module as something like `multidict._testcapi` so that we don't have to waste anymore time with trying to play a game of "try brute-forcing this in and will it work?" Did you have any other ideas on compiling testcapi with the workflow that I didn't try yet?

Vizonex:
I have to get off to work today but I should be back in 7-8 hours but if anybody wants to try and disabling 3.13t from trying to install cffi and seeing if testcapi passes on all other systems and then merge the pull-request if it all of it passes be my guest.

Vizonex:
@asvetlov I came up with a temporary solution for CFFI 2.0.0 on 3.13t, I'll revert the workflow script when it's done but just know this can come back to bite us later if were not careful with how this hack is used...
```
cffi == 1.17.1; python_version=="3.13" # CFFI-2.0.0 breaks workflows under 3.13t
```

asvetlov:
Superseded by #1480

PR #1191, opened 2025-06-22
vizonex capi -- draft

asvetlov:
Resurrected version of @Vizonex work

I saved it locally just in case, and I'll delete it when the CAPI work will land on the master


asvetlov:
Superseded by #1480

PR #1192, opened 2025-06-23
Shorted fixture parametrization ids

asvetlov:
The change significantly reduces generated test names to fit better on the local terminal, in logs, and benchmark web view.

For example, `test_keys_view_xor[case-insensitive-pure-python-module]` becomes `test_keys_view_xor[ci-py]`, `test_keys_view_xor[case-sensitive-c-extension-module]` becomes `test_keys_view_xor[cs-c]`


asvetlov:
@webknjaz the PR doesn't change fixture names, but their parameterization IDs only.

Please, let's try to live with these names. They fit UI screens much better than before.

PR #1193, opened 2025-06-23
🧪 Sync `reusable-cibuildwheel.yml` w/ upstream

webknjaz:
## What do these changes do?

This is an infra change that generalizes the `cibuildwheel` workflow as a part of preparation to my effort of further extraction externally.

This includes having input names in line with `reusable-tox.yml` and configuring `cibuildwheel` via env vars passed through inputs.

Nope.

## Related issue number

N/A

## Checklist

### DIFF
--- a/CHANGES/1193.contrib.rst
+++ b/CHANGES/1193.contrib.rst
@@ -1,0 +1,3 @@
+The :file:`reusable-cibuildwheel.yml` workflow has been refactored to
+be more generic and :file:`ci-cd.yml` now holds all the configuration
+toggles -- by :user:`webknjaz`.


asvetlov:
@webknjaz could you wait for tomorrow please.
The PR looks promising but I'm too tired now, I'd like to carefully review next day.

webknjaz:
I don't need a review, really. It's thing I'll be hosting outside aio-libs, eventually. This is to ease the migration.

webknjaz:
It's going to be PyPA so no. But I'm not migrating just yet. I need to solve a few things and document usage strategies.

PR #1194, opened 2025-06-24
Fix md_get_all() return value

asvetlov:
It is a follow-up PR for #1187 that corrects retval of internal `md_get_all()` function.


asvetlov:
Thanks @Vizonex for catching.

The PR doesn't need a changenote because it corrects an internal function, and #1187 has not been released yet.

PR #1197, opened 2025-06-24
Release 6.5.1

asvetlov:


webknjaz:
The change log should have consistent writing style. I see it's currently partially in imperative mood which would be better fitting for a TODO list or Git commits. Best to either use the past tense or explanatory present simple.

CHANGES.rst
```diff
@@ -14,6 +14,52 @@ Changelog
...
+6.5.1
+=====
+
+*(2025-06-24)*
+
+
+Bug fixes
+---------
+
+- Fix a bug in C implementation when during multidict resizing the function could drop
```
asvetlov:
```suggestion
- Fixed a bug in C implementation when during multidict resizing the function could drop
```

asvetlov:
Fixed, thanks!

CHANGES.rst
```diff
@@ -14,6 +14,52 @@ Changelog
...
+  Patch by :user:`asvetlov`.
+
+  *Related issues and pull requests on GitHub:*
+  :issue:`1195`.
+
+
+Contributor-facing changes
+--------------------------
+
+- Setup pre-commit hook for auto-formatting both Python and C code.
```
webknjaz:
```suggestion
- A pair of code formatters for Python and C have been configured in the pre-commit tool.
```

asvetlov: As you wish. 
Technically 'setup' verb has the same spelling in both present and past forms, isn't it?

webknjaz: No, it's not. `set up` is a verb or an adjective. `setup` or `set-up` is a noun.

CHANGES.rst
```diff
@@ -14,6 +14,52 @@ Changelog
...
+=====
+
+*(2025-06-24)*
+
+
+Bug fixes
+---------
+
+- Fixed a bug in C implementation when during multidict resizing the function could drop
+  last item under some circumstances.
```
webknjaz: From the end-user PoV, it's unclear what this mysterious circumstance is and how to identify it. Could you outline the context?

asvetlov: It occurs when the dict has deleted items and it should be resized to add a new key/value pair.
I'm afraid, for the end user, it really looks like a mysterious circumstance, sometimes all works fine, but in other cases it doesn't.
It all depends on the actual internal state, which is completely hidden from the user. I think that my trial to describe the exact state for error reproducing doesn't help with understanding. Curious people could open the corresponding PR and see the change though, the diff has only 4 lines of code except a test.

asvetlov: A user should have 5 items in the dict, then remove one and later add another one. Something like this, but other reproducers also exist. 
For 4 and 6 items in the dict it doesn't work, the problem doesn't appear.

webknjaz: Still, write something more specific instead of building a suspense. It's not a true crime novel. The changelog should not contain references nobody knows how to understand.

asvetlov:
```suggestion
- Fixed a bug in C implementation when multidict is resized and it has deleted slots.
```

asvetlov: @webknjaz is it better?

webknjaz: Yes

CHANGES.rst
```diff
@@ -14,6 +14,51 @@ Changelog
...
+  toggles -- by :user:`webknjaz`.
+
+  *Related issues and pull requests on GitHub:*
+  :issue:`1193`.
+
+
+----
+
+
 6.5.0
```
webknjaz, line 62: Could you add a `.. note::` under the date below saying that it's yanked?

webknjaz (approved):
LGTM but maybe add one more improvement.

CHANGES.rst
asvetlov: Done

asvetlov:
@webknjaz thank you very much for fault-finding review.

PR #1199, opened 2025-06-25
Sync `reusable-cibuildwheel.yml` with upstream

webknjaz:
## What do these changes do?

This patch includes setting input defaults for sdist-related fields and passing the sdist directly into `cibuildwheel`. It also normalizes the workflow name.

Nope.

## Related issue number

N/A

## Checklist

### DIFF
--- a/.github/workflows/reusable-cibuildwheel.yml
+++ b/.github/workflows/reusable-cibuildwheel.yml
@@ -1,6 +1,9 @@
 ---
 
-name: Build wheel
+name: >-
+  ❌
+  [DO NOT CLICK]
+  Reusable cibuildwheel
 
 on:  # yamllint disable-line rule:truthy
   workflow_call:
--- a/.github/workflows/reusable-cibuildwheel.yml
+++ b/.github/workflows/reusable-cibuildwheel.yml
@@ -8,12 +8,15 @@ name: >-
 on:  # yamllint disable-line rule:truthy
   workflow_call:
     inputs:
-      dists-artifact-name:
-        description: Workflow artifact name containing dists
-        required: true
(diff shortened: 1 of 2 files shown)


.github/workflows/reusable-cibuildwheel.yml
```diff
@@ -1,18 +1,24 @@
 ---
 
-name: Build wheel
+name: >-
+  ❌
```
asvetlov, line 4: Please describe why this name is so uncommon.

webknjaz: It should show up in the sidebar with a clear message.

asvetlov: Could you provide a screenshot please?

I see that message in a context where there is not a button to click.

webknjaz: https://github.com/aio-libs/multidict/actions

asvetlov: I see the text on the left sidebar, but I don't see a button that I shouldn't click.

Don't get me wrong, if you had the reason to write the text -- I respect it.

My intention is only to understand why you wrote the text and what exactly I shouldn't click.

webknjaz: Yes, that's a sidebar item, not a button. That page is never going to have any CI runs on it. And so I've been marking such pages as "don't click".
This is essentially the same concept I introduced in CPython with "modules" that are in-repo reusable workflows and aren't allowed to have any other triggers. Plus their filenames are prefixed per convention.

PR #1200, opened 2025-06-25
Don't allocate new memory for htkeys on adding new item if the dict has deleted slots

asvetlov:


CHANGES/1200.feature.rst
```diff
@@ -0,0 +1,7 @@
+Skipped reallocation of internal htkeys structure on inserting new items if the
```
webknjaz: Can't we just use inline highlighting for things that clearly reference source code identifiers? This way, we wouldn't have to keep adding special cases to the known words list. This isn't a word, anyway. So let's not pretend like it is.
```suggestion
Stopped reallocating internal ``htkeys_t`` structure when inserting new items if the
```

docs/spelling_wordlist.txt
```diff
@@ -37,6 +37,7 @@ fallback
 fastpath
 filename
 formatters
+htkeys
```
webknjaz:
```suggestion
```

CHANGES/1200.feature.rst
asvetlov: ok, thanks!

multidict/_multilib/hashtable.h
```diff
@@ -232,16 +233,52 @@ _md_resize(MultiDictObject *md, uint8_t log2_newsize, bool update)
...
 
 static inline int
 _md_resize_for_update(MultiDictObject *md)
 {
-    return _md_resize(md, calculate_log2_keysize(GROWTH_RATE(md)), true);
+    if (md->used < md->keys->nentries) {
+        return _md_shrink(md, true);
+    } else {
+        return _md_resize(md, calculate_log2_keysize(GROWTH_RATE(md)), true);
+    }
```
webknjaz, line 281: Less nesting usually improves readability.
```suggestion
    }

    return _md_resize(md, calculate_log2_keysize(GROWTH_RATE(md)), true);
```

CHANGES/1200.feature.rst
```diff
@@ -0,0 +1,7 @@
+Skipped reallocation of internal ``htkeys_t`` structure on inserting new items if the
```
webknjaz: (writing style)
```suggestion
Stopped reallocating memory for the internal ``htkeys_t`` structure when inserting new items if the
```

CHANGES/1200.feature.rst
```diff
@@ -0,0 +1,7 @@
...
+multidict has deleted items and it could be collapsed in-place.  Removal of
+``malloc()``/``free()`` could improve the performance a little.
```
webknjaz: (writing)
```suggestion
``malloc()``/``free()`` improves the performance slightly.
```

multidict/_multilib/hashtable.h
asvetlov: The library has many places like the following:
```
if (cond) {
    return a();
} else {
    return b();
}
```
Let's keep it as is.

multidict/_multilib/hashtable.h
```diff
@@ -232,16 +233,52 @@ _md_resize(MultiDictObject *md, uint8_t log2_newsize, bool update)
...
+    htkeys_t *keys = md->keys;
+    Py_ssize_t nentries = keys->nentries;
+    entry_t *entries = htkeys_entries(keys);
+    entry_t *new_ep = entries;
+    entry_t *old_ep = entries;
+    Py_ssize_t newnentries = nentries;
+    for (Py_ssize_t i = 0; i < nentries; ++i, ++old_ep) {
+        if (old_ep->identity != NULL) {
+            if (new_ep != old_ep) {
+                *new_ep++ = *old_ep;
```
bdraco, line 248: This looks like it might be the cause of the segfault

bdraco: does new_ep always need to be incremented here even if != ?

bdraco: maybe new_ep == old_ep, first entry to be overwritten by later entries, new_ep now at wrong position, anything after writes to the wrong memory address...

asvetlov: I think no.
The loop iterates over all elements and overwrites empty ones with next items.
It basically deletes entries with `identity == NULL` and collapses unused memory holes

PR #1203, opened 2025-06-25
Fix docstring for md.update()

asvetlov:


webknjaz:
I'm not sure about this. I'd love to generate API docs from docstrings. But even if not, there's still use for these. Docstrings can show up in different development UIs in editors. I think they render this markup too.

Issue #1201, opened 2025-06-25
Unstable benchmarks for getall() and pure python implementation


asvetlov:
`getall` benchmarks like `test_cimultidict_getall_istr_hit[py]`, `test_multidict_getall_str_hit[ci-py]`, `test_multidict_getall_str_miss[cs-py]`, and others has very unstable execution time when pure python implementation is tested.

It generates misleading noise for all multidict PRs, even not related to changes in python implementation.

I have no clear strategy how to mitigate this. 

From my understanding, the source of unstable execution time is Python hashseed generator.
Because of it, hashes of Python objects used in tests are different every CI run. Sometimes it could lead to hash collisions when methods like `_HtKeys.iter_hash()` are called, sometimes hashes don't collide.

In case of collision, hash iterator needs more loops to do the job.
C implementation shares the same algorithm, but it is executed much faster than py version. 
As the result, time of C benchmarks are stable but some python tests varies a lot.

I don't very care about python execution time, in the real world C version should be used anyway.
But we should do something with our benchmarks.

The simplest solution could be setting PYTHONHASHSEED envvar to any constant value. Another option is rewriting benchmark to run them not against the constant key but a set of keys. Some of them might collide but the amortized time should be more or less the same.

Perhaps PYTHONHASHSEED is the easiest think that we could apply.

I anybody has an idea -- please share it here.

asvetlov:
`PYTHONHASHSEED` helped with the variability reduction, but a few benchmarks are still reported as unstable.

Perhaps, we need plan B.

asvetlov:
Hmm, #1207 didn't help. We need creative ideas, guys.
Ironically, we have to benchmark *something* for pure python to make sure that the function didn't degrade significantly, but we really care only about C implementation. Which doesn't make any problems with any `getall` tests version

asvetlov:
All five attempts yielded the following results: variance was reduced approximately twice, but it remains at most 33%.

The hash seed is now a constant; we need another hypothesis for the variance source. 
It could be a garbage collector run. Benchmarks have no cycles; GC could be disabled during their execution.

Additionally, it appears that increasing the number of iterations helps.

I'll try both approaches.

asvetlov:
According to [pytest-codspeed sources](https://github.com/CodSpeedHQ/pytest-codspeed/blob/a4e5901701a572f4c4745e807fdd45df75a2853e/src/pytest_codspeed/plugin.py#L244) the plugin disables GC on benchmark measuring

[...]

PR #1208, opened 2025-06-26
Stable benchmarks, fourth attempt

asvetlov:
See #1201 for details


asvetlov:
Interesting, py benchmarks used C istr version; it was never intended to work this way.

I've fixed the tests; let's monitor the impact after the fix is applied.

PR #1223, opened 2025-06-28
fix shrink corruption - test run without fix to verify test fails on CI

bdraco:


bdraco:
FAILED ../../../cibw-sdist-5180xxih/multidict-6.6.2.dev0/tests/test_mutable_multidict.py::test_multidict_shrink_regression - AssertionError: First entry k0 was lost due to memory corruption!
  assert 'k0' in <MultiDict('k2': 'v2', 'k4': 'v4', 'k6': 'v6', 'k8': 'v8', 'k4': 'v4', 'new0': 'val0', 'new1': 'val1', 'new2': 'val2',...: 'val43', 'new44': 'val44', 'new45': 'val45', 'new46': 'val46', 'new47': 'val47', 'new48': 'val48', 'new49': 'val49')>

So test is catching the issue. good

PR #1228, opened 2025-06-28
[DNM] 🧪 Instrument C-extensions to collect coverage

webknjaz:
## What do these changes do?

This patch sets up the infrastructure for collecting and reporting coverage for C-extensions. It includes updates to packaging, test runner and CI/CD.

> [!warning]
> **THE PATCH IS INCOMPLETE. THIS IS @webknjaz's PLAYGROUND FOR NOW. DO NOT MERGE!**
>
> *I've been able to generate coverage reports locally, with a series of manual commands. They still need to be integrated and cleaned up to fit well together.*

This is a contributor-facing change, allowing them to expose gaps in coverage of C-extensions.

## Related issue number

N/A

## Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -44,7 +44,9 @@ if platform.system() != "Windows" and False:
         ]
     )
 
-os.environ['CC'] = 'ccache gcc'
+if DEBUG_BUILD:
+    # https://gcovr.com/en/stable/cookbook.html#how-to-collect-coverage-for-c-extensions-in-python
+    os.environ['CC'] = 'ccache gcc'  # no distutils/setuptools equivalent
 
 extensions = [
     Extension(
--- a/setup.py
+++ b/setup.py
@@ -73,7 +73,7 @@ class TraceableBinaryExtensionCmd(build_ext):
         fullname = self.get_ext_fullname(ext.name)
         return fullname.split('.')
 
-    def _ext_tracing_file_for(self, ext) -> Path:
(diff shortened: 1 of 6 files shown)


asvetlov:
I expect a far from 100% coverage because of CPython API calls error checking.
It is not a stopper but we should life with it; not all error checks are reproducible.
E.g., man calls could return `NULL` with `MemoryError` set, but realistically there is no way to write a test for such situation. And so on.

setup.py
```diff
@@ -25,20 +48,201 @@
...
+        # NOTE: $ GCOV_PREFIX=multidict/__tracing-data__/ \
+        # NOTE:   GCOV_PREFIX_STRIP=2 \
+        # NOTE:     python -Im pytest
+        build_meta_path = tracing_data_in_build_dir / 'build-metadata.json'
+        tmp_path_length = len(pathlib.Path(self.build_temp).resolve().parents)
+        build_meta_path.write_text(
+            json.dumps(
+                {
+                    'GCOV_PREFIX': str(tracing_data_in_package_dir),
+                    'GCOV_PREFIX_STRIP': str(tmp_path_length),
```
webknjaz, line 199: In https://github.com/gcovr/gcovr/issues/1112#issuecomment-3028916187 @latk implies that this is semantically incorrect. I've only not read that comment. Backreferencing it from here for context.

Issue #1195, opened 2025-06-24
multidict 6.5.0 breaks httpie cli


mburr:
multidict 6.5.0 causes [httpie cli](https://github.com/httpie/cli/) to behave erratically.  Downgrading to multidict 6.4.4 works around the problem.

Please see https://github.com/httpie/cli/issues/1637 for details.

However I don't know if the problem is because of a bug in httpie or multidict.

asvetlov:
multidict 6.5.1 will fix the bug. The GitHub bot is preparing the release right now.

mburr:
Wow that was fast!

asvetlov:
Thank you for the reproducer; it helped me a lot in figuring out the problem's root.

asvetlov:
multidict 6.5.1 is on PyPI now

Romain-Geissler-1A:
Hi,

When trying to upgrade to multidict 6.6.2 in Fedora rawhide, I also get an issue when re-building httpie following: https://download.copr.fedorainfracloud.org/results/romaingeissler1a/python-multidict/fedora-rawhide-aarch64/09225681-httpie/builder-live.log.gz I have unfortunately no more local details about the abort error. Note: fedora used so far multidict 6.4.4.

bdraco:
Please open an issue with the stack trace, and a way to reproduce it
```
Current thread's C stack trace (most recent call first):
  Binary file "/lib64/libpython3.14.so.1.0", at _Py_DumpStack+0x5c [0xffffa1bfe14c]
  Binary file "/lib64/libpython3.14.so.1.0", at +0xc2dec [0xffffa1c02dec]
  Binary file "linux-vdso.so.1", at __kernel_rt_sigreturn+0x0 [0xffffa21c2840]
  Binary file "/lib64/libc.so.6", at +0x8f7bc [0xffffa19ff7bc]
  Binary file "/lib64/libc.so.6", at gsignal+0x1c [0xffffa19ab6fc]
  Binary file "/lib64/libc.so.6", at abort+0x2c [0xffffa1995d08]
  Binary file "/lib64/libc.so.6", at +0x34220 [0xffffa19a4220]
  Binary file "/usr/lib64/python3.14/site-packages/multidict/_multidict.cpython-314-aarch64-linux-gnu.so", at +0xb4b4 [0xffff9ec4b4b4]
[...]

PR #1229, opened 2025-06-29
Fix inconsistencies generated by _md_shrink.

Romain-Geissler-1A:
This is causing assertion failures in ASSERT_CONSISTENT if an entry used to have a non null identity. Since the new entries will be concentrated at the beginning of the array, we shall "erase" the old entries that are at the end of the array and that won't be used anymore.

This is enough to fix the core dump issues found in the testsuite of httpie in Fedora Rawhide (see https://github.com/aio-libs/multidict/issues/1195#issuecomment-3016756886)


asvetlov:
@Romain-Geissler-1A could you confirm that the error was raised by `ASSERT_CONSISTENT()` call?
I see neither a test nor a reproducer.
If it is an assertion in self-consistency check, I would suggest another fix: `memset()` the tail of the entries array with zeroes. It emulates the behavior of `_md_resize()` without requiring new memory allocation.

Romain-Geissler-1A:
Hi,

Yes, in my case, the assert failure was exactly this one: https://github.com/aio-libs/multidict/blob/8ea69422ea5e80293cec2258ecc8c1a4b3c6239e/multidict/_multilib/hashtable.h#L1939 and came from this call site: https://github.com/aio-libs/multidict/blob/8ea69422ea5e80293cec2258ecc8c1a4b3c6239e/multidict/_multilib/hashtable.h#L1313 (there is a second call to ASSERT_CONSISTENT after the call to `md_post_update` in `multidict.c` as well. I investigate all this with gdb.

Despite my effort, I was unable to reproduce this reliably any other way then building the rpm with fedora packaging tool, installing a local `httpie` from the github sources, creating a virtual environment from the httpie Makefile, then removing the installation of multidict (which by default is pure python) and replace it by the one from the system using the previously built RPM. I also tried to compile a local multidict shared object manually but then for some reason I don't get, it was working. When trying to debug calls to multidict in httpie, I noticed that the fact of adding some `print()` calls has an influence on the reproducer (with enough traces, it works...).

I can try your suggestion of zeroing out every trailing entries.

CHANGES/1229.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fix inconsistencies generated by the C implementation of ``_md_shrink()`` which might later lead to assertion failures and crash -- by :user:`Romain-Geissler-1A`.
```
webknjaz:
```suggestion
Fixed inconsistencies generated by the C implementation of ``_md_shrink()`` which might later lead to assertion failures and crash -- by :user:`Romain-Geissler-1A`.
```

Romain-Geissler-1A: Done

Romain-Geissler-1A:
Technically, if you run an OCI container 'quay.io/fedora/fedora:rawhide', and run this inside it:

```
bash-5.2# dnf install -y gdb https://download.copr.fedorainfracloud.org/results/romaingeissler1a/python-multidict/fedora-rawhide-x86_64/09227274-python-multidict/python3-multidict-6.6.2-1.fc43.x86_64.rpm https://download.copr.f
edorainfracloud.org/results/romaingeissler1a/python-multidict/fedora-rawhide-x86_64/09227274-python-multidict/python3-multidict-debuginfo-6.6.2-1.fc43.x86_64.rpm             
Updating and loading repositories:
 Fedora rawhide openh264 (From Cisco) - x86_64                                                                                                                                             100% |   5.9 KiB/s |   5.8 KiB |  00m01s
 Fedora - Rawhide - Developmental packages for the next Fedora release                                                                                                                     100% |   9.8 MiB/s |  21.8 MiB |  00m02s
Repositories loaded.
 https://download.copr.fedorainfracloud.org/results/romaingeissler1a/python-multidict/fedora-rawhide-x86_64/09227274-python-multidict/python3-multidict-debuginfo-6.6.2-1.fc43.x86_64.rpm  100% | 325.5 KiB/s | 183.9 KiB |  00m01s
 https://download.copr.fedorainfracloud.org/results/romaingeissler1a/python-multidict/fedora-rawhide-x86_64/09227274-python-multidict/python3-multidict-6.6.2-1.fc43.x86_64.rpm            100% | 122.7 KiB/s |  93.8 KiB |  00m01s
 (output skipped...)

bash-5.2# gdb /usr/lib64/python3.14/site-packages/multidict/_multidict.cpython-314-x86_64-linux-gnu.so
GNU gdb (Fedora Linux) 16.3-3.fc43
Copyright (C) 2024 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-redhat-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<https://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from /usr/lib64/python3.14/site-packages/multidict/_multidict.cpython-314-x86_64-linux-gnu.so...
Reading symbols from /usr/lib/debug/usr/lib64/python3.14/site-packages/multidict/_multidict.cpython-314-x86_64-linux-gnu.so-6.6.2-1.fc43.x86_64.debug...
(gdb) disas _md_check_consistency
Dump of assembler code for function _md_check_consistency:
Address range 0x2220 to 0x24d7: 
   0x0000000000002220 <+0>:     push   %rbp
   0x0000000000002221 <+1>:     mov    %rsp,%rbp
   0x0000000000002224 <+4>:     push   %r13
   0x0000000000002226 <+6>:     push   %r12
```

you can see that the symbol `_md_check_consistency` is defined, so that `NDEBUG` was not defined at compilation time and thus assertions are enabled.

PR #1233, opened 2025-07-27
Fix Memory leaks and add tests to prevent memory leaks during md_clear from passing

### LINKED ISSUE #1232: Memory leak in CIMultiDict.pop() method?
I noticed a big memory leak in my application.

With tracemalloc I found that memory is leaking little by little when calling the `.pop()` method.
I used multidict version: `6.6.3`

From #1134 I took an example for the script and got this:

```python
import os

import psutil
from multidict import MultiDict, __version__ as multidict_version

print(f"multidict version: {multidict_version}")


def get_memory_usage():
    process = psutil.Process(os.getpid())
    memory_info = process.memory_info()
    return memory_info.rss / (1024 * 1024)


keys = [f"X-Any-{i}" for i in range(1000)]
headers = {key: key * 2 for key in keys}


def main():
    for _ in range(1000):
        for _ in range(1000):
            result = MultiDict(headers)

            for key in keys:
                result.pop(key, None) # leak
                # result.popall(key) # leak
                # del result[key] # leak

                # result.pop('123321', None) # very slow leak

        del result
        usage = get_memory_usage()
        print(f"Memory usage: {usage:.2f} MB")
        if usage > 200:
            raise MemoryError("Memory usage exceeded 200 MB")


if __name__ == "__main__":
    main()

```

From the notable:

1. If you increase `if usage > 200` to a super large value, the script runs without errors (but should this script really consume so much memory? The maximum I saw was about 6000 MB, and at the end of the script the memory stabilizes and remains at some still large value)
2. Just in case, I tried different functions, most of them behave the same. But if you do `.pop` from a non-existent key, then the memory leaks very slowly, about 0.5MB in 1-2 minute.

Vizonex:
## What do these changes do?

@F1int0m had discovered a memory leak related to the following functions in the C extension of multidict, all the following functions are affected (This goes for both `MultiDict` and `CIMultiDict`)
- `pop(...)` & `popone(...)`
- `popall(...)`
- `__delitem__(...)` or `del item[...]`
- surprisingly `popitem(...)` is not affected so I'm pretty sure it's a ref-count bug where a there's more refs being made than needed. But having gone through the code I have not found the bug yet. However I think all versions after the Hashtable was introduced (6.6.0+) are affected. 
Reason this is a draft and not a pull request is that the bug has not been found however to prevent future bugs like this one I have added an isolated test to test for memory-leaks upon the deletion of values. However this bug is pretty sneaky and could make it's way into aiohttp under conditions where the user is modifying a `MultiDict` via deletion of keys.

- Feel free to help me change any code as I go along attempting to tweak any C Functions or if you think the isolated-test I've added should use a more intense stress-test.

~~Edit: Based on trying the test with trimming memory. This claim mentioned above might be bogus. However I shall give credit where due and say that having this test should prevent bougus claims about deletion of items casuing memory leaks from appearing in the future.~~
Edit 2: I stand corrected this is a problem.

Memory Leaks Related to item-deletion should not be allowed to pass now thanks to these changes which means that we will no longer have bogus claims about such things.

## Related issue number

- 'fixes #1232'

## Checklist


Vizonex:
@asvetlov When this passes or you decide to put this new fix into effect feel free to yank 6.6.3 for me since it leaks memory and I don't think users want to deal with that.

Vizonex:
I think I'm going to change how `md_clear` is approched a bit later today by reverting some of it's functionality but making a new function or parameter for when dellocating and deletion happens. The parameter or new function should not be exposed to the CAPI that I was working on however as it should be tied directly to the dealloc function.

asvetlov (changes requested):
Thanks for the patch!

Your fix is correct.
However, I'd like to ask you to revert all changes in `.c` files except the `md_clear()` function body.
They make a mess in git history without actually fixing anything.

asvetlov:
Don't redo please but revert only changes in C source files that are not related to `md_clear()` implementation.

Maybe I was not clear about *git history* concern. We squash pull requests before merging, so the number of commits doesn't matter. What I don't want to see are artefacts of your debugging sessions: commented out debugger helper function, meaningless replacements of Py_DECREF with Py_XDECREF etc.

If it is uncomfortable for you, I can tidy up the source myself.

Vizonex:
> Don't redo please but revert only changes in C source files that are not related to `md_clear()` implementation.

@asvetlov Ok I can do that. That shouldn't be a problem for me to take care of the one fear I had was reverting changes if a new memory-leak happened but I will take you at your word this time. 

> Maybe I was not clear about _git history_ concern. We squash pull requests before merging, so the number of commits doesn't matter. What I don't want to see are artefacts of your debugging sessions: commented out debugger helper function, meaningless replacements of Py_DECREF with Py_XDECREF etc.
I see now that makes more sense to me. I'll get rid of my debugger for you. I was using to track where the memory leak originated from since it was more or less a puzzle to me until I figured out what I needed to fix.

> If it is uncomfortable for you, I can tidy up the source myself.
No it's ok I can do it real quickly. I should be able capable of doing so.

Vizonex:
I cleaned up the artifacts now I will wait. I'm not worried if this takes any longer because I already let users who are aware of this issue know that they can temporarily use this fork I'm trying to merge until multidict has a new release.

multidict/_multidict.c
```diff
@@ -674,7 +674,6 @@ multidict_popone(MultiDictObject *self, PyObject *const *args,
     if (md_pop_one(self, key, &ret_val) < 0) {
         return NULL;
     }
```
asvetlov:
```suggestion
    }
    
```

multidict/_multilib/hashtable.h
```diff
@@ -1877,21 +1877,18 @@ md_traverse(MultiDictObject *md, visitproc visit, void *arg)
 static inline int
 md_clear(MultiDictObject *md)
 {
-    if (md->used == 0) {
+    if ((md->keys == &empty_htkeys) || (md->state == NULL)) {
         return 0;
     }
     md->version = NEXT_VERSION(md->state);
```
asvetlov, line 1883:
```suggestion
    md->version = NEXT_VERSION(md->state);
    
```

multidict/_multilib/hashtable.h
```diff
@@ -1877,21 +1877,18 @@ md_traverse(MultiDictObject *md, visitproc visit, void *arg)
...
-        if (entry->identity != NULL) {
-            Py_CLEAR(entry->identity);
-            Py_CLEAR(entry->key);
-            Py_CLEAR(entry->value);
-        }
+        // Py_CLEAR has null checks of it's own making it easier to free.
+        Py_CLEAR(entry->identity);
+        Py_CLEAR(entry->key);
+        Py_CLEAR(entry->value);
     }
```
```suggestion
    }
    
```

multidict/_multilib/hashtable.h
```diff
@@ -1910,7 +1907,6 @@ _md_check_consistency(MultiDictObject *md, bool update)
 
 #define CHECK(expr) assert(expr)
     //    do { if (!(expr)) { assert(0 && Py_STRINGIFY(expr)); } } while (0)
```
```suggestion
    //    do { if (!(expr)) { assert(0 && Py_STRINGIFY(expr)); } } while (0)
    
```

multidict/_multilib/hashtable.h
```diff
@@ -1929,7 +1925,6 @@ _md_check_consistency(MultiDictObject *md, bool update)
         Py_ssize_t ix = htkeys_get_index(keys, i);
         CHECK(DKIX_DUMMY <= ix && ix <= calc_usable);
     }
```
```suggestion
    }
    
```

multidict/_multilib/hashtable.h
```diff
@@ -1995,6 +1990,7 @@ _md_dump(MultiDictObject *md)
     printf("\n");
     return 1;
 }
+
```
```suggestion
```

multidict/_multidict.c
```diff
@@ -674,7 +674,7 @@ multidict_popone(MultiDictObject *self, PyObject *const *args,
...
-
+
```
asvetlov:
```suggestion

```

multidict/_multilib/hashtable.h
```diff
@@ -1877,21 +1877,20 @@ md_traverse(MultiDictObject *md, visitproc visit, void *arg)
 static inline int
 md_clear(MultiDictObject *md)
 {
-    if (md->used == 0) {
+    if ((md->keys == &empty_htkeys) || (md->state == NULL)) {
         return 0;
     }
     md->version = NEXT_VERSION(md->state);
-
+
```
```suggestion

```

multidict/_multilib/hashtable.h
```diff
@@ -1877,21 +1877,20 @@ md_traverse(MultiDictObject *md, visitproc visit, void *arg)
...
-            Py_CLEAR(entry->key);
-            Py_CLEAR(entry->value);
-        }
+        // Py_CLEAR has null checks of it's own making it easier to free.
+        Py_CLEAR(entry->identity);
+        Py_CLEAR(entry->key);
+        Py_CLEAR(entry->value);
     }
-
+
```
```suggestion

```

multidict/_multilib/hashtable.h
```diff
@@ -1910,7 +1909,7 @@ _md_check_consistency(MultiDictObject *md, bool update)
 
 #define CHECK(expr) assert(expr)
     //    do { if (!(expr)) { assert(0 && Py_STRINGIFY(expr)); } } while (0)
-
+
```
```suggestion

```

multidict/_multilib/hashtable.h
```diff
@@ -1929,7 +1928,7 @@ _md_check_consistency(MultiDictObject *md, bool update)
         Py_ssize_t ix = htkeys_get_index(keys, i);
         CHECK(DKIX_DUMMY <= ix && ix <= calc_usable);
     }
-
+
```
```suggestion

```

multidict/_multilib/hashtable.h
```diff
@@ -1877,21 +1877,18 @@ md_traverse(MultiDictObject *md, visitproc visit, void *arg)
...
-
     entry_t *entries = htkeys_entries(md->keys);
     for (Py_ssize_t pos = 0; pos < md->keys->nentries; pos++) {
         entry_t *entry = entries + pos;
-        if (entry->identity != NULL) {
-            Py_CLEAR(entry->identity);
-            Py_CLEAR(entry->key);
-            Py_CLEAR(entry->value);
-        }
+        // Py_CLEAR has null checks of it's own making it easier to free.
```
asvetlov: Functionally it is the same, but an explicit check requires a single check for skipping already deleted slots.
If identity is not null, `Py_CLEAR` also runs fast because the deleted reference is already in the processor's registry. Even without a help of compiler's optimizer the double null check in this case is super fast. Let's keep the current design.

multidict/_multilib/hashtable.h
```diff
@@ -1877,21 +1877,18 @@ md_traverse(MultiDictObject *md, visitproc visit, void *arg)
...
+        Py_CLEAR(entry->identity);
+        Py_CLEAR(entry->key);
+        Py_CLEAR(entry->value);
```
```suggestion
        if (entry->identity != NULL) {
            Py_CLEAR(entry->identity);
            Py_CLEAR(entry->key);
            Py_CLEAR(entry->value);
        }
```

multidict/_multidict.c
asvetlov:
```suggestion
    }

```

multidict/_multilib/hashtable.h
asvetlov, line 1883:
```suggestion
    md->version = NEXT_VERSION(md->state);

```

multidict/_multilib/hashtable.h
```suggestion
    }

```

multidict/_multilib/hashtable.h
```suggestion
    //    do { if (!(expr)) { assert(0 && Py_STRINGIFY(expr)); } } while (0)

```

multidict/_multilib/hashtable.h
```suggestion
    }

```

multidict/_multilib/hashtable.h
```diff
@@ -1877,7 +1877,7 @@ md_traverse(MultiDictObject *md, visitproc visit, void *arg)
 static inline int
 md_clear(MultiDictObject *md)
 {
-    if (md->used == 0) {
+    if ((md->keys == &empty_htkeys) || (md->state == NULL)) {
```
```suggestion
    if (md->keys == &empty_htkeys) {
```
`md->state` is never assigned to NULL by the code

multidict/_multilib/hashtable.h
```diff
@@ -1877,7 +1877,7 @@ md_traverse(MultiDictObject *md, visitproc visit, void *arg)
 static inline int
 md_clear(MultiDictObject *md)
 {
-    if (md->used == 0) {
+    if (md->keys == &empty_htkeys) {
```
```suggestion
    if (md->keys == NULL || md->keys == &empty_htkeys) {
```
`md->state` is never assigned to NULL by the code

PR #1235, opened 2025-08-21
add Python 3.14 CI

kumaraditya303:
## Checklist

- [ ] I think the code is well written

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -174,8 +174,8 @@ jobs:
     strategy:
       matrix:
         pyver:
-        - 3.14t-dev
-        - 3.14-dev
+        - 3.14t
+        - 3.14
         - 3.13t
         - 3.13
         - 3.12
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -174,7 +174,9 @@ jobs:
     strategy:
       matrix:
         pyver:
-        - 3.13-freethreading
+        - 3.14t-dev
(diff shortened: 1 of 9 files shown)


Vizonex (approved):
I thought something was altered from the tests I've added but this doesn't seem to be the case, glad this has now been added in this should be good for stress-testing the 3.14 update.

Vizonex:
> > I did not notice this earlier but it seems some tests are failing for some reason let me go in and see if I can suggest any changes. The others are on a tighter schedule than me apparently.
> 
> The failing tests are leak tests, it is likely that it is because of gc changes in 3.14 or just higher memory usage on 3.14 interpreter though I haven't looked at it closely.

Makes sense. I mainly implemented the newer memory-leak test to prevent newer memory leak issues in the future since there have been several incidents in MultiDict's history of this happening repetitively. 

I am surprised it's not related to that one and only istr. When `tests/test_istr.py::test_leak[py]` is ran in isolation it passes which I find very strange.

kumaraditya303:
> Great job, I'll be sure to contact one of the contributors over on matrix If I don't see any activity here in the next couple days.


Thanks I have updated the tests for CPython 3.14, specifically these two changes:
- `tests/test_istr.py`: Changed it to check for `case_insensitive_str_class` instances rather than all objects which isn't stable.
- `tests/isolated/multidict_pop.py`: Change it to check for relative increase in memory rather than absolute memery usage.

bdraco (approved):
Thanks @kumaraditya303

PR #1252, opened 2025-10-05
Increase wheel build timeout to 25m

bdraco:
We build so many versions and the free-threading as well now so the release failed


Vizonex:
@bdraco Shall I help with anything?

bdraco:
Not unless you can make the CI run faster :). Seems to be quite the backlog right now

webknjaz:
@Vizonex no, this setting is unrelated to how GHA's job scheduler operates. I bet the problem is a big backlog. Usually, "heavy" runners (macos/win) may experience more delays picking up the jobs, among other things. Also, rate-limiting considers entire org resources as shared — this means that one repository can DDoS the CI for the org. Oftentimes, a lot of this can be fixed by setting up auto-cancelation in every workflow in the org.

PR #1284, opened 2026-01-22
Fix memory leak caused by identity when default value is inplace

### LINKED ISSUE #1273: Memory Leak on 6.6.4 CIMultiDict pop
After upgrading to multidict `6.6.4` and running it for a few months, we are still observing continuous memory allocation that is not being released. We enabled tracemalloc, and the `.pop()` function is still the culprit, even though this issue was marked as resolved: https://github.com/aio-libs/multidict/issues/1232.  Tagging @Vizonex for ideas.

Vizonex:
## What do these changes do?

This was reported a while ago without a reproducer to use until @rodrigobnogueira provided me some very detailed information about where the problem was and how to fix it. So I went ahead and added in something to ensure that the identity value is properly handled if or when the user has provided a default value to use.

A Slow memory leak should be patched now.

## Related issue number

fixes #1273

## Checklist

### DIFF
--- a/CHANGES/1284.bugfix.rst
+++ b/CHANGES/1284.bugfix.rst
@@ -1,2 +1,2 @@
-Fixed slow memory leak caused by identity by adding ``Py_DECREF`` to identity value before leaving ``md_pop_one`` on success.
--- by :user:`Vizonex`
+Fixed slow memory leak caused by identity by adding ``Py_DECREF`` to identity value before leaving ``md_pop_one`` on success
+-- by :user:`Vizonex`.
--- a/CHANGES/1284.bugfix.rst
+++ b/CHANGES/1284.bugfix.rst
@@ -1,0 +1,2 @@
+Fixed slow memory leak caused by identity by adding ``Py_DECREF`` to identity value before leaving ``md_pop_one`` on success.
+-- by :user:`Vizonex`
--- a/tests/isolated/multidict_pop.py
+++ b/tests/isolated/multidict_pop.py
@@ -65,7 +65,9 @@ def _test_popone() -> None:
 
 # SEE: https://github.com/aio-libs/multidict/issues/1273
 def _test_pop_with_default() -> None:
-    result = MultiDict()
+    # XXX: mypy wants an annotation so the only
+    # thing we can do here is pass the headers along.
+    result = MultiDict(headers)
     for i in range(100_000):
         result.pop(f"missing_key_{i}", None)
     check_for_leak()
--- a/tests/isolated/multidict_pop.py
+++ b/tests/isolated/multidict_pop.py
@@ -61,6 +61,14 @@ def _test_popone() -> None:
                 result.popone(k)
         check_for_leak()
 
+# SEE: https://github.com/aio-libs/multidict/issues/1273
+def _test_pop_with_default() -> None:
+    for _ in range(10):
+        for _ in range(100):
+            result = MultiDict(headers)
+            for k in keys:
+                result.pop(k, None)
+        check_for_leak()
 
 def _test_del() -> None:
     for _ in range(10):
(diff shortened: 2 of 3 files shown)


CHANGES/1284.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+Fixed slow memory leak caused by identity by adding ``Py_DECREF`` to identity value before leaving ``md_pop_one`` on success.
+-- by :user:`Vizonex`
```
webknjaz:
```suggestion
Fixed slow memory leak caused by identity by adding ``Py_DECREF`` to identity value before leaving ``md_pop_one`` on success
-- by :user:`Vizonex`.
```

rodrigobnogueira (approved):
The fix looks good. I'd add a test for it. thanks

bdraco (approved):
Fix looks correct. Note there are still other review comments

bdraco:
Thanks @Vizonex

bdraco (approved):

tests/isolated/multidict_pop.py
```diff
@@ -63,6 +63,16 @@ def _test_popone() -> None:
         check_for_leak()
 
 
+# SEE: https://github.com/aio-libs/multidict/issues/1273
+def _test_pop_with_default() -> None:
+    for _ in range(10):
+        for _ in range(100):
+            result = MultiDict(headers)
+            for k in keys:
+                result.pop(k, None)
```
rodrigobnogueira: to properly reproduce the leak, just pop a missing key

```python
    for i in range(100_000):
        d.pop(f"missing_key_{i}", None)
```

tests/isolated/multidict_pop.py
```diff
@@ -63,6 +63,16 @@ def _test_popone() -> None:
         check_for_leak()
 
 
+# SEE: https://github.com/aio-libs/multidict/issues/1273
+def _test_pop_with_default() -> None:
+    # XXX: mypy wants an annotation so the only
+    # thing we can do here is pass the headers along.
+    result = MultiDict(headers)
+    for i in range(100_000):
```
rodrigobnogueira: @Vizonex , apologies, I previously suggested 100_000 iterations, but I've verified that this only generates a leak of ~6MB. Since `check_for_leak()` has a 50MB threshold, the test currently passes even with the leak present.

For this test suite to effectively catch the regression, checking for leaks with 1_000_000 iterations is necessary (which results in > 50MB leaked). 
Maybe the 50MB threshold is too high for these isolated checks.

Your test with 1M iterations and the bug present in hashtable.h outputs:

```
rodrigo@rbn multidict % PYTHONPATH=. .venv/bin/python tests/isolated/multidict_pop.py
  File "/Users/rodrigo/Proj/wvscode/multidict/tests/isolated/multidict_pop.py", line 91, in <module>
    _run_isolated_case()
    ~~~~~~~~~~~~~~~~~~^^
  File "/Users/rodrigo/Proj/wvscode/multidict/tests/isolated/multidict_pop.py", line 87, in _run_isolated_case
    _test_pop_with_default()
    ~~~~~~~~~~~~~~~~~~~~~~^^
  File "/Users/rodrigo/Proj/wvscode/multidict/tests/isolated/multidict_pop.py", line 71, in _test_pop_with_default
    check_for_leak()
    ~~~~~~~~~~~~~~^^
  File "/Users/rodrigo/Proj/wvscode/multidict/tests/isolated/multidict_pop.py", line 35, in check_for_leak
    assert usage < 50, f"Memory leaked at: {usage} MB"
           ^^^^^^^^^^
AssertionError: Memory leaked at: 61.3125 MB
```

rodrigobnogueira (changes requested):

rodrigobnogueira (approved):

bdraco (approved):
Thanks @Vizonex

PR #1285, opened 2026-01-22
Fix CI mypy errors

bdraco:
## What do these changes do?

Fix CI mypy errors

no

### DIFF
--- a/.mypy.ini
+++ b/.mypy.ini
@@ -27,5 +27,8 @@ warn_return_any = True
 warn_unreachable = True
 warn_unused_ignores = True
 
+[mypy-sphinx.*]
+follow_untyped_imports = False
+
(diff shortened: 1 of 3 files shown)


.mypy.ini
```diff
@@ -27,5 +27,8 @@ warn_return_any = True
 warn_unreachable = True
 warn_unused_ignores = True
 
+[mypy-sphinx.*]
+follow_untyped_imports = False
```
webknjaz, line 31: Is this necessary?

bdraco: Probably not after `Sphinx >= 5.3.0, < 8.0.0`

Will check and remove in followup if its truely unneeded

bdraco: https://github.com/aio-libs/multidict/pull/1286

PR #1288, opened 2026-01-25
Feat: add the to_dict()

### LINKED ISSUE #783: Enhancement: `to_dict` method
Hi there,

Thanks for this library. I'd like to ask for a convenience method that output a dict of a multidict - with lists for multi-values. Currently I have this:

```python
    def dict(self) -> Dict[str, List[Any]]:
        """

        Returns:
            A dict of lists
        """
        return {k: self.getall(k) for k in set(self.keys)}
```

While this works, its not very performant.

rodrigobnogueira:
## What do these changes do?

Implement `to_dict()` methods for `MultiDict`, `CIMultiDict`, `MultiDictProxy`, and `CIMultiDictProxy`. This method groups values with the same key into a list.

### Implementation Details

- **C Extension**: Added `multidict_to_dict` in `_multidict.c` with direct `PyDict` and `PyList` manipulation for performance.
- **Python Fallback**: Added generic `to_dict` in `_multidict_py.py`.
- **Tests**: Added comprehensive test coverage in `tests/test_to_dict.py`, including verification of order preservation, case-insensitivity, mixed types, and proxy mutation isolation.
- **Memory Safety**: Added a new isolated leak check (`tests/isolated/multidict_to_dict.py`) and integrated it into the CI suite (`tests/test_leaks.py`). Validated with intentional leak inserted in code (removed after test).

### Example

```python
md = MultiDict([("a", 1), ("b", 2), ("a", 3)])
md.to_dict()

# Result: {'a': [1, 3], 'b': [2]}
```

No existing behavior changes. This adds a new method `to_dict()` 

## Related issue number

Fixes #783 (Add `to_dict` method)

## Checklist


Vizonex (commented):
I would've applied some smarter logic like traverse all keys and gathering all the values up with something like `getall(...)` when going over `keys` since it would be a bit faster due to less time needing to check weather or not something is a list. Otherwise I think this is a good start in the right direction. There might be a quicker method though that you can try which I've attempted to visualize for you although I haven't benchmarked it but just incase here was my idea. 

It's advantages mainly revolve around not needing to check on weather or not a key has already been used.

```python
# incase you need a visual of the logic I'm trying to explain
def to_dict(md: MutliDict[str]):
    return {k: md.getall(k) for k in md.keys()}
```

Vizonex (approved):
Converter looks good. I have no complaints with this one. Great job with the pytest module also.

CHANGES/783.feature
```diff
@@ -0,0 +1 @@
+Added ``to_dict()`` method to export multidict contents as a standard dict with value lists. (#783)
```
webknjaz: Don't duplicate the issue number in the text — Towncrier adds it from the file name.

tests/test_to_dict.py
```diff
@@ -0,0 +1,180 @@
...
+        """Test basic conversion with unique keys."""
+        d = cls([("a", 1), ("b", 2)])
+        result = d.to_dict()
+        assert result == {"a": [1], "b": [2]}
+
+    def test_to_dict_multi_values(self, cls: DictFactory) -> None:
+        """Test grouping multiple values under the same key."""
+        d = cls([("a", 1), ("b", 2), ("a", 3)])
+        result = d.to_dict()
+        assert result == {"a": [1, 3], "b": [2]}
```
webknjaz: Looks like these are the same test w/ different params. They can be collapsed.

tests/test_to_dict.py
```diff
@@ -0,0 +1,180 @@
...
+
+
+class DictFactory(Protocol):
+    def __call__(
+        self, arg: Optional[Iterable[tuple[str, object]]] = None
+    ) -> MultiMapping[object]:
+        raise NotImplementedError  # pragma: no cover
+
+
+class BaseToDictTests:
```
webknjaz: Do we really need classes here? This adds extra indentation and unused args. Plus the `cls` fixture isn't self-explanatory with a generic name. Instead, we could rely on the fixtures like `any_multidict_class` and many more that are defined in `conftest.py` globally, and integrated with a CLI based mechanism for opting in or out into running testing against the C-extensions or skipping them.

webknjaz:
@rodrigobnogueira note that there's no need to use conventional commit prefixes as they only eat up space. Towncrier uses change note-based identifiers in file names.

Dreamsorcerer:
@rodrigobnogueira Do you think anything here could be pulled into https://github.com/aio-libs/aiohttp/pull/7679?

Dreamsorcerer:
> @rodrigobnogueira Do you think anything here could be pulled into [aio-libs/aiohttp#7679](https://github.com/aio-libs/aiohttp/pull/7679)?

I'm guessing probably not, as it produces a list, rather than a concatenated string. I wonder how useful this feature will actually be (i.e. a user could just use a dict of lists in the first place if that's what they want...).

rodrigobnogueira:
> > @rodrigobnogueira Do you think anything here could be pulled into [aio-libs/aiohttp#7679](https://github.com/aio-libs/aiohttp/pull/7679)?
> 
> I'm guessing probably not, as it produces a list, rather than a concatenated string. I wonder how useful this feature will actually be (i.e. a user could just use a dict of lists in the first place if that's what they want...).

I opened this PR a few months ago to address the 2022 issue requesting this feature. If this isn't the direction we want to take, I’m fine with closing it alongside #783 to avoid adding unnecessary complexity to the library.

The value of this feature is probably a more performant `to_dict` method, that doesn't use the `self.getall` described  in the issue.

Dreamsorcerer:
> If this isn't the direction we want to take, I’m fine with closing it alongside #783 to avoid adding unnecessary complexity to the library.

I don't have a strong opinion, though it'd be nice to know if this is a somewhat common usecase for people. I'll let someone else decide if this is a good idea or not.

Vizonex:
> > If this isn't the direction we want to take, I’m fine with closing it alongside #783 to avoid adding unnecessary complexity to the library.
> 
> I don't have a strong opinion, though it'd be nice to know if this is a somewhat common usecase for people. I'll let someone else decide if this is a good idea or not.

@Dreamsorcerer I think it would be good for libraries such as [msgspec](https://github.com/msgspec/msgspec) to have this feature implemented. I could see some use-cases such as sterilizing header files for debugging an http request for instance or for other configuration systems.

rodrigobnogueira:
> > > If this isn't the direction we want to take, I’m fine with closing it alongside #783 to avoid adding unnecessary complexity to the library.
> > 
> > 
> > I don't have a strong opinion, though it'd be nice to know if this is a somewhat common usecase for people. I'll let someone else decide if this is a good idea or not.
> 
> @Dreamsorcerer I think it would be good for libraries such as [msgspec](https://github.com/msgspec/msgspec) to have this feature implemented. I could see some use-cases such as sterilizing header files for debugging an http request for instance or for other configuration systems.

Use-case evidence, since that was the open question: this mirrors an established convenience on multidict-shaped types. Werkzeug's MultiDict has exactly this as to_dict(flat=False), and Django's QueryDict exposes it through .lists().
The request in #783 came from the litestar author, and the typical situation is not one where the user could have started with a dict of lists: the multidict arrives from a library, parsed headers or query strings, and needs exporting for JSON, test comparison, or msgspec/pydantic input, as @Vizonex mentioned. The C implementation also beats the naive getall() comprehension by a wide margin (benchmarks above).

Meanwhile all review notes are addressed, to_dict() is now documented in the API reference, and the full suite is green on 3.12 and 3.13t.

multidict/_abc.py
```diff
@@ -42,6 +42,24 @@ def getone(self, key: str, default: _T) -> _V_co | _T: ...
...
+        equal under a normalisation the iteration does not apply, such as
+        case-insensitive mappings, must override this; the default would
+        emit one entry per spelling.
+        """
+        result: dict[str, list[_V_co]] = {}
+        for key in self:
+            # Iteration yields a duplicated key once per occurrence, so the
+            # membership test is what keeps getall() from being applied twice.
+            if key not in result:
+                result[key] = list(self.getall(key))
```
asvetlov, line 60:
```suggestion
                result[key] = self.getall(key)
```
`getall()` already returns a new list instance, `list(lst)` call is redundant

multidict/_multidict.c
```diff
@@ -256,6 +256,81 @@ _multidict_proxy_copy(MultiDictProxyObject* self, PyTypeObject* type)
...
+
+static PyObject*
+multidict_to_dict(MultiDictObject* self)
+{
+    PyObject* result = PyDict_New();
+    if (result == NULL) {
+        return NULL;
+    }
+
+    PyObject* seen = PyDict_New();
```
asvetlov, line 270: `seen` extra list is a red flag; the algorithm is probably sub-optimal.

Perhaps the following should work better:
1. iterate over all entries
2. for each entry collect all values using approach borrowed from getall() implementation but don't clean 'visited' mark after collecting key's ivalues.
3. if the next entry is already visited -- skip it
4. at the end, iterate over all entries again and clear 'visited' mark

multidict/_multidict.c
```diff
@@ -256,6 +256,81 @@ _multidict_proxy_copy(MultiDictProxyObject* self, PyTypeObject* type)
...
+    PyObject* value = NULL;
+
+    int tmp;
+    while ((tmp = md_next(self, &pos, &identity, &key, &value)) > 0) {
+        /* `seen` maps identity to the very list held in `result`, not to the
+           first key.  Looking the list up again in `result` would have to
+           hash a key that may be a str subclass with a user-defined
+           __hash__, and PyDict_GetItem() swallows the exception from that,
+           so a miss would return NULL with nothing set. */
+        PyObject* lst = PyDict_GetItem(seen, identity);
```
asvetlov, line 289: Safe `PyDict_GetItemRef` should be used instead

docs/multidict.rst
```diff
@@ -71,6 +71,41 @@ MultiDict
...
+      ``{k: d.getall(k) for k in d}`` comprehension gets wrong: it emits a
+      separate entry for every spelling, each holding the full list::
+
+         >>> ci = CIMultiDict([('A', '1'), ('a', '2')])
+         >>> ci.to_dict()
+         {'A': ['1', '2']}
+         >>> {k: ci.getall(k) for k in ci}
+         {'A': ['1', '2'], 'a': ['1', '2']}
+
+      .. versionadded:: 6.8
```
asvetlov:
```suggestion
      .. versionadded:: 6.9
```

docs/multidict.rst
```diff
@@ -301,6 +336,13 @@ MultiDictProxy
 
       Return a shallow copy of the underlying multidict.
 
+   .. method:: to_dict()
+
+      Return a :class:`dict` mapping each key to a list of all its
+      values, preserving insertion order.
+
+      .. versionadded:: 6.8
```
asvetlov:
```suggestion
      .. versionadded:: 6.9
```

asvetlov (approved):
lgtm

asvetlov:
Great job! 
Thanks, @rodrigobnogueira

PR #1292, opened 2026-02-02
Use SPDX license id

vargenau:
Use SPDX license id so that tools can easily generate valid SPDX SBOMs.


PaulGarnerUk:
Apologies, but we're also having the same issue with our tooling constantly flagging this library as having a non-compliant license because it can't identify via SPDX

I'm also struggling to find any discussion about this. Would anyone care to provide the link please?  Maybe @webknjaz ?

PR #1296, opened 2026-02-10
Fix missing space in ValueError message

veeceey:
## Summary

Fix missing space in the `ValueError` message raised by `_parse_args()` -- the implicit string concatenation between the two f-strings was missing a trailing space, producing `"element #0has"` instead of `"element #0 has"`.

## Test plan

- Verified the corrected message output manually

### DIFF
--- a/CHANGES/1295.bugfix.rst
+++ b/CHANGES/1295.bugfix.rst
@@ -1,0 +1,3 @@
+Fixed a missing space in the :exc:`ValueError` message
+raised when constructing a multidict from a sequence
+-- by :user:`veeceey`.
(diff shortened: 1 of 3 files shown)


webknjaz:
Alright.. Please add a regression test.

webknjaz (changes requested):
No need to duplicate an existing test. Just update it.

Vizonex (approved):
Looks good no complaints from me.

veeceey:
Done -- updated the existing `test_cannot_create_from_unaccepted` to check the full message spacing instead of adding a separate test.

veeceey:
Hi @webknjaz, just following up — I updated the existing test as you suggested instead of adding a separate one. Would you mind taking another look when you get a chance? Thanks\!

veeceey:
Hey @webknjaz, I've updated the existing test as you suggested instead of duplicating it. Could you take another look when you get a chance? Thanks\!

webknjaz (approved):

PR #1304, opened 2026-03-22
Add __reversed__ support to keys, values, and items views

### LINKED ISSUE #448: Make keys view reversible
Since Python 3.8 keys view is reversible (but `Mapping`/`MutableMapping` ABCs are not for the sake of backward compatibility).

It would be nice to support reversibility in multidict keys views as well.

r266-tech:
## Summary

Implements the `__reversed__` method for `_KeysView`, `_ValuesView`, and `_ItemsView`, allowing `reversed()` iteration over multidict views.

Since Python 3.8, built-in `dict` views support `reversed()`. This PR brings multidict views to parity.

Closes #448.

## Changes

- Added `iter_entries_reverse()` to `_HtKeys` for reverse entry traversal
- Added `__reversed__` and `_iter_reversed` to `_KeysView`, `_ValuesView`, and `_ItemsView`
- All view types (including CIMultiDict and proxies) support `reversed()`

## Example

```python

md = MultiDict([('a', 1), ('b', 2), ('c', 3)])
list(reversed(md.keys()))   # ['c', 'b', 'a']
list(reversed(md.values())) # [3, 2, 1]
list(reversed(md.items()))  # [('c', 3), ('b', 2), ('a', 1)]
```

### DIFF
--- a/CHANGES/448.feature.rst
+++ b/CHANGES/448.feature.rst
@@ -1,0 +1,1 @@
+Added support for reversing keys, values, and items views via ``reversed()`` -- implements :issue:`448`.
--- a/multidict/_multidict_py.py
+++ b/multidict/_multidict_py.py
@@ -101,6 +101,15 @@ class _ItemsView(_ViewBase[_V], ItemsView[str, _V]):
                 raise RuntimeError("Dictionary changed during iteration")
             yield self._md._key(e.key), e.value
 
+    def __reversed__(self) -> Iterator[tuple[str, _V]]:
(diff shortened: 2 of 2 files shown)


Vizonex (commented):
I would recommend you add a C Implementation and not just a Python One. This is because most users will end up with compiled binaries and multidict will default to the C Module if found.

r266-tech:
Thanks for the review, @Vizonex — agreed that a Python-only implementation falls short given the C module is the default for most installs. #1340 now covers both the C extension and the pure-Python fallback and has been approved, so closing this one in favor of that PR.

Issue #1306, opened 2026-03-24
multidict C Extension Analysis Report


devdanzin:
I ran a [code analysis tool](https://github.com/devdanzin/cext-review-toolkit) I'm developing on multidict and it generated an interesting report: https://gist.github.com/devdanzin/a36588ae7e2b73f0d85f8a925fb13b3d. It contains issue descriptions, suggested fixes, and reproducers for some of the issues.

From the summary:

> The most critical finding is a **double-free/use-after-free in `md_contains`** and a **crash from swapped format string arguments** in `_err_cannot_fetch`. The iterator and view types are also missing the heap type lifecycle management (`Py_VISIT(Py_TYPE(self))`) that the main dict types correctly implement.

I hope it helps.

asvetlov:
All issues are addressed

PR #1310, opened 2026-03-28
fix format error that leads to segfaults

Vizonex:
## What do these changes do?

This was based off a user-accessible segfault found by @devdanzin. Not sure how this appeared out of nowhere but my best assumption it could've been caused by developer fatigue or burn out. The reproducible seen in this gist here https://gist.github.com/devdanzin/a36588ae7e2b73f0d85f8a925fb13b3d is not just limited to 3.14 but also 3.10 (which I have now recently tested myself) so I've gone ahead and fixed it. Feel free to let me know if this should somehow be added to pytest although IDK where to put it. Originally was going to write this off as an ffmpeg fiasco but this does not seem to be the case. Seems we might have couple of other sneaky bugs to patch.

Reproducer from the gist brought here for reference
```python
"""Reproducer: multidict _err_cannot_fetch swapped format args — abort.

hashtable.h:1467-1475 passes (name, i) but format expects (i, name).
  %zd receives a char* → prints garbage number
  %s receives an int → dereferences integer as pointer → crash

No special setup required.
Tested: multidict 6.7.1, Python 3.14.
"""

class BadItem:
    """Looks like a 2-element sequence but __getitem__ raises."""
    def __len__(self):
        return 2
    def __getitem__(self, i):
        raise RuntimeError("intentional getitem failure")

MultiDict([BadItem()])
# Expected: ValueError with proper error message
# Actual: Aborted (core dumped)
#   Assertion `!_PyErr_Occurred(tstate)' failed. 
```

#1306 

## Checklist


bdraco:
Fix looks good. Can you add a test?

CHANGES/1310.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+fix segfault that can be triggered when getting data from getting an item.
+-- by :user:`Vizonex`
```
webknjaz:
```suggestion
A segmentation fault that could be triggered when getting an item is now fixed
-- by :user:`Vizonex`.
```

tests/isolated/multidict_bad_item.py
```diff
@@ -0,0 +1,18 @@
+from multidict import MultiDict
```
bdraco: We don't need to isolate it now that the bug is fixed

multidict/_multidict_py.py
```diff
@@ -834,8 +834,16 @@ def _parse_args(
...
-                    identity = identity_func(item[0])
-                    yield _Entry(hash(identity), identity, item[0], item[1])
+                    try:
+                        key = item[0]
+                        value = item[1]
+                    except Exception as exc:
+                        raise ValueError(
+                            f"multidict update sequence element #{pos}'s "
+                            f"key could not be fetched"
+                        ) from exc
```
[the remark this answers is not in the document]
bdraco: kinda meh on that. should match the c version

bdraco:
stall in test_cannot_create_from_item_with_failing_getitem[c-cls0]

bdraco (approved):

PR #1316, opened 2026-04-11
Drop Python 3.9 support

bdraco:
## What do these changes do?

Drops Python 3.9 from the CI matrix, classifiers, `python_requires`, and the ruff target; bumps the PyPy matrix entry to pypy-3.10. Python 3.9 hit end of life in October 2025, so this removes a slot that no longer gets upstream fixes and trims the CI matrix (which has been timing out).

Users on Python 3.9 will no longer receive new releases; minimum supported version is now Python 3.10.

## Checklist

### DIFF
--- a/multidict/_multidict_py.py
+++ b/multidict/_multidict_py.py
@@ -481,7 +481,7 @@ class _HtKeys(Generic[_V]):  # type: ignore[misc]
     log2_size: int
     usable: int
 
-    indices: array  # type: ignore[type-arg] # in py3.9 array is not generic
+    indices: array[int]
     entries: list[Optional[_Entry[_V]]]
 
     @functools.cached_property
--- a/multidict/_multidict_py.py
+++ b/multidict/_multidict_py.py
@@ -468,7 +468,7 @@ class _Entry(Generic[_V]):
 
 
 @dataclass
-class _HtKeys(Generic[_V]):
+class _HtKeys(Generic[_V]):  # type: ignore[misc]
     LOG_MINSIZE: ClassVar[int] = 3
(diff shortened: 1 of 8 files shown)


Vizonex:
I'm for removing 3.9 if that keeps production and maintenance stable. Perhaps I should consider moving my current version of python from 3.10 to 3.11 and drop 3.9 on some of my own libraries that still support it.

multidict/_multilib/pythoncapi_compat.h
Dreamsorcerer, line 1: I've added some cleanup, if someone can check that.
I'm not clear what should be changed in tools/upgrade_pythoncapi.py. Just remove the classes that reference Python 3.9? Or only remove the NEED_PYTHONCAPI_COMPAT bit?

bdraco: I think we download the new https://github.com/python/pythoncapi-compat/blob/main/upgrade_pythoncapi.py, set it to 3, 10 and re-run

bdraco: It put some of it back, but its upstream code so its fine. It compiles to nothing anyways

multidict/_multidict_py.py
```diff
@@ -481,7 +481,7 @@ class _HtKeys(Generic[_V]):
     log2_size: int
     usable: int
 
-    indices: array[int]
+    indices: array  # type: ignore[type-arg] # in py3.9 array is not generic
```
Dreamsorcerer: Ah, couldn't see anything in the docs saying which version. Need to atleast change the comment then:
`# TODO(PY312): array[int]`

PR #1314, opened 2026-04-03
fix coverage of isolated tests

Vizonex:
## What do these changes do?
Dreamsorcerer has brought up with me a pretty interesting idea with me on matrix to help us fix code coverage with the isolated tests which I have now chosen to take his advice on.

This PR is meant to assist with fixing code coverage of isolated tests being added. This has been a recent problem with #1310 & #1311 as they fail to pass due to missing it's code coverage. 

## Are there changes in behavior for the user?

Maintainers working on multidict weather it's patching bugs or adding new features should have a less stressful time trying to put together bug scripts for the C Extension of multidict now if this PR passes through.

## Checklist

- [x] I think the code is well written
- [x] Unit tests for the changes exist
- [x] Documentation reflects the changes

PR #1317, opened 2026-04-12
Add free-threading race tests for the C extension

rodrigobnogueira:
## What do these changes do?

Adds two race tests for the C extension on free-threaded builds: one iterates a multidict while other threads mutate it, the other extends from a source that other threads are resizing.

#1431 through #1446 made these paths safe. Against the tree before that series both tests segfault on every run; against current master they pass. There is no other threaded test in the suite, so this is what would turn a regression into a red free-threaded CI leg instead of silence.

The C changes this PR originally carried are superseded by that series and have been dropped.

No.

No. One test file and a `contrib` fragment; the tests are `c_extension`-marked and take about a quarter of a second.

## Related issue number

Follows #1431, #1433, #1438, #1441, #1445, #1446.

## Checklist

- [ ] Documentation reflects the changes -- N/A
- [ ] If you provide code modification, please add yourself to `CONTRIBUTORS.txt` -- N/A

### DIFF
--- a/CHANGES/1317.bugfix.rst
+++ b/CHANGES/1317.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed a memory-safety race condition resulting in segmentation faults (Use-After-Free) when iterating and modifying a ``MultiDict`` concurrently in CPython free-threaded mode (3.13t+). Read/Write accesses to the internal ``keys`` buffer are now wrapped in ``Py_BEGIN_CRITICAL_SECTION``.
--- a/CHANGES/1317.bugfix.rst
+++ b/CHANGES/1317.bugfix.rst
@@ -1,1 +1,2 @@
-Fixed a memory-safety race condition resulting in segmentation faults (Use-After-Free) when iterating and modifying a ``MultiDict`` concurrently in CPython free-threaded mode (3.13t+). Read/Write accesses to the internal ``keys`` buffer are now wrapped in ``Py_BEGIN_CRITICAL_SECTION``.
+Fixed a memory-safety race condition resulting in segmentation faults (Use-After-Free) when iterating and modifying a ``MultiDict`` concurrently in CPython free-threaded mode (3.13t+). Read/Write accesses to the internal ``keys`` buffer are now wrapped in ``Py_BEGIN_CRITICAL_SECTION``
+-- by :user:`rodrigobnogueira`.
--- a/CHANGES/1317.bugfix.rst
+++ b/CHANGES/1317.bugfix.rst
@@ -1,2 +1,2 @@
-Fixed a memory-safety race condition resulting in segmentation faults (Use-After-Free) when iterating and modifying a ``MultiDict`` concurrently in CPython free-threaded mode (3.13t+). Read/Write accesses to the internal ``keys`` buffer are now wrapped in ``Py_BEGIN_CRITICAL_SECTION``
+Fixed a memory-safety race condition resulting in segmentation faults (Use-After-Free) when iterating and modifying a :class:`~multidict.MultiDict` concurrently in CPython free-threaded mode (3.13t+). Read/Write accesses to the internal ``keys`` buffer are now wrapped in ``Py_BEGIN_CRITICAL_SECTION``
 -- by :user:`rodrigobnogueira`.
--- a/CHANGES/1317.bugfix.rst
+++ b/CHANGES/1317.bugfix.rst
@@ -1,2 +1,6 @@
-Fixed a memory-safety race condition resulting in segmentation faults (Use-After-Free) when iterating and modifying a :class:`~multidict.MultiDict` concurrently in CPython free-threaded mode (3.13t+). Read/Write accesses to the internal ``keys`` buffer are now wrapped in ``Py_BEGIN_CRITICAL_SECTION``
+Fixed a memory-safety race condition resulting in segmentation faults
+(use-after-free) when iterating and modifying a :class:`~multidict.MultiDict`
+concurrently in CPython free-threaded mode (3.13t+). Read and write accesses
+to the internal ``md->keys`` buffer are now wrapped in
+:c:macro:`Py_BEGIN_CRITICAL_SECTION`
 -- by :user:`rodrigobnogueira`.
--- a/tests/test_free_threading.py
+++ b/tests/test_free_threading.py
@@ -99,4 +99,3 @@ def test_pure_python_free_threaded_warning() -> None:
         text=True,
     )
     assert result.returncode == 0, result.stderr
-
--- a/tests/test_free_threading.py
+++ b/tests/test_free_threading.py
@@ -33,7 +33,7 @@ def test_race_condition_iterator_vs_mutation(
         for i in range(256):
             try:
                 target[f"k-{i % 64}"] = f"v{i}"
-            except RuntimeError:
+            except RuntimeError:  # pragma: no cover
                 pass  # "MultiDict is changed during iteration" is expected
             except Exception as e:  # pragma: no cover
                 import traceback
--- a/tests/test_free_threading.py
+++ b/tests/test_free_threading.py
@@ -13,11 +13,11 @@ def test_race_condition_iterator_vs_mutation(
     """Test that concurrent iterations and mutations do not cause a memory safety violation.
 
     This test specifically triggers use-after-free scenarios if the underlying C extension
-    hash table `md->keys` resizes concurrently during an unresolved iteration sequence.
+    hash table ``md->keys`` resizes concurrently during an unresolved iteration sequence.
     Under free-threaded CPython (GIL disabled), this previously resulted in a SIGSEGV.
 
     With the issue fixed, the code securely catches size mutations and cleanly raises
-    a standard Python `RuntimeError` ('MultiDict is changed during iteration'), preventing
+    a standard Python ``RuntimeError`` ('MultiDict is changed during iteration'), preventing
     crashes.
     """
     if cls.__module__ == "multidict._multidict_py":
@@ -30,14 +30,9 @@ def test_race_condition_iterator_vs_mutation(
     errors: list[tuple[str, int, str, str, str]] = []
 
     def writer(target: Union[CIMultiDict[str], MultiDict[str]]) -> None:
-        for i in range(5000):
+        for i in range(256):
             try:
                 target[f"k-{i % 64}"] = f"v{i}"
-                if i % 13 == 0:
-                    try:
-                        del target[f"k-{i % 64}"]
-                    except KeyError:
-                        pass  # Deleting non-existent keys dynamically is fine
             except RuntimeError:
                 pass  # "MultiDict is changed during iteration" is expected
             except Exception as e:
@@ -48,7 +43,7 @@ def test_race_condition_iterator_vs_mutation(
                 )
 
     def reader(target: Union[CIMultiDict[str], MultiDict[str]]) -> None:
-        for i in range(5000):
+        for i in range(256):
             try:
                 list(target.items())
                 list(target.keys())
--- a/tests/test_free_threading.py
+++ b/tests/test_free_threading.py
@@ -1,5 +1,8 @@
 from __future__ import annotations
 
+import os
+import subprocess
+import sys
 import threading
 
 import pytest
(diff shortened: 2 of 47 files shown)


rodrigobnogueira:
Note: This PR temporarily includes the changes from #1316 (Drop Python 3.9 support) so that the CI/cibuildwheel pipeline can pass — the cp39 wheel build was failing due to setuptools >= 82.0.1 being incompatible with Python 3.9. Once #1316 is merged into master, I will rebase this PR to remove that commit and keep only the free-threaded race condition fix.

FYI @Dreamsorcerer

Vizonex (commented):
Looks good surprised this wasn't seen with the sniffer from #1306. Be sure you update your branches if you haven't since #1314 Got merged which I wrote to help with broken coverage but other than that I have no suggestions.

Vizonex:
@rodrigobnogueira it's been 2 weeks so I'll see if we can get someone else in here to merge since I don't have that authorization to do so yet. for now I recommend staying updated with the main branch.

CHANGES/1317.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+Fixed a memory-safety race condition resulting in segmentation faults (Use-After-Free) when iterating and modifying a ``MultiDict`` concurrently in CPython free-threaded mode (3.13t+). Read/Write accesses to the internal ``keys`` buffer are now wrapped in ``Py_BEGIN_CRITICAL_SECTION``
```
webknjaz: Can we use RST roles to reference the highlighted identifiers?

tests/test_free_threading.py
```diff
@@ -0,0 +1,70 @@
...
+from typing import Type, Union
+
+import pytest
+
+from multidict import CIMultiDict, MultiDict
+
+
+@pytest.mark.parametrize("cls", [CIMultiDict, MultiDict])
+def test_race_condition_iterator_vs_mutation(
+    cls: Union[Type[CIMultiDict[str]], Type[MultiDict[str]]],
```
webknjaz: This should probably be just
```suggestion

def test_race_condition_iterator_vs_mutation(
    any_multidict_class: type[CIMultiDict[str] | type[MultiDict[str]],
```

Vizonex: I'm pretty sure it wasn't updated with some of the newer branches. I'm sure if @rodrigobnogueira were to update the branch pre-commit would catch it and try to autofix it.

rodnogueira:
While fixing the C extension race condition, our test suite (now running the `any_multidict_class` fixture against both implementations) uncovered a related but separate thread-safety bug in _multidict_py.py on free-threaded CPython (3.13t).

Root cause: `__setitem__` temporarily sets `e.hash = -1` as a sentinel marker while traversing entries, relying on `restore_hash()` to clean up afterwards. Under free-threaded CPython, a concurrent thread can enter `_add_with_hash` → `_resize` → `build_indices(update=False)` in that window, where it hits:

```python
assert hash_ != -1  # line 543
```

...and raises AssertionError because it observed the in-progress -1 sentinel from the other thread.

The fix for the C extension in this PR uses Py_BEGIN_CRITICAL_SECTION, but the pure-Python implementation has no equivalent synchronization. Fixing it properly would require adding a threading.Lock to MultiDict.

I've skipped the test for the pure-Python variant for now to unblock this PR.

rodnogueira:
Error showed up in CI:

```
('writer', 34, 'AssertionError', '', 'Traceback...
  File "_multidict_py.py", line 881, in __setitem__
    self._add_with_hash(_Entry(hash_, identity, key, value))
       ↓ (usable <= 0, triggers resize)
  File "_multidict_py.py", line 1088, in _add_with_hash
    self._resize((self._used * 3 | _HtKeys.MINSIZE - 1).bit_length(), False)
       ↓
  File "_multidict_py.py", line 1083, in _resize
    newkeys.build_indices(update)
       ↓
  File "_multidict_py.py", line 543, in build_indices
    assert hash_ != -1
AssertionError
```

Thread A (__setitem__) sets e.hash = -1 on line 874 as a sentinel
Thread B (concurrently in _add_with_hash) triggers a resize, calling build_indices(update=False) which iterates the same entries and hits Thread A's -1 sentinel → AssertionError

rodrigobnogueira:
While fixing the C extension race condition, our test suite (now running the `any_multidict_class` fixture against both implementations) uncovered a related but separate thread-safety bug in _multidict_py.py on free-threaded CPython (3.13t).

Root cause: `__setitem__` temporarily sets `e.hash = -1` as a sentinel marker while traversing entries, relying on `restore_hash()` to clean up afterwards. Under free-threaded CPython, a concurrent thread can enter `_add_with_hash` → `_resize` → `build_indices(update=False)` in that window, where it hits:

```python
assert hash_ != -1  # line 543
```

...and raises AssertionError because it observed the in-progress -1 sentinel from the other thread.

The fix for the C extension in this PR uses Py_BEGIN_CRITICAL_SECTION, but the pure-Python implementation has no equivalent synchronization. Fixing it properly would require adding a threading.Lock to MultiDict.

I've skipped the test for the pure-Python variant for now to unblock this PR.

rodrigobnogueira:
Error showed up in CI (traceback from **Thread B**):

```
  File "_multidict_py.py", line 881, in __setitem__
    self._add_with_hash(_Entry(hash_, identity, key, value))
       ↓ (usable <= 0, triggers resize)
  File "_multidict_py.py", line 1088, in _add_with_hash
    self._resize((self._used * 3 | _HtKeys.MINSIZE - 1).bit_length(), False)
       ↓
  File "_multidict_py.py", line 1083, in _resize
    newkeys.build_indices(update)
       ↓
  File "_multidict_py.py", line 543, in build_indices
    assert hash_ != -1
AssertionError
```

Two concurrent `__setitem__` calls race:
- **Thread A** is updating an **existing** key — it sets `e.hash = -1` (line 874) as a sentinel while iterating, and has not yet called `restore_hash()`.
- **Thread B** is inserting a **new** key — it enters the `not found` branch (line 880–881), calls `_add_with_hash` → `_resize` → `build_indices(update=False)`, which iterates the **same shared entries** and hits Thread A's `-1` sentinel → `AssertionError`.

Vizonex:
See #1327

Dreamsorcerer:
I'm unclear how much we should be concerned with such issues. The entire point of aio-libs is to provide asyncio libraries, so threading issues shouldn't be relevant. I'd rather not start introducing threading locks into our libraries..

Vizonex:
> I'm unclear how much we should be concerned with such issues. The entire point of aio-libs is to provide asyncio libraries, so threading issues shouldn't be relevant. I'd rather not start introducing threading locks into our libraries..

@Dreamsorcerer You have to remember to that multidict is used in other libraries or projects besides our own. An example that I have is [litestar](https://github.com/litestar-org/litestar/blob/ca374cec33fdb3b7338f3310e1d2aa17c3dae0c4/pyproject.toml#L42) however I'm with you on questioning weather or not thread locks should be considered acceptable or unacceptable however.

rodrigobnogueira:
> I'm unclear how much we should be concerned with such issues. The entire point of aio-libs is to provide asyncio libraries, so threading issues shouldn't be relevant. I'd rather not start introducing threading locks into our libraries..

Adding threading.Lock to the pure-Python fallback would add overhead for all users to solve a problem that is probably an edge case. Would it be acceptable to emit a RuntimeWarning at import time only when both conditions are true: running on free-threaded CPython (GIL disabled) AND the C extension is not available?

Dreamsorcerer:
> Adding threading.Lock to the pure-Python fallback would add overhead for all users to solve a problem that is probably an edge case. Would it be acceptable to emit a RuntimeWarning at import time only when both conditions are true: running on free-threaded CPython (GIL disabled) AND the C extension is not available?

We're talking about just iterating and modifying the same multidict, right? Can just mention that this shouldn't be done in a threaded situation. Maybe issue a warning.

rodrigobnogueira:
> > Adding threading.Lock to the pure-Python fallback would add overhead for all users to solve a problem that is probably an edge case. Would it be acceptable to emit a RuntimeWarning at import time only when both conditions are true: running on free-threaded CPython (GIL disabled) AND the C extension is not available?
> 
> We're talking about just iterating and modifying the same multidict, right? Can just mention that this shouldn't be done in a threaded situation. Maybe issue a warning.

Added a RuntimeWarning in __init__.py that fires only when the pure-Python fallback is used on free-threaded CPython (GIL disabled + C extension unavailable).

Vizonex:
Are we still planning to move forward with these free-threaded PRs at all or will these be closed?

rodrigobnogueira:
pre-commit surfaced the new PLC0415 rule, fixed in https://github.com/aio-libs/multidict/commit/4079a3ee787295e7690e87e9fd80a42566130de2.

rodrigobnogueira:
> Are we still planning to move forward with these free-threaded PRs at all or will these be closed?

IMO this solves the issue, and the main cost is how many functions it has to touch: every entry point that reads md->keys needs the critical section, the same way CPython's dict does it. The sections are no-ops on regular builds, so only free-threaded builds pay.

I think it deserves fixing: this is memory unsafety, not a logical race. A pure-Python user on 3.13t can segfault the process, where CPython's dict raises RuntimeError. The lighter alternative is documenting it as unsupported and keeping only the RuntimeWarning, but that leaves a crash reachable from Python code.

devdanzin:
This is the right approach — mirroring CPython `dict`'s per-object critical sections at every entry point, no-op on the default build. Two follow-ups:

**1. The cross-object paths lock only `self`.** `multidict_tp_richcompare` correctly uses `Py_BEGIN_CRITICAL_SECTION2(self, other_md)`, but `multidict_extend` / update / merge-from-another-multidict are wrapped `Py_BEGIN_CRITICAL_SECTION(self)` only — and `md_update_from_ht` / `md_clone_from_ht` read the **source** `other->keys`. A thread mutating `other` while another does `a.update(other)` still races the read. These want `CRITICAL_SECTION2(self, other)` too. (`multidict_copy` is fine — there `self` *is* the source.)

**2. A few of the crashes are reachable *single-threaded*, so they survive the critical sections.** A critical section is a no-op on the default (GIL) build and doesn't guard same-thread reentrancy, so these need the function bodies fixed regardless of locking — they're borrowed-pointer-across-callback bugs, not pure data races:

- **`md_update_from_ht`** caches `htkeys_entries(other->keys)` and holds borrowed `entry->key/value` across a resize / `key.lower()`. Reproduces on a plain debug and ASan build, no threads:

  ```python
  d = MultiDict([(f"k{i}", i) for i in range(64)])
  d.extend(d)   # self-alias: growing d resizes/frees d->keys mid-copy -> SIGSEGV
  ```

- **`md_update_from_seq`** caches the list size once, then a CI key's `.lower()` can shrink the list → `PyList_GET_ITEM` past the end (OOB). Also reproduces single-threaded.
- **`md_eq`** holds `lft_entries`/`rht_entries` across `PyObject_RichCompareBool`; a value's `__eq__` that mutates a dict dangles the cached base (reentrant even under the GIL).

The fix for these is in the bodies: re-fetch the entries base after any call that can mutate/resize (the `md_next`/`md_find_next` re-fetch pattern already used elsewhere), snapshot the size, and for the self-alias case detect `other == self`. I have runnable reproducers for each and am glad to open them as separate issues (they overlap the C-extension report #1306). None of this blocks the concurrency fix here — just flagging that the critical sections alone won't close these particular crashes.

espressolee:
Doing a free-threading pass over multidict I built a tree with this PR, #1327 and #1379 all merged, and one path this PR does not reach turned up. Flagging it here rather than opening an issue, since it belongs to the work you are already doing.

`_multidict_extend` (`_multidict.c:81`) calls `md_update_from_ht(self, other, op)` with no critical section. `md_update_from_ht` then takes a raw pointer into the *other* multidict's table and walks it:

```c
entry_t *entries = htkeys_entries(other->keys);
Py_ssize_t nentries = other->keys->nentries;

for (pos = 0; pos < nentries; pos++) {
    entry_t *entry = entries + pos;
```

The four `Py_BEGIN_CRITICAL_SECTION2` uses this PR adds are all in `multidict_tp_richcompare`, so on the extend path `other` is never locked. A concurrent insert into `other` can resize and free the array `entries` points at.

Measured on 3.14.0rc1t with all three PRs applied — 8 threads running `MultiDict().extend(other)`, 3 mutating `other`:

```
original                                    SIGSEGV 12/12
control: no mutator                         clean 12/12
control: mutator touches a decoy multidict  clean 12/12
control: md->is_ci == other->is_ci          SIGSEGV 11, SIGTRAP 1
GIL build, identical script                 clean 12/12
```

The last control is the one that places it: with matching `is_ci`, `recalc_identity` is false and no Python runs inside the loop, and it still crashes. So this is not a borrow-across-Python problem — it is the `_md_resize` race from devdanzin's #1321 cluster 3, reachable through a path this PR leaves unprotected. `Py_BEGIN_CRITICAL_SECTION2(self, other)` around the `md_update_from_ht` calls looks like it would close it, but you know the locking design here better than I do.

Not reporting this as a separate finding — it is the race already documented, not a new one. Reproducer available if useful.

bdraco:
Mobile so I can’t check if it’s the same one but there is a zero performance impact to the non free threaded version of this in a GHSA report

rodrigobnogueira:
Reworked this to tests only. #1431 through #1446 covered everything the C side here was doing, so I dropped it and kept the two race tests: iterate while other threads mutate, and extend while other threads resize the source. Against the tree before that series both segfault on every run; against current master they pass 10/10.

tests/test_free_threading.py
```diff
@@ -0,0 +1,153 @@
+from __future__ import annotations
```
asvetlov:
```suggestion
```

asvetlov (approved):

asvetlov:
Thanks for tests, @rodrigobnogueira !

Issue #1321, opened 2026-04-19
multidict Free-Threading Analysis Report


devdanzin:
I ran a free threading analysis tool on multidict and it found that:

> **Overall difficulty: Easy-Medium.** multidict is one of the *best-positioned* extensions for free-threading — all the hard structural work (multi-phase init, heap types, module state) is already done. What's missing is purely per-object synchronization, which is a mechanical addition of `Py_BEGIN_CRITICAL_SECTION` calls.

The full report is at https://gist.github.com/devdanzin/fa00e03ef041c030660f083116a7e1b6. It contains issue descriptions, suggested fixes, and reproducers for some of the issues.

I hope it helps.

webknjaz:
@devdanzin thanks! I think it would be useful to have the context included right in the issue so that the discussion isn't fragmented. Perhaps, as a collapsed block. 
Though, it may be better to file multiple issues that could be addressed independently. Sending PRs with just reproduces would be a great demo too.

devdanzin:
Thanks @webknjaz! That makes sense, and sorry for the slow follow-up. Inlining the context below as a collapsed block, and I agree the right shape is a few independently-addressable issues rather than one big report. I've since reproduced each item on a free-threaded TSan build, so I can attach a runnable reproducer to each.

<details>
<summary><b>Free-threading analysis — multidict (summary)</b></summary>

**The good news first:** multidict is unusually well-positioned. The hard structural work is already done — multi-phase init, heap types, per-module state, `Py_MOD_GIL_NOT_USED`. What's missing is purely per-object synchronization: the C extension declares GIL-not-used but does no per-object locking, so every public entry point that touches `md->keys`/`md->used`/`md->version` can race, and `_md_resize` can free `md->keys` under a concurrent reader.

Reproduced on a free-threaded TSan build (`PYTHON_GIL=0`): **~40 data races per run across ~31 unique sites**, and a run aborts under contention. The clusters:

| # | site | what races |
|---|------|-----------|
| 1 | `state.h` `NEXT_VERSION` (`++global_version`) | non-atomic RMW on the module-global counter (all instances share it) |
| 2 | `hashtable.h` `_md_add_*` insert path | `nentries`/`usable`/`used`/entry slots/index written unprotected |
| 3 | `hashtable.h` `_md_resize` | reallocs + frees `md->keys` while another thread reads it → use-after-free |
[...]

PR #1323, opened 2026-04-21
add atomic global counter system

Vizonex:
## What do these changes do?

These changes are based off @devdanzin's second fuzzer for freethreading mode which noted using a global atomic counter system instead of a normal `uint64_t` value I did however make one small nitpick/change from that original design to just recast `(_Atomic(uint64_t)*)&state->global_version` which was not in the original suggestion mainly due to my code editor misbehaving on me when I would go to set it into the actual structure. 

Just bug fixes.

## Related issue number
#1321 

## Checklist

### DIFF
--- a/CHANGES/1328.bugfix.rst
+++ b/CHANGES/1328.bugfix.rst
@@ -1,2 +1,2 @@
-Fixed global counter system using an atomic variable.
+Fixed global counter system using an atomic variable
 -- by :user:`Vizonex`.
--- a/CHANGES/1328.bugfix.rst
+++ b/CHANGES/1328.bugfix.rst
@@ -1,0 +1,2 @@
+Fixed global counter system using an atomic variable.
+-- by :user:`Vizonex`.
--- a/multidict/_multilib/state.h
+++ b/multidict/_multilib/state.h
@@ -134,7 +134,7 @@ NEXT_VERSION(mod_state *state)
      * itself */
     return atomic_fetch_add_explicit(
                &state->global_version, 1, memory_order_relaxed) +
-           1
+           1;
 }
 
 #ifdef __cplusplus
--- a/multidict/_multilib/state.h
+++ b/multidict/_multilib/state.h
@@ -28,7 +28,7 @@ typedef struct {
     PyObject *str_lower;
     PyObject *str_name;
 
-    uint64_t global_version;
+    _Atomic uint64_t global_version;
 } mod_state;
 
 static inline mod_state *
@@ -131,7 +131,7 @@ static inline uint64_t
 NEXT_VERSION(mod_state *state)
 {
     return atomic_fetch_add_explicit(
-        (_Atomic(uint64_t) *)&state->global_version, 1, memory_order_relaxed);
+        &state->global_version, 1, memory_order_relaxed);
 }
 
 #ifdef __cplusplus
--- a/multidict/_multilib/state.h
+++ b/multidict/_multilib/state.h
@@ -131,7 +131,8 @@ static inline uint64_t
 NEXT_VERSION(mod_state *state)
 {
     return atomic_fetch_add_explicit(
-        &state->global_version, 1, memory_order_relaxed) + 1;
+               &state->global_version, 1, memory_order_relaxed) +
+           1;
 }
 
 #ifdef __cplusplus
--- a/multidict/_multilib/state.h
+++ b/multidict/_multilib/state.h
@@ -131,7 +131,7 @@ static inline uint64_t
 NEXT_VERSION(mod_state *state)
 {
     return atomic_fetch_add_explicit(
-        &state->global_version, 1, memory_order_relaxed);
+        &state->global_version, 1, memory_order_relaxed) + 1;
 }
 
 #ifdef __cplusplus
--- a/multidict/_multilib/state.h
+++ b/multidict/_multilib/state.h
@@ -130,9 +130,10 @@ get_mod_state_by_def(PyObject *self)
 static inline uint64_t
 NEXT_VERSION(mod_state *state)
 {
-    return atomic_fetch_add_explicit(
-               &state->global_version, 1, memory_order_relaxed) +
-           1;
+    /* threads have the ability to screw with the global version,
+     * using an atomic variable ensures the global version is
+     * always accurate. */
+    return atomic_fetch_add(&state->global_version, 1) + 1;
 }
 
 #ifdef __cplusplus
--- a/multidict/_multilib/state.h
+++ b/multidict/_multilib/state.h
@@ -130,7 +130,8 @@ get_mod_state_by_def(PyObject *self)
 static inline uint64_t
 NEXT_VERSION(mod_state *state)
 {
-    return atomic_fetch_add_explicit((_Atomic(uint64_t)*)&state->global_version, 1, memory_order_relaxed);
+    return atomic_fetch_add_explicit(
+        (_Atomic(uint64_t) *)&state->global_version, 1, memory_order_relaxed);
 }
 
 #ifdef __cplusplus
--- a/multidict/_multilib/state.h
+++ b/multidict/_multilib/state.h
@@ -130,10 +130,8 @@ get_mod_state_by_def(PyObject *self)
 static inline uint64_t
 NEXT_VERSION(mod_state *state)
 {
-    /* threads have the ability to screw with the global version,
-     * using an atomic variable ensures the global version is
-     * always accurate. */
-    return atomic_fetch_add(&state->global_version, 1) + 1;
+    /* relaxed is fine here as we only care about the atomicity of the RMW itself */
+    return atomic_fetch_add_explicit(&state->global_version, 1, memory_order_relaxed) + 1
 }
 
 #ifdef __cplusplus
--- a/multidict/_multilib/state.h
+++ b/multidict/_multilib/state.h
@@ -5,6 +5,8 @@
 extern "C" {
 #endif
 
+#include <stdatomic.h>
+
 /* State of the _multidict module */
 typedef struct {
     PyTypeObject *IStrType;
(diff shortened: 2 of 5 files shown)


multidict/_multilib/state.h
```diff
@@ -128,7 +130,8 @@ get_mod_state_by_def(PyObject *self)
 static inline uint64_t
 NEXT_VERSION(mod_state *state)
 {
-    return ++state->global_version;
+    return atomic_fetch_add_explicit(
+        &state->global_version, 1, memory_order_relaxed);
```
bdraco:
```suggestion
    return atomic_fetch_add_explicit(
        &state->global_version, 1, memory_order_relaxed) + 1;
```

Lets keep the semantics the same

bdraco: could use a comment explaining why relaxed is safe here

bdraco (approved):

multidict/_multilib/state.h
Vizonex: Agreed I didn't exactly know which is why I left it blank but if it had to make a guess it has to do with different threads sharing a single state. I'm surprised we didn't use multiple states one per dictionary.

multidict/_multilib/state.h
```diff
@@ -128,7 +130,10 @@ get_mod_state_by_def(PyObject *self)
 static inline uint64_t
 NEXT_VERSION(mod_state *state)
 {
-    return ++state->global_version;
+    /* threads have the ability to screw with the global version,
+     * using an atomic variable ensures the global version is
+     * always accurate. */
+    return atomic_fetch_add(&state->global_version, 1) + 1;
```
bdraco: We don't need a full memory fence, thats a lot more expensive. your original relaxed was fine here. I just wanted a comment explaining why.

 
```suggestion
    /* relaxed is fine here as we only care about the atomicity of the RMW itself */
    return atomic_fetch_add_explicit(&state->global_version, 1, memory_order_relaxed) + 1
```

setup.py
```diff
@@ -24,6 +24,13 @@
             "-Werror",
         ]
     )
+else:
```
bdraco: probably should gate to MSVC only. as it will break non MSVC

Vizonex: True, Part of the problem I had was that msvc is very picky about what it wants for stdatomic flags do you have any suggestions on how I could approach checking msvc after the first platform check?

setup.py
bdraco:
```python
BASE_CFLAGS = ["-O0", "-g3", "-UNDEBUG"] if DEBUG_BUILD else ["-O3", "-DNDEBUG"]                                                                                                                                                                
  
UNIX_CFLAGS = [                                                                                                                                                                                                                                 
      "-std=c11", "-Wall", "-Wsign-compare", "-Wconversion",
      "-fno-strict-aliasing", "-Wno-conversion", "-Werror",                                                                                                                                                                                       
  ]                                                                                                                                                                                                                                               

MSVC_CFLAGS = ["/std:c11", "/experimental:c11atomics"]  
```

bdraco: than gate in `BuildExt` `build_extensions` on `self.compiler.compiler_type` == "msvc" else. the unix flags

bdraco: you'll have to refactor a bit somthing like
```python
extensions = [Extension("multidict._multidict", ["multidict/_multidict.c"])]                                                                                                                                                                    
                                                                                                                                                                                                                                                  
  if not NO_EXTENSIONS:      
      setup(ext_modules=extensions, cmdclass={"build_ext": BuildExt})
  else:                                                                                                                                                                                                                                           
      setup()
```

bdraco: than 
```python
  class BuildExt(build_ext):
      def build_extensions(self):                                                                                                                                                                                                                 
          check self.compiler.compiler_type here and build flags
          for ext in self.extensions:                                                                                                                                                                                                             
              ext.extra_compile_args = build up here                                                                                                                                                      
          super().build_extensions() 
```

bdraco: sorry for the disjointed suggestions. traveling today

bdraco (approved):
Thanks @Vizonex

bdraco:
Looks like the tests still need some work

Vizonex:
> Looks like the tests still need some work

@bdraco I think the functions aren't all locked yet which is why the numbers are still off other than that I agree with you. I think I'll create an extra PR today for locking the functions down in free-threaded mode since I have the time to.

CHANGES/1328.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+Fixed global counter system using an atomic variable.
```
webknjaz:
```suggestion
Fixed global counter system using an atomic variable
```

tests/isolated/multidict_global_counter.py
```diff
@@ -0,0 +1,33 @@
...
+N, M = 3, 100
+baseline = multidict.getversion(md)  # type: ignore[arg-type]
+
+
+def worker(tid: int) -> None:
+    for i in range(M):
+        md[f"k{tid}_{i}"] = i
+
+
+if (__name__ == "__main__") and FREETHREADED:
```
webknjaz, line 20: Same concern as in other PRs: we shouldn't need to gate on FT.

cc @Dreamsorcerer

tests/isolated/multidict_global_counter.py
```diff
@@ -0,0 +1,33 @@
...
+
+import multidict
+from multidict import MultiDict
+
+FREETHREADED = bool(sysconfig.get_config_var("Py_GIL_DISABLED"))
+
+
+md: MultiDict[int] = MultiDict()
+N, M = 3, 100
+baseline = multidict.getversion(md)  # type: ignore[arg-type]
```
webknjaz, line 12: why ignore?

tests/isolated/multidict_global_counter.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+if (__name__ == "__main__") and FREETHREADED:
+    threads = [threading.Thread(target=worker, args=(tid,)) for tid in range(N)]
+    for t in threads:
+        t.start()
+    for t in threads:
+        t.join()
+
+    observed = multidict.getversion(md) - baseline  # type: ignore[arg-type]
```
webknjaz, line 27: why ignore?

tests/isolated/multidict_global_counter.py
```diff
@@ -0,0 +1,33 @@
+import sysconfig
+import threading
+
+import multidict
```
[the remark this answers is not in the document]
Dreamsorcerer: Should probably resolve this, no reason to import it twice.

tests/isolated/multidict_global_counter.py
Dreamsorcerer: If there's an error here, then maybe my strict `Multidict[object]` is causing problems. Might need to be changed to `Multidict[Any]`. Probably a contravariant thing I missed.

tests/isolated/multidict_global_counter.py
Vizonex: `MultiDict[object]` is what getversion(..) wants but here we have a completely different type hence `type: ignore[arg-type]`

bdraco:
@aiolibsbot review

tests/isolated/multidict_global_counter.py
aiolibsbot: Agreeing with this concern. Now that `global_version` is atomic, the invariant `delta == N*M` should hold under the GIL too — it just becomes a less interesting test. Gating on `FREETHREADED` means non‑FT CI runs the script for no benefit (still pays a subprocess + import) and any future regression that reverts `NEXT_VERSION` back to a plain `++` won't be caught unless someone happens to run an FT build locally. Unless there's a concrete `__setitem__` path on the GIL build that legitimately skips a `NEXT_VERSION` call, I'd drop the gate. If the assertion *does* fail on a GIL build today, that's a separate bug that this PR is masking by gating it out.

tests/isolated/multidict_global_counter.py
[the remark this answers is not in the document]
aiolibsbot: Concur — easiest fix is `from multidict import MultiDict, getversion` and then call `getversion(md)` directly. That also removes one of the two `type: ignore[arg-type]` comments below, since `getversion` would be imported with its real signature rather than accessed through the bare module namespace.

setup.py
aiolibsbot: The standard pattern for distutils/setuptools is to override `build_ext.build_extensions` and branch on `self.compiler.compiler_type` (which is `"msvc"`, `"unix"`, `"mingw32"`, …) rather than on `platform.system()`. That way you're matching the *toolchain*, not the OS — important because someone could build with clang‑cl or mingw on Windows, where MSVC flags would be wrong. @bdraco's sketch in the other comments shows exactly this. Heads up: with the current `BASE_CFLAGS = ["O0", "g3", ...]` approach, MSVC debug builds will emit `/O0` and `/g3`, both of which are invalid MSVC options (`/Od` and `/Z7`/`/Zi` are the correct ones). Keeping the original `-`‑prefixed Unix flags and writing an explicit MSVC table (`/Od`, `/Zi`, `/UNDEBUG` for debug; `/O2`, `/DNDEBUG` for release) sidesteps that entirely.

devdanzin:
The core change looks correct to me — `_Atomic uint64_t global_version` with `atomic_fetch_add_explicit(..., memory_order_relaxed)` is exactly right (relaxed is fine; the counter only needs RMW atomicity, not ordering). The one thing to keep in mind is the co-dependency already noted in the thread: on its own the atomic counter doesn't make `MultiDict.__setitem__` thread-safe (the per-object `md->version` read and the insert path still need #1317's critical sections), so this and #1317 really want to land together, and the counter test will stay flaky until then. No concerns with the change itself.

asvetlov:
Thanks for the PR, @Vizonex 
The recently landed #1446 has a very similar fix

PR #1324, opened 2026-04-22
Upgrade typehints to 3.10 annotations and remove 3.9 `Union` and `Optional` typehints

Vizonex:
## What do these changes do?

3.9 was dropped recently so I've chosen to upgrade the annotations away from 3.9 and into 3.10 which includes new the popular pipe syntax instead of `Union[A, B]` or `Optional[A]` in favor of `A | B` `A | None` instead.

Users will likely enjoy the changes from the new 3.10 annotations over the older `Union` and `Optional` type hints especially since these types are both deprecated.

## Related issue number

This pull request is related to changes from #1316

## Checklist


bdraco:
Looks like we are missing pyupgrade on this repo

Vizonex:
> Looks like we are missing pyupgrade on this repo

@bdraco do you have any ideas of how we should go about doing something like that? I was thinking about implementing a few flags within ruff to get something like that to work.

PR #1325, opened 2026-04-22
Enable ruff to perform syntax upgrades

Vizonex:
## What do these changes do?

This change enables ruff to make automatic upgrades when moving up versions of python or dropping versions (for example 3.9)

This should be helpful for other contributors and users.

## Related issue number

#1324 this was discussed by me and @bdraco, I decided to branch out once again as this portion is a big enough change that requires another PR.
close #1324

## Checklist

### DIFF
--- a/multidict/_multidict_py.py
+++ b/multidict/_multidict_py.py
@@ -18,9 +18,7 @@ from typing import (
     ClassVar,
     Generic,
     NoReturn,
-    Optional,
     TypeVar,
-    Union,
     cast,
     overload,
 )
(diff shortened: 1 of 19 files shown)


bdraco:
Definitely easier to review since only a few things need to be manually checked as we don't have to check each line of the pre-commit update.

multidict/_multidict_py.py
bdraco: I wonder why it didn't hit on this one

bdraco:
```

  >       def __init__(self, arg: MultiDict[_V] | "MultiDictProxy[_V]"):
                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  E       TypeError: unsupported operand type(s) for |: 'types.GenericAlias' and 'str'
```

bdraco (approved):
Thanks @Vizonex

PR #1326, opened 2026-04-22
Drop 3.13t support

ngoldbaum:
## What do these changes do?

See the note at https://cibuildwheel.pypa.io/en/stable/changelog/#v341 and https://github.com/pypa/cibuildwheel/pull/2787.

The `cpython-freethreading` option will be removed in the next release of cibuildwheel.

I'm going through repositories on GitHub that enable this option and sending in PRs to disable it. Happy to answer questions about this. I also took the opportunity to drop 3.13t CI.

If you're curious why 3.13t support is going away so soon, see https://py-free-threading.github.io/ci/#building-free-threaded-wheels-with-cibuildwheel for further guidance.

The next release will not include cp313t wheel builds.

## Checklist

### DIFF
--- a/CHANGES/1326.breaking.rst
+++ b/CHANGES/1326.breaking.rst
@@ -1,1 +1,0 @@
-Dropped support for free-threaded Python 3.13 -- by :user:`ngoldbaum`.
--- a/CHANGES/1326.breaking.rst
+++ b/CHANGES/1326.breaking.rst
@@ -1,1 +1,1 @@
-Dropped support for free-threaded Python 3.13
+Dropped support for free-threaded Python 3.13 -- by :user:`ngoldbaum`.
--- a/CHANGES/1326.breaking.rst
+++ b/CHANGES/1326.breaking.rst
@@ -1,0 +1,1 @@
+1326.packaging.rst
--- a/CHANGES/1326.breaking.rst
+++ b/CHANGES/1326.breaking.rst
@@ -1,0 +1,1 @@
+Dropped support for free-threaded Python 3.13
--- a/CHANGES/1326.packaging.rst
+++ b/CHANGES/1326.packaging.rst
@@ -1,1 +1,1 @@
-CHANGES/1326.breaking.rst
+Dropped support for free-threaded Python 3.13 -- by :user:`ngoldbaum`.
--- a/CHANGES/1326.packaging.rst
+++ b/CHANGES/1326.packaging.rst
@@ -1,0 +1,1 @@
+CHANGES/1326.breaking.rst
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -176,6 +176,7 @@ jobs:
         pyver:
         - 3.14t
         - 3.14
+        # Skipping 3.13t because cibuildwheel dropped support
         - 3.13
         - 3.12
         - 3.11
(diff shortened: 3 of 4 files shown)


Vizonex:
@ngoldbaum Glad to see you here, seems some other people might be busy so I'll help you with the reviewing process.

Vizonex (approved):
All of this looks good might need to add a comment or two about why __3.13t__ is being removed but other than that I approve of this pull request.

Vizonex (approved):
That should be good enough now.

Vizonex (commented):
Waiting for feedback about the changes I let everyone on the matrix know that a workflow was left hanging.

CHANGES/1326.breaking.rst
webknjaz, line 1: Could you symlink think to `packaging`? Also, I'm not sure if this counts as breaking.

cc @Dreamsorcerer @bdraco WDYT?

CHANGES/1326.breaking.rst
```diff
@@ -0,0 +1 @@
+Dropped support for free-threaded Python 3.13
```
webknjaz:
```suggestion
Dropped support for free-threaded Python 3.13 -- by :user:`ngoldbaum`.
```

webknjaz (commented):
Apart from the change note, looks good. I'll wait for Nick and Sam to take a look before proceeding.

bdraco (approved):
LGTM once other comments are addressed. clearly 3.13t was not ready for production use so removing wheels is reasonable

ngoldbaum:
Sorry for taking two weeks to click the button! I think this is ready now.

Vizonex:
> Sorry for taking two weeks to click the button! I think this is ready now.

Good because we were worried that you just left on an hiatus, if I don't see anything new needing to be changed I'll request for this PR to be merged.

CHANGES/1326.breaking.rst
webknjaz: @ngoldbaum have you missed the symlink request?

ngoldbaum: Thanks - I misunderstood what you were referring to with the symlink. See latest push.

ngoldbaum: Except it looks like towncrier doesn't like symlinks, so I just removed the `breaking` note and called it `packaging`.

webknjaz: @ngoldbaum Towncrier works with symlinks just fine. This replaces having two text files with exactly the same contents :man_shrugging:

The failure in your previous commit was caused by an HTTP 500 when uploading MyPy type preciseness coverage to Coveralls: https://github.com/aio-libs/multidict/actions/runs/25672626609/job/75361690773?pr=1326#step:8:232

webknjaz: I'm sure restarting that job will make it green..

CHANGES/1326.packaging.rst
```diff
@@ -0,0 +1 @@
+CHANGES/1326.breaking.rst
```
webknjaz: @ngoldbaum the problem here was that you've added a symlink relative to the repo root and so any tooling resolving it, would search for `CHANGES/CHANGES/1326.packaging.rst`. The correct way would be running `ln -s` in the `CHANGES/` dir so that the relative link points to just `1326.packaging.rst` w/o the dir prefix...

webknjaz (approved):
Not sure if we need to also have this filed under “breaking changes” (see the old symlink discussion threads inline), but this looks good to me regardless.

CHANGES/1326.packaging.rst
ngoldbaum: Thanks for explaining, see latest commit.

PR #1327, opened 2026-04-28
Fix hashtable insertion race. (Free-Threaded)

Vizonex:
## What do these changes do?

This change comes related to [_md_add_with_hash_steal_refs](https://gist.github.com/devdanzin/fa00e03ef041c030660f083116a7e1b6#2-hashtable-insert-race--_md_add_with_hash_steal_refs) Which mentions a possible hashtable insertion race in free-threaded mode.

this should not affect the pure python version of multidict only the C version.

## Checklist

### DIFF
--- a/CHANGES/1327.bugfix.rst
+++ b/CHANGES/1327.bugfix.rst
@@ -1,2 +1,3 @@
 Fixed insertion race on free-threaded mode.
+
 -- by :user:`Vizonex`
--- a/CHANGES/1327.bugfix.rst
+++ b/CHANGES/1327.bugfix.rst
@@ -1,0 +1,2 @@
+Fixed hashtable insertion race on free-threaded mode.
+-- by :user:`Vizonex`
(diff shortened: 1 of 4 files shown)


Vizonex:
Would be nice if we made a label for free-threaded pull requests but I don't know if that will be necessary yet.

Vizonex:
3.13t plans to be removed by #1326 so I'm not so concerned if 3.13t fails.

webknjaz:
Labels are cheap, I've added one.

CHANGES/1327.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+Fixed insertion race on free-threaded mode.
+-- by :user:`Vizonex`
```
webknjaz, line 3:
```suggestion

-- by :user:`Vizonex`
```

tests/isolated/multidict_add_ft.py
```diff
@@ -0,0 +1,30 @@
+import os
+import subprocess
+import sys
+import sysconfig
+import textwrap
+
+FREETHREADED = bool(sysconfig.get_config_var("Py_GIL_DISABLED"))
```
webknjaz, line 7: I'm not convinced we should gate on this here.

cc @Dreamsorcerer

asvetlov:
I guess it is done by recent PRs that fix the free-threading mode.
Any multidict modification is now protected by critical sections.

PR #1328, opened 2026-04-28
Fix `__repr__` race (Free-Threaded)

Vizonex:
## What do these changes do?

Fixes another race condition seen in https://gist.github.com/devdanzin/fa00e03ef041c030660f083116a7e1b6#7-repr-race--md_repr Item 7. 

## Related issue number

#1321

## Checklist

### DIFF
--- a/CHANGES/1328.bugfix.rst
+++ b/CHANGES/1328.bugfix.rst
@@ -1,3 +1,3 @@
-Fixed race when calling for a string representation of a :py:class:`multidict.MultiDict` object.
+Fixed race when calling for a string representation of a :py:class:`~multidict.MultiDict` object.
 
 -- by :user:`Vizonex`
--- a/CHANGES/1328.bugfix.rst
+++ b/CHANGES/1328.bugfix.rst
@@ -1,2 +1,3 @@
 Fixed race when calling for a string representation of a ``MultiDict`` object.
+
 -- by :user:`Vizonex`
--- a/CHANGES/1328.bugfix.rst
+++ b/CHANGES/1328.bugfix.rst
@@ -1,2 +1,3 @@
 Fixed race when calling for a string representation of a :class:`MultiDict` object.
+
 -- by :user:`Vizonex`
--- a/CHANGES/1328.bugfix.rst
+++ b/CHANGES/1328.bugfix.rst
@@ -1,2 +1,2 @@
-Fixed race when calling for a string representation of a ``MultiDict`` object.
+Fixed race when calling for a string representation of a :class:`MultiDict` object.
 -- by :user:`Vizonex`
--- a/CHANGES/1328.bugfix.rst
+++ b/CHANGES/1328.bugfix.rst
@@ -1,3 +1,3 @@
-Fixed race when calling for a string representation of a :py:class:`MultiDict` object.
+Fixed race when calling for a string representation of a :py:class:`multidict.MultiDict` object.
 
 -- by :user:`Vizonex`
--- a/CHANGES/1328.bugfix.rst
+++ b/CHANGES/1328.bugfix.rst
@@ -1,3 +1,3 @@
-Fixed race when calling for a string representation of a :class:`MultiDict` object.
+Fixed race when calling for a string representation of a :py:class:`MultiDict` object.
 
 -- by :user:`Vizonex`
--- a/CHANGES/1328.bugfix.rst
+++ b/CHANGES/1328.bugfix.rst
@@ -1,0 +1,2 @@
+Fixed race when calling for a string representation of a ``MultiDict`` object.
+-- by :user:`Vizonex`
--- a/tests/isolated/multidict_repr_ft.py
+++ b/tests/isolated/multidict_repr_ft.py
@@ -1,0 +1,37 @@
+import os
+import subprocess
+import sys
+import sysconfig
+import textwrap
+
+FREETHREADED = bool(sysconfig.get_config_var("Py_GIL_DISABLED"))
+
+
+if __name__ == "__main__" and FREETHREADED:
+    child = textwrap.dedent("""
+        import signal, threading, time
+        signal.alarm(20)
+        from multidict import MultiDict
+        md = MultiDict([(f"k{i}", i) for i in range(50)])
+        errors = []
+        stop = threading.Event()
+        def repr_loop():
+            while not stop.is_set():
+                try: repr(md)
+                except Exception as e: errors.append(type(e).__name__)
+        def mutate_loop():
+            i = 0
+            while not stop.is_set(): md.add(f"m{i}", i); i += 1
+        ts = [threading.Thread(target=repr_loop, daemon=True) for _ in range(2)]
+        ts += [threading.Thread(target=mutate_loop, daemon=True) for _ in range(2)]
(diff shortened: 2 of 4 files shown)


CHANGES/1328.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+Fixed race when calling for a string representation of a ``MultiDict`` object.
+-- by :user:`Vizonex`
```
webknjaz, line 3:
```suggestion

-- by :user:`Vizonex`
```

CHANGES/1328.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+Fixed race when calling for a string representation of a ``MultiDict`` object.
```
webknjaz: Can you link it through the class role?

tests/isolated/multidict_repr_ft.py
```diff
@@ -0,0 +1,37 @@
+import os
+import subprocess
+import sys
+import sysconfig
+import textwrap
+
+FREETHREADED = bool(sysconfig.get_config_var("Py_GIL_DISABLED"))
```
webknjaz, line 7: I think this should be moved to the `skipif` mark in the pytest param.

webknjaz (commented):
Structural comments.

tests/isolated/multidict_repr_ft.py
Vizonex: I do also since we seem to have a ton of bugs to patch. I was thinking about a new module called `test_free_threaded.py` when #1327 is merged to just have both of them in a new file.

webknjaz: Let's see if @Dreamsorcerer and @bdraco have opinions on this, then.

Vizonex: Sounds good. :)

Dreamsorcerer: Surely the tests should pass on regular versions of Python?

CHANGES/1328.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+Fixed race when calling for a string representation of a :class:`MultiDict` object.
+-- by :user:`Vizonex`
```
webknjaz, line 3:
```suggestion

-- by :user:`Vizonex`
```

tests/isolated/multidict_repr_ft.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+if __name__ == "__main__" and FREETHREADED:
+    child = textwrap.dedent("""
```
webknjaz, line 11: I'd also get rid of the double-subprocess indirectly here. If we want to invoke a bunch of Python code with different Python flags, we can put those flags into the test parametrization. Additionally, a chunk of Python in a string is not lintable, and the coverage is not getting measured. It should definitely live in a Python module as a first-class citizen.

cc @Dreamsorcerer

CHANGES/1328.bugfix.rst
```diff
@@ -0,0 +1,3 @@
+Fixed race when calling for a string representation of a :py:class:`multidict.MultiDict` object.
```
webknjaz: You can see the current preview at https://multidict--1328.org.readthedocs.build/en/1328/changes/#bug-fixes

The identifier can be rendered shorter as follows:
```suggestion
Fixed race when calling for a string representation of a :py:class:`~multidict.MultiDict` object.
```

devdanzin:
Thanks for tackling this @Vizonex. I think there's a lock-release bug in the current form, though: the critical section is opened at the top of `md_repr`, but `Py_END_CRITICAL_SECTION()` is only in the `fail:` block. `md_repr` has three returns that *don't* go through `fail:` — the success return `return PyUnicodeWriter_Finish(writer);`, the `"MultiDict changed during iteration"` `return NULL;`, and the early `if (writer == NULL) return NULL;` — so on the common (successful) path the section is entered and never ended, which leaves `md` locked and unbalances the thread's critical-section stack.

On a free-threaded build that should deadlock as soon as the same object is `repr()`'d a second time (the first `repr()` never released the lock):

```python
# free-threaded build (Py_GIL_DISABLED), single thread is enough:
md = MultiDict([("a", 1)])
repr(md)   # success path returns WITHOUT Py_END_CRITICAL_SECTION -> md stays locked
repr(md)   # re-enters Py_BEGIN_CRITICAL_SECTION(md) on the same thread -> hangs
```

If CI is green it's worth checking that `tests/isolated/multidict_repr_ft.py` actually re-`repr`s the same object after a successful one — the loop should hit this. A single exit, or a `Py_END_CRITICAL_SECTION()` before each `return`, fixes it. Two smaller notes: the `"changed during iteration"` path also still `return NULL`s without `PyUnicodeWriter_Discard(writer)`, so it leaks the writer (a separate, pre-existing bug from the #1306 report); and this may be redundant with #1317, which already wraps every `md_repr` caller in a critical section — if #1317 lands, `md_repr` is covered at the call site and this PR might not be needed, so it's worth deciding which one owns it so they don't double up.

Vizonex:
> Thanks for tackling this @Vizonex. I think there's a lock-release bug in the current form, though: the critical section is opened at the top of `md_repr`, but `Py_END_CRITICAL_SECTION()` is only in the `fail:` block. `md_repr` has three returns that _don't_ go through `fail:` — the success return `return PyUnicodeWriter_Finish(writer);`, the `"MultiDict changed during iteration"` `return NULL;`, and the early `if (writer == NULL) return NULL;` — so on the common (successful) path the section is entered and never ended, which leaves `md` locked and unbalances the thread's critical-section stack.
> 
> On a free-threaded build that should deadlock as soon as the same object is `repr()`'d a second time (the first `repr()` never released the lock):
> 
> ```python
> # free-threaded build (Py_GIL_DISABLED), single thread is enough:
> from multidict import MultiDict
> md = MultiDict([("a", 1)])
> repr(md)   # success path returns WITHOUT Py_END_CRITICAL_SECTION -> md stays locked
> repr(md)   # re-enters Py_BEGIN_CRITICAL_SECTION(md) on the same thread -> hangs
> ```
> 
> If CI is green it's worth checking that `tests/isolated/multidict_repr_ft.py` actually re-`repr`s the same object after a successful one — the loop should hit this. A single exit, or a `Py_END_CRITICAL_SECTION()` before each `return`, fixes it. Two smaller notes: the `"changed during iteration"` path also still `return NULL`s without `PyUnicodeWriter_Discard(writer)`, so it leaks the writer (a separate, pre-existing bug from the #1306 report); and this may be redundant with #1317, which already wraps every `md_repr` caller in a critical section — if #1317 lands, `md_repr` is covered at the call site and this PR might not be needed, so it's worth deciding which one owns it so they don't double up.

@devdanzin The other maintainers and I are still debating on Free-Threaded support, while we don't necessary see a need for it to be included I'm glad your still keeping us going. I might return to this bug fix if I feel more encouragement though. There was also an idea to move to Rust (Pyo3) to stop a lot of these noticeable memory leaks and other related bugs but the probability of that happing is very small as I see rust hindering the current performance benefits.

asvetlov:
Superseded by #1431
@Vizonex thanks for the initial implementation

PR #1351, opened 2026-05-18
ci: install Python via astral-sh/setup-uv to skip slow downloads

bdraco:
## What do these changes do?

Port aio-libs/yarl#1704 to multidict. Three logical steps, mirrored
1:1 in the commit history:

1. Swap ``actions/setup-python@v6`` for ``astral-sh/setup-uv@v8`` in
   the two CI jobs that install a Python interpreter
   (``build-pure-python-dists`` and the matrix ``test``).
   ``activate-environment: true`` keeps ``python``/``pip`` on PATH so
   the rest of the pipeline is unchanged. The ``allow-prereleases``
   input has no equivalent on setup-uv because uv treats prereleases
   as first-class.
2. Pin to the immutable ``v8.1.0`` tag. The floating ``v8`` major tag
   does not exist; setup-uv switched to immutable releases at v8.0.0
   and only publishes per-version tags from then on.
3. Drop the pip-cache wrapper and ``py-actions/py-dependency-install``
   and install via ``uv pip`` instead. ``setup-uv`` does not seed
   pip into the venv, so the existing ``pip cache dir`` probe would
   crash the job; rather than seed pip back in just to keep the
   wrapper working, switch to ``enable-cache: true`` (which persists
   uv's own wheel cache across runs) and ``uv pip install -r
   requirements/pytest.txt``. The ``pip --dry-run --report=- | jq``
   wheel-discovery trick collapses into a single ``uv pip install
   --find-links=./dist`` call; uv picks the right wheel for the
   active interpreter directly.

Net: 58 fewer lines in the workflow and one tool (uv) doing both
the interpreter install and the dependency install with one cache.

No. This only affects CI runner provisioning. End-user-visible
behavior of ``multidict`` is unaffected.

Unlike yarl, multidict computes its runtime Python version in a
separate ``Compute runtime Python version`` step rather than
reading ``steps.python-install.outputs.python-version`` off the
setup action, so no Codecov flag shape changes here.

No. ``astral-sh/setup-uv`` is widely used across aio-libs and the
broader Python ecosystem; ``activate-environment: true`` plus
``uv pip install`` reproduce the behavior the existing pipeline
already relies on.

## Related issue number

N/A; ports https://github.com/aio-libs/yarl/pull/1704 across to
multidict, whose matrix has the same ``3.14t`` entry on macOS and
Windows.

## Checklist

- [x] Unit tests for the changes exist -- N/A (CI configuration)
- [x] Documentation reflects the changes -- N/A (no user-visible change)
- [x] If you provide code modification, please add yourself to ``CONTRIBUTORS.txt``
- [x] Add a new news fragment into the ``CHANGES/`` folder (``1351.contrib.rst``)

<details>

Drafted with Claude Code (Opus 4.7); reviewed by @bdraco.

Direct port of aio-libs/yarl#1704 (merged). Commits mirror the
yarl PR's structure so the diff is easy to compare side-by-side:

| multidict commit | yarl commit |
| --- | --- |
| ``ci: install Python via astral-sh/setup-uv to skip slow downloads`` | ``d10eecb`` |
| ``ci: pin astral-sh/setup-uv to the exact v8.1.0 tag`` | ``07327e0`` |
| ``ci: drop pip cache wrapper, install via uv pip`` | ``d4ab3be`` |

(yarl's intermediate ``Symlink CHANGES/1704.contrib.rst -> 1703.contrib.rst``
commit does not apply here -- multidict has no separate tracking issue,
so the news fragment is named after the PR directly.)

This is a CI-only change; it cannot be exercised against the local
test suite. The full proof is the post-merge CI run on this PR
showing the matrix completing on all three OSes, which the
operator will check before flipping the PR out of draft.

Local verification:

- ``make doc-spelling`` passes for the new fragment (the five
  misspellings reported are pre-existing entries in ``CHANGES.rst``;
  the new ``1351.contrib.rst`` is scanned by
  ``sphinxcontrib-towncrier`` and clean).
- ``pre-commit`` ran on each commit: YAML lint, workflow
  validation, and the changelog/news-fragment role check all pass.

</details>

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -71,8 +71,9 @@ jobs:
       with:
         python-version: ${{ env.PYTHON_LATEST }}
         activate-environment: true
+        enable-cache: true
     - name: Install core libraries for build
-      run: python -Im pip install build
+      run: uv pip install build
     - name: Build sdists and pure-python wheel
       env:
         MULTIDICT_NO_EXTENSIONS: Y
@@ -256,6 +257,7 @@ jobs:
       with:
         python-version: ${{ matrix.pyver }}
         activate-environment: true
+        enable-cache: true
     - name: Compute runtime Python version
       id: python-install
       run: |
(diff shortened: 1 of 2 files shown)


bdraco:
Closing in favor of a fresh PR with a clean single-commit history; the iterative back-and-forth on this branch (initial swap, v8.1.0 pin, pip-cache drop, PYTHON_LATEST fix) is noise. Reopening as a single coherent commit.

PR #1361, opened 2026-06-09
Declare `istr` with `__slots__ = ()` in `_multidict_py.py`

### LINKED ISSUE #1360: Add an empty tuple as `__slots__` to `istr`
Because `istr` is a `str` subclass and strings have no instance dictionaries, and especially in the case of this library where instances of `istr` are used as dictionary keys and many of them are created for large case-insensitive multidicts, it would be beneficial to have a `__slots__ = ()` declaration in the class body of `istr`. This will save the memory allocated to create a dictionary every time a `istr` is constructed to be used as a key.

jonathandung:
## What do these changes do?

Optimizes memory usage by preventing a dictionary to store attributes to be created on every `istr` instance.

Users can no longer set attributes on the keys of the Python version of case-insensitive multidicts, mutable or otherwise.

## Related issue number

Fixes #1360.

## Checklist

- [ ] I think the code is well written

### DIFF
--- a/multidict/_multidict_py.py
+++ b/multidict/_multidict_py.py
@@ -34,7 +34,7 @@ else:
 class istr(str):
     """Case insensitive str."""
 
-    __slots__ = ()
+    __slots__ = ("__istr_identity__",)
 
     __is_istr__ = True
     __istr_identity__: str | None = None
--- a/multidict/_multidict_py.py
+++ b/multidict/_multidict_py.py
@@ -38,7 +38,10 @@ class istr(str):
 
     __is_istr__ = True
     __istr_identity__: str | None
-
+    def __new__(cls, s: Any, /):
+        self = super().__new__(cls, s)
+        self.__istr_identity__ = None
+        return self
 
 _V = TypeVar("_V")
 _T = TypeVar("_T")
@@ -441,10 +444,9 @@ class _CIMixin:
 
     def _identity(self, key: str) -> str:
         if isinstance(key, istr):
-            ret = getattr(key, "__istr_identity__", None)
+            ret = key.__istr_identity__
             if ret is None:
-                ret = key.lower()
-                key.__istr_identity__ = ret
+                key.__istr_identity__ = ret = key.lower()
             return ret
         if isinstance(key, str):
             return key.lower()
--- a/multidict/_multidict_py.py
+++ b/multidict/_multidict_py.py
@@ -38,10 +38,12 @@ class istr(str):
 
     __is_istr__ = True
     __istr_identity__: str | None
+
     def __new__(cls, s: Any, /):
         self = super().__new__(cls, s)
         self.__istr_identity__ = None
         return self
+
 
 _V = TypeVar("_V")
 _T = TypeVar("_T")
--- a/multidict/_multidict_py.py
+++ b/multidict/_multidict_py.py
@@ -441,7 +441,7 @@ class _CIMixin:
 
     def _identity(self, key: str) -> str:
         if isinstance(key, istr):
-            ret = getattr(key, '__istr_identity__', None)
+            ret = getattr(key, "__istr_identity__", None)
             if ret is None:
                 ret = key.lower()
                 key.__istr_identity__ = ret
--- a/multidict/_multidict_py.py
+++ b/multidict/_multidict_py.py
@@ -37,7 +37,7 @@ class istr(str):
     __slots__ = ("__istr_identity__",)
 
     __is_istr__ = True
-    __istr_identity__: str | None = None
+    __istr_identity__: str | None
 
 
 _V = TypeVar("_V")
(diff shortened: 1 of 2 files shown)


multidict/_multidict_py.py
```diff
@@ -34,6 +34,8 @@
 class istr(str):
     """Case insensitive str."""
 
+    __slots__ = ()
```
Vizonex:
```suggestion
    __slots__ = ("__istr_identity__",)
```
@jonathandung the workflows seem to fail because python has ignored that `__istr_identity__` is an editable slot. ```AttributeError: 'istr' object attribute '__istr_identity__' is read-only```

Vizonex (changes requested):
If your going to use slots I would recommend trying to add `__istr_identitiy__` to it. While I like your idea for speeding up pure python the tests reflect the issue.

`AttributeError: 'istr' object attribute '__istr_identity__' is read-only` 

A Way you could slove this one is to try

`__slots__ = ("__istr_identity__",)`

multidict/_multidict_py.py
jonathandung: I see.

jonathandung:
Wait, but it's impossible to define a default value for a slot by setting it on the class, since that overwrites the slot. I changed the relevant logic a bit.

multidict/_multidict_py.py
```diff
@@ -439,7 +441,7 @@ def _key(self, key: str) -> str:
 
     def _identity(self, key: str) -> str:
         if isinstance(key, istr):
-            ret = key.__istr_identity__
+            ret = getattr(key, "__istr_identity__", None)
```
Vizonex: revert to `ret = key.__istr_identitiy__` It is why we are seeing a performance regression in your pull request.
```suggestion
            ret = key.__istr_identity__"
```

Vizonex (changes requested):

multidict/_multidict_py.py
```diff
@@ -37,7 +37,7 @@ class istr(str):
     __slots__ = ("__istr_identity__",)
 
     __is_istr__ = True
-    __istr_identity__: str | None = None
```
Vizonex, line 38: I don't know if this is the route you should be going down. Default values are here for a reason.

multidict/_multidict_py.py
Vizonex: Wait a minute I think I know of something less costly we could try.
```python            
    if not hasattr(key, "__istr_identity__"):
        key.__istr_identity__ = key.lower()
```    
@jonathandung Do you like this improvement idea I can up with? :)

multidict/_multidict_py.py
Vizonex: SEE: https://github.com/aio-libs/multidict/pull/1361#discussion_r3397639105 I think I came up with a better solution.

multidict/_multidict_py.py
jonathandung: `hasattr`, according to the [docs](https://docs.python.org/3/library/functions.html#hasattr), calls `getattr` and catches `AttributeError`, so I don't think that would be any quicker. I couldn't use the default value approach because that overwrites the slot. However, the benefit of my change would be the reduced memory usage, so I think it is for the better.

Vizonex: @jonathandung You have to worry about the performance regression however, you may have to initialize `__istr_idenitiy__ = None` via `__init__` or `__new__` but how that should work is unknown. However I might suggest having a look at the CPython code for clues on how to go about this if you are familiar with the C programming language.

jonathandung: It shouldn't be that hard to compute `lower()`. Since the case-insensitive version of the key is expected to be used, it should be possible to do the following:

```python
class istr(str):
    __slots__ = ()
    def __new__(cls, s, /): return super().__new__(cls, s.lower())
```

or even

```python
class istr(str):
    __slots__ = ()
    def __new__(cls, s, /): return cls.lower(s)
```

but the second solution would be more radical because an instance of `str`, rather than `istr`, is returned.

jonathandung: As for your suggestion, it is not that hard to implement:

```python
class istr(str):
    __slots__ = "__istr_identity__",
    def __new__(cls, s, /):
         self = super().__new__(cls, s)
         self.__istr_identity__ = None
         return self
```

Vizonex: > As for your suggestion, it is not that hard to implement:
> 
> ```python
> class istr(str):
>     __slots__ = "__istr_identity__",
>     def __new__(cls, s, /):
>          self = super().__new__(cls, s)
>          self.__istr_identity__ = None
>          return self
> ```

@jonathandung This is perfect, now be sure it has the `__is_istr__` attribute to it and it should be enough to bypass the performance regression were seeing.

Vizonex:

jonathandung:
I'm not familiar with this benchmarking framework, and I don't understand what it means by "different runtime environments detected".

jonathandung:
I have done all I can, and I think the regression was caused by CodSpeed running on a different virtual CPU. I'm not sure though.

Vizonex:
@webknjaz Any thoughts on this PR?

jonathandung:
Closing in favour of #1386, I suppose. Can't believe I'm losing to AI

Issue #1159, opened 2025-05-04
Segfault calling `MultiDict().setdefault()`


devdanzin:
Using latest git version of 3.13, calling `MultiDict().setdefault("a")` segfaults:

```python
import multidict
m = multidict.MultiDict()
m.setdefault("a")
```

The backtrace references `cpython-3.13.2`, so I tested the repro code against that version  and still got a segfault.

Backtrace:
```console
Program received signal SIGSEGV, Segmentation fault.
0x00007ffff6d38402 in Py_INCREF (op=0x0)
    at /opt/_internal/cpython-3.13.2/include/python3.13/object.h:826
warning: 826    /opt/_internal/cpython-3.13.2/include/python3.13/object.h: No such file or directory

#0  0x00007ffff6d38402 in Py_INCREF (op=0x0)
    at /opt/_internal/cpython-3.13.2/include/python3.13/object.h:826
#1  _pair_list_add_with_hash (hash=-8311172392442909167, value=0x0,
    key=0x555555b2e628 <_PyRuntime+94600>, identity=0x555555b2e628 <_PyRuntime+94600>,
    list=0x555555e8b610) at multidict/_multilib/pair_list.h:360
#2  pair_list_set_default (value=<optimized out>, key=<optimized out>, list=0x555555e8b610)
    at multidict/_multilib/pair_list.h:802
#3  multidict_setdefault (self=0x555555e8b600, args=<optimized out>, nargs=<optimized out>,
    kwnames=<optimized out>) at multidict/_multidict.c:576
#4  0x00005555556815b3 in method_vectorcall_FASTCALL_KEYWORDS (func=0x7ffff6da1850,
    args=0x7ffff7fb0640, nargsf=<optimized out>, kwnames=0x0) at Objects/descrobject.c:420
#5  0x00005555556746fa in _PyObject_VectorcallTstate (tstate=0x555555b604c0 <_PyRuntime+299040>,
    callable=callable@entry=0x7ffff6da1850, args=args@entry=0x7ffff7fb0640,
    nargsf=9223372036854775810, kwnames=kwnames@entry=0x0) at ./Include/internal/pycore_call.h:168
#6  0x00005555556747d6 in PyObject_Vectorcall (callable=callable@entry=0x7ffff6da1850,
    args=args@entry=0x7ffff7fb0640, nargsf=<optimized out>, kwnames=kwnames@entry=0x0)
    at Objects/call.c:327
[...]

PR #1376, opened 2026-07-24
Fix segfault in `add()` with a keyword-only required argument

devdanzin:
## What do these changes do?

`MultiDict().add(key="k")` — a required argument supplied only by keyword, with the other missing —
segfaulted the interpreter instead of raising `TypeError`. The two-argument FASTCALL parser `parse2()`
(`_multilib/parser.h`) did not enforce `minargs` in its keyword branch, so `value` reached the caller as
a NULL pointer that was then dereferenced. Calls with more than two keyword arguments were also
mis-routed into the one-keyword path.

`parse2()` now enforces `minargs` when the only supplied argument comes by keyword, and rejects
more-than-two keyword arguments. The pure-Python backend already raised the correct `TypeError`s for
these calls, so this only brings the C extension in line — no pure-Python change is needed.

Yes: `MultiDict().add(key="k")` and `add(key="k", value="v", extra="e")` now raise `TypeError` (the same
messages the pure-Python implementation raises) instead of crashing / silently misbehaving.

No. A localized fix in one argument-parser function, plus a parametrized regression test.

## Related issue number

Same argument-parser family as the `setdefault()` segfault #1159 (fixed narrowly by #1160); `add()` was
still affected. Surfaced by the C-extension analysis report #1306.

## Checklist

- [x] Unit tests for the changes exist (`test_add_missing_required_argument`, both backends)


Vizonex (changes requested):
@devdanzin The workflows seem to fail along with a segfault my recommendation would be to go through the code carefully. Sometimes I tend to do is use print statements to track where the segfault occurred letting me have real time feedback without needing to bruteforce. When I am done I clear all the print statements I have made and then push the new commit when completed.

multidict/_multilib/parser.h
```diff
@@ -101,13 +100,36 @@ parse2(const char* fname, PyObject* const* args, Py_ssize_t nargs,
...
+                argname = PyTuple_GetItem(kwnames, i);
+                if (argname == NULL) {
+                    return -1;
+                }
+                if (PyUnicode_CompareWithASCIIString(argname, arg1name) != 0 &&
+                    PyUnicode_CompareWithASCIIString(argname, arg2name) != 0) {
+                    return raise_unexpected_kwarg(fname, argname);
+                }
+            }
+            return raise_unexpected_kwarg(fname, argname);
```
asvetlov:
```suggestion
            PyErr_Format(PyExc_TypeError,
                 "%.150s() got more than 2 expected arguments'",
                 fname);
            return -1;
```

multidict/_multilib/parser.h
```diff
@@ -101,13 +100,39 @@ parse2(const char* fname, PyObject* const* args, Py_ssize_t nargs,
...
+                if (argname == NULL) {
+                    return -1;
+                }
+                if (PyUnicode_CompareWithASCIIString(argname, arg1name) != 0 &&
+                    PyUnicode_CompareWithASCIIString(argname, arg2name) != 0) {
+                    return raise_unexpected_kwarg(fname, argname);
+                }
+            }
+            PyErr_Format(PyExc_TypeError,
+                         "%.150s() got more than 2 expected arguments'",
```
asvetlov:
```suggestion
                         "%.150s() got more than 2 expected arguments",
```

asvetlov:
@aiolibsbot rr

multidict/_multilib/parser.h
aiolibsbot: Superseded by your follow-up suggestion. HEAD carries the corrected string, no stray quote. parser.h:133 reads `"%.150s() got more than 2 expected arguments",`.

multidict/_multilib/parser.h
aiolibsbot: Applied. parser.h:133 matches exactly. One note: that error only fires when `kwnames` holds duplicate names, unreachable from Python syntax. Every reachable >2-keyword call exits earlier via `raise_unexpected_kwarg()`.

multidict/_multilib/parser.h
```diff
@@ -101,13 +100,39 @@ parse2(const char* fname, PyObject* const* args, Py_ssize_t nargs,
...
+                argname = PyTuple_GetItem(kwnames, i);
+                if (argname == NULL) {
+                    return -1;
+                }
+                if (PyUnicode_CompareWithASCIIString(argname, arg1name) != 0 &&
+                    PyUnicode_CompareWithASCIIString(argname, arg2name) != 0) {
+                    return raise_unexpected_kwarg(fname, argname);
+                }
+            }
+            PyErr_Format(PyExc_TypeError,
```
asvetlov, line 134:
```suggestion
            // Unreachable from Python code, could be called only for hand-built vectorcall
            PyErr_Format(PyExc_TypeError,
```

CHANGES/1376.bugfix.rst
```diff
@@ -0,0 +1,3 @@
+Fixed a segfault when calling :py:meth:`MultiDict.add() <multidict.MultiDict.add>` (and other two-argument methods) with a required argument supplied only by keyword, e.g. ``d.add(key="k")``. The C argument parser now raises :exc:`TypeError`, matching the pure-Python implementation.
```
asvetlov:
```suggestion
Fixed a segfault when calling :py:meth:`~multidict.MultiDict.add` with only one of its two required arguments supplied by keyword, e.g. ``d.add(key="k")``. Extra keyword arguments passed to the lookup and removal methods are also now rejected with :exc:`TypeError` instead of being silently ignored.
```

asvetlov (approved):

asvetlov:
Thanks, @devdanzin !

PR #1379, opened 2026-07-24
Fix use-after-free / out-of-bounds read from borrowed table pointers in bulk updates

devdanzin:
## What do these changes do?

Two C-extension crashes caused by holding a raw pointer into a hash table across an operation that
reshapes it:

1. `md_update_from_ht()` cached `htkeys_entries(other->keys)` and then inserted into `md`. When `md` **is**
   `other` (`d.extend(d)`, and the update/merge variants), an insert-triggered resize freed the very
   table being iterated — a use-after-free. `extend(self)` now reserves capacity up front (no resize
   during the loop) and snapshots the entry count; `update(self)`/`merge(self)` are short-circuited as
   the no-ops they are.
2. `md_update_from_seq()` cached `PyList_GET_SIZE(seq)` once. A case-insensitive key's `.lower()` can run
   arbitrary Python that shrinks `seq`, after which the stale size let `PyList_GET_ITEM` read past the
   end. The length is now re-read each iteration.

Both reproduce as SIGSEGV on ASan and debug builds with no threads involved.

**Backend parity note (per AGENTS.md):** this fixes the C extension only. The pure-Python backend has a
*related but different* bug — it **hangs** on `d.extend(d)` (it iterates its own list while appending).
I did not fix that here because it's a separate code path in `_multidict_py.py`; happy to send it as a
follow-up so the two backends match (both should double the contents).

Yes: `d.extend(d)` now doubles the contents (was a crash); `d.update(d)`/`d.merge(d)` leave the dict
unchanged (were a crash); building from a list whose key mutates that list no longer reads out of bounds.

No. The changes are confined to two internal bulk-update helpers in `_multilib/hashtable.h`.

## Related issue number

The `extend`/`update`/`==` sites were flagged as data races in the free-threading report #1321; these
particular crashes are reachable single-threaded. Surfaced by the analysis report #1306.

## Checklist

- [x] Unit tests for the changes exist (`test_extend_update_merge_self_reference`, `test_update_from_list_mutated_by_key_lookup`)


Vizonex (approved):
@devdanzin Sorry if I didn't see this sooner. This seems to be a pretty critical bug so I approve getting this PR Merged.

asvetlov:
@copilot resolve the merge conflicts in this pull request

asvetlov (approved):

asvetlov:
Thanks, @devdanzin

PR #1311, opened 2026-03-28
fix ref bug with iter, views and istr

Vizonex:
## What do these changes do?

This is based off previous things that I have referenced in #1310 and #1306. The other bugs listed seem to like much harder puzzles to solve than anything else so I'm tackling the easy fixable ones first.

### Reproducer
```python
"""Reproducer: multidict iterator/view/istr types leak type references.

iter.h:134 — multidict_iter_dealloc: no Py_DECREF(Py_TYPE(self))
views.h:30 — multidict_view_dealloc: no Py_DECREF(Py_TYPE(self))
istr.h:22 — istr_dealloc: no Py_DECREF(Py_TYPE(self))
traverse functions also missing Py_VISIT(Py_TYPE(self)).

Tested: multidict 6.7.1, Python 3.14.
"""
from multidict import MultiDict, istr
import sys

md = MultiDict([("a", "1"), ("b", "2")])

# Test iterator type leak
it = iter(md.keys())
iter_type = type(it)
del it
baseline = sys.getrefcount(iter_type)
for i in range(1000):
    it = iter(md.keys())
    list(it)
    del it
after = sys.getrefcount(iter_type)
print(f"KeysIter type: baseline={baseline}, after={after}, leaked={after-baseline}")

# Test view type leak
view_type = type(md.keys())
baseline = sys.getrefcount(view_type)
for i in range(1000):
    v = md.keys()
    del v
after = sys.getrefcount(view_type)
print(f"KeysView type: baseline={baseline}, after={after}, leaked={after-baseline}")

# Test istr type leak
baseline = sys.getrefcount(istr)
for i in range(1000):
    s = istr("hello")
    del s
after = sys.getrefcount(istr)
print(f"istr type:     baseline={baseline}, after={after}, leaked={after-baseline}")

# Expected: 0 leaked for all
# Actual: exactly 1000 leaked for each — one per object lifecycle
```

## Checklist

- [x] I think the code is well written
- [x] Unit tests for the changes exist
- [x] Documentation reflects the changes

PR #1380, opened 2026-07-24
Fix per-instance type reference leak in the four core types (+ a shutdown use-after-free it masked)

devdanzin:
## What do these changes do?

`MultiDict`, `CIMultiDict`, `MultiDictProxy` and `CIMultiDictProxy` are heap types, so every instance
holds a reference to its type that `tp_dealloc` must release. The iterator, view and `istr` types were
fixed in #1311, but the four core container/proxy types were missed and leaked one type reference per
instance (measurable: +1000 refs per 1000 instances). `tp_dealloc` now caches the type and `Py_DECREF`s
it after `tp_free` (inside the trashcan block for the container, whose `Py_TRASHCAN_END` must stay last).

While fixing this, the leak turned out to be masking a latent **use-after-free at interpreter shutdown**:
`md_clear()` bumped the version counter via `NEXT_VERSION(md->state)` even when called from
`tp_clear()`/`tp_dealloc()`. At finalisation the per-module state can be freed before the last instances,
so that read dereferenced freed memory (caught on an ASan build). The version bump now lives in the
user-facing `clear()` (module guaranteed alive); `md_clear()` no longer touches the module state.
`clear()` still invalidates live iterators. This is a C-extension-only concern (the pure-Python backend
relies on the Python GC); no pure-Python change is needed.

No user-visible behavior change; the reference leak is gone and the shutdown read is fixed.

No. It completes the heap-type-lifecycle fix from #1311.

## Related issue number

Completes #1311 (which covered only the iterator/view/istr types). Surfaced by the analysis report #1306.

## Checklist

- [x] Unit tests for the changes exist (extends `tests/isolated/multidict_type_leak.py` to the four core types)


Vizonex (changes requested):
@devdanzin Ditto of #1376 Where I mentioned spelling checks needed correcting.

asvetlov:
Thank you for the PR, @devdanzin !
The bug is fixed by #1415 (the same changed lines, actually, except for a few different tests).

PR #1382, opened 2026-07-24
Do not swallow non-TypeError exceptions in feature-detection fallbacks

devdanzin:
## What do these changes do?

Several feature-detection fallbacks cleared *every* pending exception when a fallback path was taken:
view richcompare and items-view `__contains__` measuring `len(other)`, items-view `__contains__`
computing the identity, `MultiDict.__eq__` probing `keys()`, and `extend()`/constructor probing
`items()`. A `MemoryError` or `KeyboardInterrupt` raised by the user object was silently discarded. Each
is now guarded with `PyErr_ExceptionMatches()` so only the expected `TypeError`/`AttributeError` is
cleared and anything else propagates — the same pattern the sibling sites in `views.h` already use.
C-extension-only; the pure-Python backend propagates naturally.

Yes (edge case): an exception other than `TypeError`/`AttributeError` (notably `MemoryError`,
`KeyboardInterrupt`) raised by an object being probed now propagates instead of being swallowed. The
normal fallbacks (a non-sized or non-mapping object) are unchanged.

No. It applies the module's existing guard pattern to the few sites that missed it.

## Related issue number

Findings from the C-extension analysis report #1306.

## Checklist

- [x] Unit tests for the changes exist (`test_non_typeerror_exceptions_are_not_swallowed`)


asvetlov (approved):

tests/test_multidict.py
```diff
@@ -1447,3 +1447,41 @@ def test_view_direct_instantiation_segfault() -> None:
...
+        def __len__(self) -> int:
+            raise MemoryError("boom")
+
+    # view richcompare probes len(other)
+    with pytest.raises(MemoryError):
+        md.keys() <= BadLen()  # type: ignore[operator]  # noqa: B015
+
+    # items-view __contains__ probes len(candidate)
+    with pytest.raises(MemoryError):
+        md.items().__contains__(BadLen())
```
asvetlov:
```suggestion
        md.items().__contains__(BadLen())  # type: ignore[operator] 
```

tests/test_multidict.py
```diff
@@ -1447,3 +1447,41 @@ def test_view_direct_instantiation_segfault() -> None:
...
+        md.keys() <= BadLen()  # type: ignore[operator]  # noqa: B015
+
+    # items-view __contains__ probes len(candidate)
+    with pytest.raises(MemoryError):
+        md.items().__contains__(BadLen())  # type: ignore[operator]
+
+    # extend()/constructor probes arg.items()
+    class BadItems:
+        def keys(self) -> list[str]:
+            return ["x"]
```
asvetlov:
```suggestion
            return ["x"]  # pragma: no cover
```

tests/test_multidict.py
```diff
@@ -1447,3 +1447,41 @@ def test_view_direct_instantiation_segfault() -> None:
...
+
+        def items(self) -> object:
+            raise MemoryError("boom")
+
+        def __getitem__(self, key: str) -> int:
+            return 1
```
asvetlov:
```suggestion
            return 1  # pragma: no cover
```

asvetlov:
@devdanzin thanks!

PR #1386, opened 2026-07-30
feat: add __slots__ to istr for lower memory use

### LINKED ISSUE #1360: Add an empty tuple as `__slots__` to `istr`

ltsyk:
## Summary
Add `__slots__` for `istr` so case-insensitive keys do not allocate a per-instance `__dict__` (issue #1360).

Fixes #1360

### DIFF
--- a/CHANGES/1360.feature.rst
+++ b/CHANGES/1360.feature.rst
@@ -1,0 +1,2 @@
+Reduced the memory used by pure-Python ``istr`` instances
+-- by :user:`ltsyk`.
--- a/multidict/_multidict_py.py
+++ b/multidict/_multidict_py.py
@@ -34,10 +34,18 @@ else:
 class istr(str):
     """Case insensitive str."""
 
-    # Avoid per-instance __dict__ when used as dict keys (issue #1360).
     __slots__ = ("__istr_identity__",)
+    __istr_identity__: str | None
 
     __is_istr__ = True
+
+    def __new__(cls, value: object = "") -> Self:
+        self = super().__new__(cls, value)
+        self.__istr_identity__ = None
+        return self
+
+    def __reduce__(self) -> tuple[type[Self], tuple[str]]:
+        return type(self), (str(self),)
 
 
 _V = TypeVar("_V")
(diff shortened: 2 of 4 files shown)


Vizonex (changes requested):
@ltsyk This looks way better than the last PR related to this one that I went through a while back. The workflows are failing however something I would recommend you try and do is run the tests locally when everything is considered as passing.

Vizonex (approved):
Looks good, would recommend you re-run pre-commit locally on your system and make a new commit once fully fixed but other than that this is perfect.

ltsyk:
Thanks for reviewing. I reran the complete local test suite across both backends (1,409 passed), MyPy, and all pre-commit hooks; all pass and produced no changes, so there is no follow-up formatting commit to push. The GitHub Actions matrix is also green; the remaining pre-commit.ci status is its mergeable-check service error rather than a hook failure.

ltsyk:
Thanks @Vizonex — rebased and addressed the remaining pure-Python edge case:

Protocol 0/1 unpickling of older `istr` fixtures bypasses `istr.__new__`, so `__istr_identity__` may be unset. `_identity` now uses `getattr(..., None)` and falls back to computing/caching the identity. Added `test_load_istr_as_cimultidict_key` so loaded fixtures are actually used as `CIMultiDict` keys, not only equality-compared.

Pre-commit / suite re-checked on this tip.

asvetlov (changes requested):
The pure-python `istr` deliberately doesn't use slots for the sake of performance; it is an aware memory-speed tradeoff.
Benchmarks show this choice clearly.
I doubt that we have to change the design decision.
Thoughts?

ltsyk:
Agreed — thanks for the clear design note.

CodSpeed on this branch shows ~10–29% regressions on the pure-Python CI path, which matches the deliberate memory/speed tradeoff you describe. I should not push `__slots__` on pure-Python `istr` against that decision.

Happy to close this PR (or leave it for you to close) so the intentional design stays the source of truth.

asvetlov:
Perhaps it worth a comment with design notes for pure Python istr implementation

PR #1393, opened 2026-08-09
Pin the coverage core so the suite runs on Python 3.14

rodrigobnogueira:
## What do these changes do?

Ask Coverage.py for the `ctrace` measurement core, so the suite runs on
Python 3.14.

Every `3.14` and `3.14t` job on `master` currently fails with one error per
test (3340 of them), while the "Re-run the failing tests with maximum
verbosity" step in the same job reports `1666 passed`. Every open PR inherits
this as soon as it merges `master`.

Coverage chooses its core before the suite starts and avoids `sysmon` when
dynamic contexts are configured, but it only inspects the `dynamic_context`
setting in `.coveragerc`, which this project deliberately leaves unset because
it conflicts with `pytest-cov`. `pytest-cov` switches contexts at runtime
through `Coverage.switch_context()` instead (`--cov-context=test` in
`pytest.ini`), which core selection cannot see. So on 3.14 `sysmon` is picked,
`switch_context()` warns `no-sysmon-context` on every call, and
`filterwarnings = error` turns each warning into an error at test setup.

`ctrace` reports `supports_dynamic_contexts`, and falls back to `pytrace`,
which also supports them, where the C tracer is unavailable. PyPy is therefore
unaffected.

No. This only affects how the test suite measures coverage.

No. One setting in `.coveragerc`, with a comment explaining why it is pinned.
No workflow change.

## Related issue number

None filed. `master` has been red on these jobs since at least 2026-08-06.

## Checklist

- [ ] Unit tests for the changes exist - N/A, this is a test-infrastructure fix
- [ ] Documentation reflects the changes - N/A
- [ ] If you provide code modification, please add yourself to `CONTRIBUTORS.txt` - N/A, no such file

## Verification

Full suite with coverage enabled, exactly as CI invokes it:

| Python | before | after |
|---|---|---|
| 3.14 | errors, one per test | 1661 passed |
| 3.10 | 1661 passed | 1661 passed |

`pre-commit run --all-files` and `make doc-spelling` are clean.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -30,7 +30,14 @@ branch = true
 # `filterwarnings = error` that turns each test into an error. `ctrace`
 # reports `supports_dynamic_contexts`, and falls back to `pytrace` (which
 # also supports them) where the C tracer is unavailable, such as PyPy.
-core = ctrace
+# NOTE: Coveragepy is incompatible with dynamic context that gets enabled via
+# NOTE: `pytest-cov` through a CLI arg to pytest. Refs:
+# * https://github.com/aio-libs/multidict/pull/1393
+# * https://github.com/pytest-dev/pytest-cov/issues/755
+# * https://coverage.rtfd.io/en/latest/changes.html#version-7-15-3-2026-08-02
+# * https://github.com/coveragepy/coveragepy/issues/2200
+# * https://github.com/coveragepy/coveragepy/pull/2234
+core = ${COVERAGE_CORE-ctrace}
 cover_pylib = false
 # https://coverage.rtfd.io/en/latest/contexts.html#dynamic-contexts
 # dynamic_context = test_function  # conflicts with `pytest-cov` if set here
--- a/.coveragerc
+++ b/.coveragerc
@@ -22,6 +22,15 @@ show_missing = true
 
 [run]
 branch = true
+# `pytest-cov` switches contexts through `Coverage.switch_context()` at
+# runtime (see `--cov-context=test` in `pytest.ini`) rather than through the
+# `dynamic_context` setting below. Coverage's core selection only inspects
+# that setting, so on Python 3.14 it picks `sysmon`, which cannot record
+# dynamic contexts, and then warns from every `switch_context()` call. With
+# `filterwarnings = error` that turns each test into an error. `ctrace`
+# reports `supports_dynamic_contexts`, and falls back to `pytrace` (which
+# also supports them) where the C tracer is unavailable, such as PyPy.
+core = ctrace
 cover_pylib = false
 # https://coverage.rtfd.io/en/latest/contexts.html#dynamic-contexts
 # dynamic_context = test_function  # conflicts with `pytest-cov` if set here
--- a/CHANGES/1393.contrib.rst
+++ b/CHANGES/1393.contrib.rst
@@ -1,0 +1,5 @@
+Pinned ``coverage`` to the ``ctrace`` measurement core so the test suite runs
+on Python 3.14. Coverage picks ``sysmon`` there, which cannot record the
+dynamic contexts ``pytest-cov`` switches at runtime, and the resulting warning
+became an error under the suite's ``filterwarnings`` setting
+-- by :user:`rodrigobnogueira`.


webknjaz:
Is this https://github.com/pytest-dev/pytest-cov/issues/755?

rodrigobnogueira:
> Is this [pytest-dev/pytest-cov#755](https://github.com/pytest-dev/pytest-cov/issues/755)?

 Yes, that’s the same issue. I first encountered it in #1317 (https://github.com/aio-libs/multidict/pull/1317)’s Python 3.14 CI. Coverage selects sysmon, but our pytest-cov configuration uses dynamic contexts, so the warning becomes an error. This PR forces ctrace as a repository-level workaround until pytest-cov#755 (https://github.com/pytest-dev/pytest-cov/issues/755) is resolved.

Just saw you opened pytest-cov 755. Do we wait here?

webknjaz:
@rodrigobnogueira I'd like to solve this in a more flexible way that takes into account Cython and could be scaled into other repos by copying w/o modifications. A few of my other repos lost Cython coverage silently, and I never got to working on recovering it. I think Sam mentioned something about selecting a different core to make it work. Do you think you could look into this but start with something outside this repo that does contain Cython? frozenlist, perhaps? Bisect a coveragepy+cpython version combo where it stopped working and help me come up with a strategy for preventing this from happening again by enforcing (a non-zero `*.cyx` coverage in Codecov reports, for example).

This should give us some data points for coming up with a better strategy to roll out across all the repos in our care. Would you be up for something like this?

Vizonex:
> @rodrigobnogueira I'd like to solve this in a more flexible way that takes into account Cython and could be scaled into other repos by copying w/o modifications. A few of my other repos lost Cython coverage silently, and I never got to working on recovering it. I think Sam mentioned something about selecting a different core to make it work. Do you think you could look into this but start with something outside this repo that does contain Cython? frozenlist, perhaps? Bisect a coveragepy+cpython version combo where it stopped working and help me come up with a strategy for preventing this from happening again by enforcing (a non-zero `*.cyx` coverage in Codecov reports, for example).
> 
> This should give us some data points for coming up with a better strategy to roll out across all the repos in our care. Would you be up for something like this?

@webknjaz If were talking about cython support again I think we could bring back full cython support at somepoint the reason I stopped adding in some form of C-API capsule or cython bindings was mainly that many of us had stopped being intrested in supporting it, not to mention the `_testcapi` module was never going to work and that it needed to be in the same directory as the multidict source files. There might be an easier way of adding multidict in that doesn't involve a C-API capsule known as extern where we just need the headerfile of multidict to get it to work properly. Know however that it will not support the Pythonized version of multidict and will only work with the C extension of Multidict. However on the bright side this speeds up parsing and writing http requests by around 30% inside of aiohttp and a workaround for the non-CAPI wouldn't be hard to come up with.

```cython
cdef extern form "multidict.h":
     """you get the idea..."""
```

webknjaz:
In this context, I'm only interested in the coverage configuration at the moment. The capsule thing is a separate idea that I don't have strong opinions on, but I'm fine with in general. I think that it'd be good to also get C-code coverage instrumented before that, though, so it could be tested properly. Still, again, this is a separate topic, and I'd like to scope the conversation here to coverage configuration (coveragepy mainly).

webknjaz:
@rodrigobnogueira I wanted to draw your attention to one thing in the PR description that stands out to me:

> ## Verification
> 
> [..]
> 
> `pre-commit run --all-files` and `make doc-spelling` are clean.

FTR, this is a pattern of what a lot of agents stick into PR bodies — noisy text like internal self-checks that are expected to be implied by a typical development process already but not surfaced to the reviewers. The CI runs these properly, using a project-accepted infra (unlike agents often going for shortcuts). And the fact that you managed to run the in-project infra locally does not really contribute any useful information to the context of the PR, does it?
From the review process standpoint, we're only interested in things that are less obvious and are not a part of tribal knowledge and standard practices. Something that the CI wouldn't run, but you might've hit during exploration, troubleshooting, debugging.
Something adding context to the substance of a PR is useful. Leaking internal of your agent's harness would needlessly pollute the reviewer's context, distract them from what the scope is and delay the merge because it'd take them more time to remember things that actually matter here. Instead, point out that one thing that you intended to communicate regarding the difference so that the reviewer wouldn't be left guessing what that was about.

I've been trying to encode some tribal knowledge for agents in another project as an experiment. Take a look if you're interested: https://github.com/ansible/pylibssh/pull/867. Maybe some of it will be useful for your agent skill files.

.coveragerc
```diff
@@ -22,6 +22,15 @@ show_missing = true
 
 [run]
 branch = true
+# `pytest-cov` switches contexts through `Coverage.switch_context()` at
+# runtime (see `--cov-context=test` in `pytest.ini`) rather than through the
+# `dynamic_context` setting below. Coverage's core selection only inspects
+# that setting, so on Python 3.14 it picks `sysmon`, which cannot record
```
webknjaz, line 28: Does this mean that coveragepy checks the *config* setting called `dynamic_context` somewhere on initialization before `Coverage.switch_context()` is called?

.coveragerc
```diff
@@ -22,6 +22,15 @@ show_missing = true
...
+# dynamic contexts, and then warns from every `switch_context()` call. With
+# `filterwarnings = error` that turns each test into an error. `ctrace`
+# reports `supports_dynamic_contexts`, and falls back to `pytrace` (which
+# also supports them) where the C tracer is unavailable, such as PyPy.
+core = ctrace
```
webknjaz: I think we could attempt something I did @ https://github.com/aio-libs/aiohttp/commit/d83589d196fdab6e55658b2c9d0466403f7b97c3#diff-c2da9b21a113ff072e0c648b0d2acd92f09493e65f9a027de4b3d3bc77d24715R71 and resurrect the default to the previous behavior globally:

```suggestion
core = ${COVERAGE_CORE-ctrace}
```

It can then be set to different values externally, as needed, in `tox.ini`, `Makefile` or similar workflow tools. Something similar has been done in https://github.com/aio-libs/frozenlist/pull/763/changes#diff-662dff05c7dfe6681c1007ae601e8573d413a03f9f9d53d951c2cb99caba4fd4R254.

.coveragerc
```diff
@@ -22,6 +22,15 @@ show_missing = true
...
 [run]
 branch = true
+# `pytest-cov` switches contexts through `Coverage.switch_context()` at
+# runtime (see `--cov-context=test` in `pytest.ini`) rather than through the
+# `dynamic_context` setting below. Coverage's core selection only inspects
+# that setting, so on Python 3.14 it picks `sysmon`, which cannot record
+# dynamic contexts, and then warns from every `switch_context()` call. With
+# `filterwarnings = error` that turns each test into an error. `ctrace`
+# reports `supports_dynamic_contexts`, and falls back to `pytrace` (which
+# also supports them) where the C tracer is unavailable, such as PyPy.
```
webknjaz, line 32: Not sure how PyPy is relevant here but w/o a mechanism to change this setting externally, we'll probably break PyPy test envs.

webknjaz:
@rodrigobnogueira here's what I ended up applying elsewhere —perhaps, we could cherry-pick it here? https://github.com/re-actors/alls-green/commit/8de05dc10f508bf473c58ea66a5ee252fe7e63bd

.coveragerc
asvetlov:
```suggestion
# NOTE: Coveragepy is incompatible with dynamic context that gets enabled via
# NOTE: `pytest-cov` through a CLI arg to pytest. Refs:
# * https://github.com/aio-libs/multidict/pull/1393
# * https://github.com/pytest-dev/pytest-cov/issues/755
# * https://coverage.rtfd.io/en/latest/changes.html#version-7-15-3-2026-08-02
# * https://github.com/coveragepy/coveragepy/issues/2200
# * https://github.com/coveragepy/coveragepy/pull/2234
core = ${COVERAGE_CORE-ctrace}
```

asvetlov:
@webknjaz I've cherry-picked the config from alls-green, as you proposed.
The CI is  green now. 
Let's merge the PR.
If you want to tune the wording, please do it via GitHub suggestion comments.
My point is: the CI on master is broken now; the PR fixes it and should be merged ASAP.

.coveragerc
asvetlov: pypy on CI works well. 
Looks like it uses pytrace as the comment says.

PR #1403, opened 2026-08-27
Clarify that istr compares as a regular str

### LINKED ISSUE #1397: istr is documented as a case-folded string with case-insensitive comparison, but compares as a plain str
﻿### Describe the discrepancy

The documentation for `istr` describes it as a case-folded string with case-insensitive comparison, but in current releases `istr` behaves as a plain `str`: equality and hashing are case-sensitive, and the case-insensitive folding only happens inside `CIMultiDict` internals (via the `__is_istr__` / `__istr_identity__` markers).

Docs (docs/multidict.rst):

- *"Create a new **case-folded** string object from the given object"*
- *"..versionchanged:: 3.7 — `istr` doesn't title-case its argument anymore but uses internal lower-cased data for fast case-insensitive comparison."*

### Reproducer

```python
>>> from multidict import istr, CIMultiDict
>>> istr("Key") == "key"
False
>>> hash(istr("Key")) == hash("key")
False
>>> len({istr("Header"), istr("HEADER"), "header"})
3
>>> str(istr("Content-Type"))
'Content-Type'
```

Case-insensitive lookup still works through the dict itself:

```python
>>> CIMultiDict({"Content-Type": "x"})["content-type"]
'x'
```

### Observed vs documented

| Documented | Actual (6.7.1) |
|---|---|
| "case-folded string object" | original case preserved (`str(istr('Key')) == 'Key'`) |
| "internal lower-cased data for ... comparison" | no `__eq__`/`__hash__` override; direct comparisons are plain-`str` |

Same behavior in 6.1.0 and in the pure-Python fallback (`class istr(str)` carries only marker attributes), so this is long-standing rather than a recent regression.

### Question

Which contract is intended going forward?

1. If the marker-based design (folding handled by `CIMultiDict` only) is final, the `istr` section of the docs could be reworded so users don't expect `istr('Key') == 'key'` or set/dict deduplication by folded value.
2. If the documented per-object semantics are still the goal, then `__eq__`/`__hash__` would need to compare on the folded identity.

Happy to send a docs PR for option 1 if that is the intended reading.

### Versions

- multidict 6.7.1 (C extension) and pure-Python fallback: identical behavior
- Python 3.13

gyanu2507:
Fixes #1397.

The `istr` docs called it a case-folded string and talked about internal lower-cased data used for comparison. Direct `istr` equality is still regular `str` equality though — `istr("Key") == "key"` is False. Folding only happens inside `CIMultiDict`.

Went with the docs wording rather than changing `__eq__`/`__hash__`, since the current behavior has been around since at least 6.1.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -22,6 +22,22 @@ show_missing = true
 
 [run]
 branch = true
+# `pytest-cov` switches contexts through `Coverage.switch_context()` at
+# runtime (see `--cov-context=test` in `pytest.ini`) rather than through the
+# `dynamic_context` setting below. Coverage's core selection only inspects
+# that setting, so on Python 3.14 it picks `sysmon`, which cannot record
+# dynamic contexts, and then warns from every `switch_context()` call. With
+# `filterwarnings = error` that turns each test into an error. `ctrace`
+# reports `supports_dynamic_contexts`, and falls back to `pytrace` (which
+# also supports them) where the C tracer is unavailable, such as PyPy.
+# NOTE: Coveragepy is incompatible with dynamic context that gets enabled via
+# NOTE: `pytest-cov` through a CLI arg to pytest. Refs:
+# * https://github.com/aio-libs/multidict/pull/1393
+# * https://github.com/pytest-dev/pytest-cov/issues/755
+# * https://coverage.rtfd.io/en/latest/changes.html#version-7-15-3-2026-08-02
+# * https://github.com/coveragepy/coveragepy/issues/2200
+# * https://github.com/coveragepy/coveragepy/pull/2234
+core = ${COVERAGE_CORE-ctrace}
 cover_pylib = false
 # https://coverage.rtfd.io/en/latest/contexts.html#dynamic-contexts
 # dynamic_context = test_function  # conflicts with `pytest-cov` if set here
--- a/.coveragerc
+++ b/.coveragerc
@@ -22,15 +22,6 @@ show_missing = true
 
 [run]
 branch = true
-# `pytest-cov` switches contexts through `Coverage.switch_context()` at
-# runtime (see `--cov-context=test` in `pytest.ini`) rather than through the
-# `dynamic_context` setting below. Coverage's core selection only inspects
-# that setting, so on Python 3.14 it picks `sysmon`, which cannot record
-# dynamic contexts, and then warns from every `switch_context()` call. With
-# `filterwarnings = error` that turns each test into an error. `ctrace`
-# reports `supports_dynamic_contexts`, and falls back to `pytrace` (which
-# also supports them) where the C tracer is unavailable, such as PyPy.
-core = ctrace
 cover_pylib = false
 # https://coverage.rtfd.io/en/latest/contexts.html#dynamic-contexts
 # dynamic_context = test_function  # conflicts with `pytest-cov` if set here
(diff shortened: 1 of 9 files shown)


gyanu2507:
The failing checks here are pre-existing on `master`, not from this PR — it only touches `docs/multidict.rst` and a changelog fragment, so it can't affect test results.

Every failure is a 3.14 job, and they all error at setup inside `pytest_cov`:

```
coverage/control.py:799: in switch_context
    self._warn(... slug='no-sysmon-context')
CoverageWarning
```

Coverage's `sys.monitoring` core doesn't support dynamic contexts, and the warning is raised as an error, so every test errors during setup rather than running.

`master` shows the same thing: 4c75c93 has 24 failures (20 of them 3.14), 41c1b91 has 24 (21 on 3.14). It starts at 41c1b91 — 4d10df7 and earlier are fully green.

The coveralls status is just "Coverage remained the same at 85.29%", which is expected for a docs change.

Nothing for me to change here, but happy to rebase once 3.14 is sorted on master.

gyanu2507:
The red jobs here are the Python 3.14 / free-threaded matrix (`Test ubuntu 3.14*`, macos/windows 3.14 ext). The same set is failing on `master` right now (coverage `sys.monitoring` / free-threading setup), and this PR only touches docs + a changelog fragment.

Happy to rebase again once `master` is green on 3.14.

gyanu2507:
The Python 3.14 / free-threaded failures were CoverageWarning from sysmon under pytest-cov dynamic contexts (same breakage on master). I pinned `core = ctrace` in `.coveragerc` on this branch so the matrix can run again. That matches the approach in #1393; happy to drop it from here if that lands first.

Vizonex (approved):
@gyanu2507 I'm not much of a documentation expert but this seems to be going in the right direction so I'll approve this PR. I'm not in charge of merges so I leave that to everybody else to decide. As long as you ensure the workflows behave correctly the other authors will let it pass on through.

gyanu2507:
Coveralls is still red only because coverage stayed at 85.29% (docs-only change; \"Coverage remained the same\" is marked failure). The Python 3.14 matrix is green after the `ctrace` pin. Happy to leave that as-is unless maintainers want a Coveralls config tweak.

.coveragerc
```diff
@@ -22,6 +22,15 @@ show_missing = true
...
 branch = true
+# `pytest-cov` switches contexts through `Coverage.switch_context()` at
+# runtime (see `--cov-context=test` in `pytest.ini`) rather than through the
+# `dynamic_context` setting below. Coverage's core selection only inspects
+# that setting, so on Python 3.14 it picks `sysmon`, which cannot record
+# dynamic contexts, and then warns from every `switch_context()` call. With
+# `filterwarnings = error` that turns each test into an error. `ctrace`
+# reports `supports_dynamic_contexts`, and falls back to `pytrace` (which
+# also supports them) where the C tracer is unavailable, such as PyPy.
+core = ctrace
```
asvetlov:
```suggestion
```

gyanu2507: Dropped. `.coveragerc` matches `master` after #1393 (the `core = ${COVERAGE_CORE-ctrace}` pin). This PR no longer touches that file.

asvetlov (approved):

asvetlov:
Thanks!

PR #1407, opened 2026-09-03
Enable strict mypy checks

asvetlov:
Fix existing mypy errors as a side effect

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -131,6 +131,7 @@ repos:
     name: MyPy, for Python 3.13
     additional_dependencies:
     - types-docutils
+    - types-objgraph
     - lxml  # dep of `--txt-report`, `--cobertura-xml-report` & `--html-report`
     - pytest >= 8.4.0
     - pytest_codspeed
@@ -149,6 +150,7 @@ repos:
     name: MyPy, for Python 3.11
     additional_dependencies:
     - types-docutils
+    - types-objgraph
     - lxml  # dep of `--txt-report`, `--cobertura-xml-report` & `--html-report`
     - pytest >= 8.4.0
     - pytest_codspeed
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -124,7 +124,7 @@ repos:
   - id: clang-format
     types_or: [c, c++]
 - repo: https://github.com/pre-commit/mirrors-mypy.git
-  rev: v1.15.0
+  rev: v2.3.1
   hooks:
   - id: mypy
     alias: mypy-py313
--- a/mypy.ini
+++ b/mypy.ini
@@ -4,5 +4,5 @@ strict = True
 show_error_codes = True
 warn_unused_configs = True
 ignore_missing_imports = True
-files = multidict tests
+files = multidict,tests
 
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,9 +1,0 @@
-[mypy]
-# Strict mode (recommended)
-strict = True
-show_error_codes = True
-warn_unused_configs = True
-ignore_missing_imports = False
-files = multidict,tests
-pretty = True
-
--- a/mypy.ini
+++ b/mypy.ini
@@ -4,4 +4,5 @@ strict = True
 show_error_codes = True
 warn_unused_configs = True
 ignore_missing_imports = True
+files = multidict tests
 
--- a/mypy.ini
+++ b/mypy.ini
@@ -5,6 +5,3 @@ show_error_codes = True
 warn_unused_configs = True
 ignore_missing_imports = True
 
-[mypy-tests.*]
-# Tests often contain untyped helper code; relax strict checks for tests
-ignore_errors = True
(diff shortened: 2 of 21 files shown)


mypy.ini
webknjaz, line 1: This is the wrong config file to be editing. The correct one here is `.mypy.ini`.

mypy.ini
```diff
@@ -0,0 +1,9 @@
+[mypy]
+# Strict mode (recommended)
```
webknjaz: This comment is not informative so let's leave it out.

In general, I'm trying to keep the configs annotated and structured like so: https://github.com/ansible/awx-plugins/blob/93d0e55519f63e7de710e48b0897b60be6459da3/.mypy.ini#L1-L59.

tests/test_incorrect_args.py
```diff
@@ -55,8 +55,8 @@ def test_getall_args(
 ) -> None:
     with pytest.raises(TypeError, match=r".*argument.*"):
         multidict_object.getall(
-            *tested_method_args.positional,
-            **tested_method_args.keyword,
+            *cast(Any, tested_method_args.positional),
```
webknjaz: Will this work with casting to `object`? MyPy recommends using it in place of `Any` whenever possible.

Additionally, we could selectively disable this sort of checks that intentionally passes non-compliant types. This sounds like a justified case for either type ignored (with annotated justification comments) or dedicated config sections like https://github.com/ansible/awx-plugins/blob/93d0e55519f63e7de710e48b0897b60be6459da3/.mypy.ini#L109-L122.

webknjaz (changes requested):
Use the correct config (if really needed), rethink how to handle illegal arg types in tests.

mypy.ini
asvetlov: removed

mypy.ini
asvetlov: moved to pyproject.toml

tests/test_incorrect_args.py
asvetlov: Well, suppressing specific errors for the test file looks better to me. Thanks for the suggestion!

webknjaz (changes requested):
Do not delete `.mypy.ini`. It makes it difficult to sync changes across dozens of projects having the same config structure using `diff`.

mypy.ini
webknjaz: Don't do this. Use the established cross-project pattern that allows scaling config management practices by using the same file names and templates. Keeping `.mypy.ini` is a must for now. Migrating to a different file would involve a bigger coordinated multiproject effort which is probably out of the scope here.

tests/test_incorrect_args.py
```diff
@@ -55,8 +55,8 @@ def test_getall_args(
 ) -> None:
     with pytest.raises(TypeError, match=r".*argument.*"):
         multidict_object.getall(
-            *tested_method_args.positional,
-            **tested_method_args.keyword,
+            tested_method_args.positional,
```
webknjaz: I think this changes the intended semantics of the test so I'll let @bdraco decide if this is correct. FWIW, this looks weird to me.

tests/test_incorrect_args.py
webknjaz, line 1: The more I look into this the more I understand that the changes to this test module are plain wrong and must be reverted.

.pre-commit-config.yaml
webknjaz, line 1: The MyPy version bump could be a standalone PR as it's an atomic change that's noncontroversial. It might need to include some coupled changes if there are related violations popping up with it but this doesn't seem to be related to the declared goal of achieving strictness.

pyproject.toml
webknjaz, line 1: I mentioned above that we need to go back to the previous shape of the `.mypy.ini` until the convention changes but also, I'd like to record for the future that this patch seems to have lost a number of important toggles that were previously present in the config and are now nowhere to be found — they are not related to MyPy's strictness directly — some control how the output is presented or enable additional checks that aren't otherwise on by default.

tests/test_incorrect_args.py
asvetlov: agree, my error

pyproject.toml
asvetlov: reverted to `.mypy.ini`

.pre-commit-config.yaml
asvetlov: Make sense, I'll do it in a separate PR

Dreamsorcerer: I assume all the remaining changes here are from the version bump though. The strict argument is pointless, it just changes the defaults for several options which we already specify explicitly (and stricter than what strict does). Which also means the description of the PR/changelog is misleading, we're not actually changing any strictness here..

asvetlov: @Dreamsorcerer looks like you are right; there are no new type errors after adding `string=True`.
Honestly, the almost infinite list of mypy flags scares me; I can never remember which options belong to `strict` and which don't. 
Newly added options are another challenge.

Anyway, this PR should be closed. Thank you all for the review.

Dreamsorcerer: > Newly added options are another challenge.

I'm subscribed to the release log and update them. At some point I'll figure out a central org-wide config to keep everything synced.

PR #1406, opened 2026-09-03
Use faster identity comparison function

asvetlov:
It is equal to PyUnicode_Equal() from Python 3.14+ but without type checks, arguments are always strings.

PR #1411, opened 2026-09-05
Add nonascii benchmark

asvetlov:
The test needs a separate PR to measure the effect in #1406


asvetlov:
The change is not user-faced, skip changenote

bdraco (approved):

webknjaz:
This could've been a 4-line change with`@pytest.mark.parametrize`, it seems.

PR #1426, opened 2026-09-08
Mark temporarily-removed entries via the hash's high bit, not -1

asvetlov:
## What do these changes do?

Both backends marked a hash table entry invalid mid-update by overwriting its hash with `-1`, then recomputed it from the identity to restore it. This replaces that with a marker bit (the hash's high bit): OR to set, AND to clear. That needs every hash non-negative first, so hashes are folded once, right where computed (`& MAXSIZE` in Python, inside `_unicode_hash()` in C). Restoring is then a bit-clear or a direct assignment of an already-known value, never a recomputation.

Python's `hash()` is arbitrary precision, so a negative hash already reads the high bit as set, which is why plain OR/AND needs the fold first. C's `Py_hash_t` is fixed-width, so once folded, "is marked" is just `entry->hash < 0`. `_unicode_hash()` folds only the value it returns to multidict, never the unicode object's own cached hash slot (shared process-wide). The ~15 `_unicode_hash(...) == -1` checks for a genuine hash-computation error are untouched — a different, CPython-standard meaning of `-1`.

Also adds a test for a related previously-uncovered path: a multidict already at `update()`'s size-estimate cap needing a resize mid-`update()`, which no existing test reached.

CodSpeed flagged a transient regression on `getall` partway through this PR (a read-modify-write per marked entry where the `-1` sentinel only wrote a constant); fixed by assigning known values directly and hoisting the loop-invariant mark value out of the hot loops. Latest CodSpeed run shows a **12.98% improvement**, including measurable C-extension `getall` gains.

No. Internal implementation detail of both backends.

No.

## Related issue number

N/A

## Checklist

- [x] Unit tests for the changes exist (existing suite covers these paths; added a test for the previously-uncovered resize-during-update path)

<details>

Tests: `pytest -q` (C extension, 1802 passed) and `MULTIDICT_NO_EXTENSIONS=1 pytest -q --no-c-extensions` (pure Python, 914 passed), plus a full run under a debug build (`-UNDEBUG`) so `_md_check_consistency()`'s assertions verify every mutation.

Verified with a stress script exercising every mutating method against a deliberate mix of negative- and positive-hash keys, since normal runs only hit negative hashes by chance of `PYTHONHASHSEED`.

Compiled clean with `-Werror -Wsign-compare -Wconversion -std=c11` in both release and debug modes.

`make doc-spelling`: passes for this change; one pre-existing, unrelated failure on `master` (a "fallbacks" false positive) is not introduced by this PR.

</details>

### DIFF
--- a/CHANGES/1426.misc.rst
+++ b/CHANGES/1426.misc.rst
@@ -1,0 +1,5 @@
+In the pure-Python implementation, a temporarily removed hash table entry
+is now marked by flipping the high bit of its hash instead of overwriting
+it with a sentinel value, so restoring the entry no longer needs to
+recompute the hash
+-- by :user:`asvetlov`
--- a/CHANGES/1426.misc.rst
+++ b/CHANGES/1426.misc.rst
@@ -1,5 +1,5 @@
-In both the pure-Python and C implementations, a temporarily removed hash
-table entry is now marked by setting the high bit of its hash instead of
+Changed both the pure-Python and C implementations to mark a temporarily
+removed hash table entry by setting the high bit of its hash instead of
 overwriting it with a sentinel value, so restoring the entry no longer
-needs to recompute the hash
+recomputes the hash
 -- by :user:`asvetlov`
--- a/CHANGES/1426.misc.rst
+++ b/CHANGES/1426.misc.rst
@@ -1,5 +1,5 @@
-In the pure-Python implementation, a temporarily removed hash table entry
-is now marked by flipping the high bit of its hash instead of overwriting
-it with a sentinel value, so restoring the entry no longer needs to
-recompute the hash
+In both the pure-Python and C implementations, a temporarily removed hash
+table entry is now marked by setting the high bit of its hash instead of
+overwriting it with a sentinel value, so restoring the entry no longer
+needs to recompute the hash
 -- by :user:`asvetlov`
--- a/CHANGES/1426.misc.rst
+++ b/CHANGES/1426.misc.rst
@@ -1,5 +1,5 @@
 Changed both the pure-Python and C implementations to mark a temporarily
 removed hash table entry by setting the high bit of its hash instead of
 overwriting it with a sentinel value, so restoring the entry no longer
-recomputes the hash
+recomputes the hash; :meth:`~multidict.MultiDict.getall` is faster ~10% now in C version
 -- by :user:`asvetlov`
--- a/multidict/_multidict_py.py
+++ b/multidict/_multidict_py.py
@@ -569,7 +569,7 @@ class _HtKeys(Generic[_V]):
             else:
                 assert not (hash_ & HASH_MARK)
             i = hash_ & mask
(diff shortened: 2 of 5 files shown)


asvetlov:

asvetlov:
@aiolibsbot rr

multidict/_multidict_py.py
```diff
@@ -30,6 +30,17 @@
...
+MAXSIZE = sys.maxsize
+
+# hash() never returns a value outside [-(MAXSIZE + 1), MAXSIZE]. XORing a
+# real hash with its highest bit always pushes the result outside that
+# range, so it can be used to mark an entry's hash as temporarily invalid
+# (its slot is being processed) without losing the original bits: XORing
+# the same bit again restores the exact original hash, no recomputation
+# needed. A plain OR/AND wouldn't do: Python ints are arbitrary precision,
+# so any negative hash already reads that bit as set.
+HASH_MARK = MAXSIZE + 1
```
Vizonex, line 38: Explanation makes perfect sense. We wouldn't want to use a value that an object could be theoretically hashed into as it would confuse the hash-table itself such as `sys.maxsize`.

Vizonex (approved):
Over all not really much for me to say and I hope this approach gets added in and I like your idea of using a `HASH_MARK` instead of `sys.maxsize`. Decided to review for the fun of doing so as programming and patching bugs when I started contributing here has always interested me. While I may not be as familiar with XOR operations, I see the intent of putting the hashed number out of bounds and then putting them back into bounds

asvetlov:
@Vizonex, yes, the idea is to use the highest hash bit as a marker. 
It should significantly speed up the cleanup stage: instead of hash re-calculation the system could just reset the bit back. XOR is required on pure-Python side; in C we can just cast to unsigned int.
Let me double-check; maybe regular INT/OR arithmetic works well too.

asvetlov:
@Vizonex, we finally have a visible speedup in C `getall`, as I expected when I started this PR.

multidict/_multilib/hashtable.h
```diff
@@ -36,6 +36,9 @@ typedef enum _UpdateOp {
     Merge,
 } UpdateOp;
 
+/* The hash range's high bit; see the marking scheme described below. */
```
```suggestion
```

CHANGES/1426.misc.rst
```diff
@@ -0,0 +1,5 @@
+Changed both the pure-Python and C implementations to mark a temporarily
+removed hash table entry by setting the high bit of its hash instead of
+overwriting it with a sentinel value, so restoring the entry no longer
+recomputes the hash
```
asvetlov:
```suggestion
Changed both the pure-Python and C implementations to mark a temporarily
removed hash table entry by setting the high bit of its hash instead of
overwriting it with a sentinel value, so restoring the entry no longer
recomputes the hash; :meth:`~multidict.MultiDict.getall` is faster ~10% now in C version
```

PR #1434, opened 2026-09-09
Protect pure-Python update()/extend()/merge() with a per-instance lock

asvetlov:
## What do these changes do?

`update()`/`merge()` in the pure-Python `MultiDict` mark a replaced
duplicate entry by nulling its key and value, then clean the table up
afterwards in `_post_update()`. That leaves the table in an
externally-visible intermediate state across several statements. On
the free-threaded build, a concurrent thread reading the same
multidict (as the argument to its own `update()`/`extend()`/`merge()`
call, or mutating it at the same time) could observe a table slot
whose index still points at an already-nulled-out entry, raising
`AttributeError: 'NoneType' object has no attribute 'hash'`.

Built-in containers get automatic per-object locking under
free-threading, but plain Python classes do not. This adds a
`threading.Lock` per `MultiDict` instance and holds it (plus the
argument's lock too, when the argument is a distinct multidict or
proxy, always ordered by `id()` to avoid deadlock) for the whole
extent of `update()`/`extend()`/`merge()`.

This is a pure-Python-only fix. The C extension has its own, separate
critical-section fix for the equivalent bug, tracked independently;
this PR does not depend on it and touches only
`multidict/_multidict_py.py`.

No observable behavior change on the default (GIL) build. On a
free-threaded build, `update()`/`extend()`/`merge()` on the
pure-Python backend no longer risk crashing under concurrent access.

No. The lock is only acquired inside `update()`/`extend()`/`merge()`;
every other method is untouched.

## Related issue number

N/A, no linked issue.

## Checklist

<details>

Reproduced the crash against a free-threaded CPython 3.13.7t build
(pyenv) and confirmed the fix there. `pytest -q` (both `--c-extensions`
and `--no-c-extensions` runs): 1803 passed. `mypy --config-file
.mypy.ini`: no new errors (pre-existing unrelated errors in
`tests/isolated/*.py` and `docs/conf.py`). `ruff check` / `ruff
format --diff`: clean. `make doc-spelling`: one pre-existing failure
(`fallbacks` in the compiled `CHANGES.rst`) reproduces identically on
a clean `upstream/master` checkout; unrelated to this change and not
touched here.

The new regression test
(`tests/test_update.py::test_pure_python_update_extend_merge_thread_safety`)
imports `multidict._multidict_py.MultiDict` directly (matching the
existing pattern in this file) rather than gating on a C-extension
marker, so it also exercises the race under the plain GIL build,
where it fails reliably against the pre-fix code and passes reliably
against the fix.

</details>

Drafted with Claude Code (Sonnet 5); reviewed by andrew.svetlov@gmail.com.


asvetlov:
Superseded by #1447

PR #1435, opened 2026-09-10
Fix a use-after-free in _md_resize()/md_clone_from_ht() under free threading

asvetlov:
Stacks on top of #1433 (not yet merged): the diff below will include
those commits until it lands, and will narrow to just the two commits
on this branch afterward.

## What do these changes do?

Fixes a residual use-after-free in `_md_resize()` and
`md_clone_from_ht()` (`multidict/_multilib/hashtable.h`) found while
stress-testing #1433's locking. Both functions cached a keys-table
pointer or its size, called `PyMem_Malloc()` for a new table, and then
used the cached value again afterward. On the free-threaded build, a
`PyMem_Malloc()` call that blocks acquiring an internal allocator lock
transiently suspends the calling thread's critical section (CPython
suspends critical sections around blocking lock acquisitions; see
`Include/cpython/critical_section.h`), letting a concurrent
update()/extend()/merge()/copy() on the same instance run to completion
and free or replace the cached table in the meantime. `.copy()` calls
`PyMem_Malloc()` on every invocation, which made it easy to hit under
concurrent `update()`-vs-`copy()` stress.

Both functions now re-read the table pointer, its size, and the entry
count only after the allocation returns, and retry (discarding the
stale allocation) if a concurrent operation invalidated the assumptions
the allocation size was based on. `_md_shrink()` doesn't allocate, so
it isn't affected.

No behavioral change for single-threaded code. On the free-threaded
build under concurrent load, this fixes a use-after-free that could
corrupt a `MultiDict`'s internal state or crash the process.

No. The change is confined to two functions already covered by the
existing critical-section locking; no new public API.

## Related issue number

Related to #1433 (no separate issue filed).

## Checklist

<details>

Root cause was confirmed against CPython's own source (`Python/pystate.c`,
`Python/lock.c`, `Include/cpython/critical_section.h`): a contended
`PyMutex_Lock()` uses `_PY_LOCK_DETACH`, and a detach calls
`_PyCriticalSection_SuspendAll()`. The free-threaded allocator hits a
contended `PyMutex` under concurrent allocation pressure (the
interpreter's deferred-free queue mutex in `Objects/obmalloc.c`), which
is enough to trigger this in practice.

Reproduced by widening the race window: temporarily wrapped the
`PyMem_Malloc()` calls with `Py_BEGIN_ALLOW_THREADS`/`nanosleep()`/
`Py_END_ALLOW_THREADS` (a real detach/attach, unlike a plain sleep,
which does not suspend the critical section) to simulate a slow
allocator lock. This reproduced the reported
`Assertion 'entry->hash >= 0' failed` in `_md_check_consistency()` on
the first run for concurrent `update()`-vs-`update()` and
`update()`-vs-`copy()`, and a segfault for `merge()`-vs-`merge()` and
`update()`-vs-constructor-clone. After the fix, 25+ runs with the same
widened window (up to 3ms per allocation, versus a real window on the
order of microseconds) show zero failures, on top of 1800+ passing
tests and 60+ natural-timing stress runs with no injected delay.

Tests: `pytest tests/` passes on all three backends exercised --
free-threaded debug build (`3.14.7t`, `MULTIDICT_DEBUG_BUILD=1`, 1807
passed / 5 skipped), the default C-extension build on `3.13` (1812
passed), and the pure-Python build (`MULTIDICT_NO_EXTENSIONS=1`, 1812
passed, unaffected since this is a C-extension-only fix).

Lint: `make doc-spelling` fails on this branch, but on an unrelated,
pre-existing word ("fallbacks" in `CHANGES.rst:93`, from an already
released changelog entry predating this branch) confirmed present on
both `master` and `bulk-critical-sections` -- not introduced here. This
PR's own news fragment passes the spell check cleanly.

</details>


asvetlov:
Cherry-picked into #1441. Closing in favor of that PR.

PR #1436, opened 2026-09-10
Implement tp_vectorcall for MultiDict, CIMultiDict and their proxies

asvetlov:
## What do these changes do?

`MultiDict(...)`, `CIMultiDict(...)`, `MultiDictProxy(...)` and `CIMultiDictProxy(...)` now go through a dedicated `tp_vectorcall` function in the C extension instead of the classic `tp_new`/`tp_init` pair, avoiding the positional-args tuple and keyword-args dict that `type.__call__()` otherwise builds for every construction call. The new path reuses the same `md_init()`/`md_update_from_*()` primitives as the existing constructors, so behaviour is unchanged.

Since the object under construction is freshly allocated and not yet reachable from any other thread, only an existing multidict, proxy, or plain dict argument needs a critical section during construction, not `self`.

The pure-Python implementation is unaffected: there is no equivalent C-level calling protocol to opt into there, so no change was needed on that side.

No. Construction is faster in the common case; results, error messages, and subclassing behaviour (a `__new__`/`__init__` override in a Python subclass still runs, since the fast path is not inherited by subclasses) are unchanged.

No. The new code is a self-contained addition that shares its core logic with the existing constructors rather than duplicating it, and does not change any existing code paths.

## Related issue number

N/A

## Checklist

<details>

Tests: `pytest -q` (1831 passed) against the default C-extension build; `MULTIDICT_NO_EXTENSIONS=1 pip install ...` + `pytest -q --no-c-extensions` (926 passed, 905 deselected) for the pure-Python leg; free-threaded debug build (3.13, `-O0 -g3`, `PYTHON_GIL=0`) via `pytest -q -k "not benchmark"` (1557 passed, 5 skipped).

Concurrency: a 16-thread stress test constructing `MultiDict`/`CIMultiDict`/`MultiDictProxy`/`CIMultiDictProxy` from a shared, concurrently-read source found no crashes or corruption. A separate stress test that also *mutates* the shared source concurrently does reproduce a segfault, but it reproduces identically on unmodified `master`; it's the already-tracked residual free-threading race, not a regression from this change.

Lint: `pre-commit run --all-files` passes (ruff check/format, clang-format, mypy for 3.11 and 3.13). `make doc-spelling` passes for the new content (the word `vectorcall` was added to `docs/spelling_wordlist.txt`); one pre-existing, unrelated spelling failure already exists on `master` in `CHANGES.rst` and is untouched by this PR.

</details>

### DIFF
--- a/CHANGES/1436.feature.rst
+++ b/CHANGES/1436.feature.rst
@@ -1,5 +1,5 @@
 ``MultiDict``, ``CIMultiDict``, ``MultiDictProxy`` and ``CIMultiDictProxy`` in the
-C extension now implement the vectorcall calling convention for construction,
-speeding up ``MultiDict(...)``-style calls by avoiding an intermediate
+C extension implemented the vectorcall calling convention for construction.
+This sped up ``MultiDict(...)``-style calls by avoiding an intermediate
 positional-arguments tuple and keyword-arguments dictionary in the common case
 -- by :user:`asvetlov`.
--- a/CHANGES/1436.feature.rst
+++ b/CHANGES/1436.feature.rst
@@ -1,0 +1,5 @@
+``MultiDict``, ``CIMultiDict``, ``MultiDictProxy`` and ``CIMultiDictProxy`` in the
+C extension now implement the vectorcall calling convention for construction,
+speeding up ``MultiDict(...)``-style calls by avoiding an intermediate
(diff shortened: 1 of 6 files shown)


asvetlov:
Ignore getall flaky benchmarks degradation

PR #1437, opened 2026-09-11
Fetch list items as new references on free-threaded builds

rodrigobnogueira:
## What do these changes do?

Bulk updates from a list and `[key, value] in md.items()` borrowed items from a list the caller still owns and increfed them in a separate step. On a free-threaded build another thread can drop the item in between, or shrink the list between its length check and the reads, and the C extension reads freed memory.

Each item is now taken as a new reference with `PyList_GetItemRef` on free-threaded builds, which fetches atomically and locks the list only if its lock-free attempt fails; GIL builds keep the `PyList_GET_ITEM` macro, so their hot paths are unchanged. Follows #1379, which fixed the stale length in the same loop.

A critical section on the list, as #1433 takes on a dict argument, would not do: it covers neither a `[key, value]` item inside the list nor the probe handed to `in`, and it is released whenever the thread detaches, which a key's `__hash__` or `__eq__` can cause mid-loop. A strong reference per fetch is what actually holds.

On free-threaded builds only: a call that finds a list item gone after the list's length was checked now raises `RuntimeError("list changed size during iteration")`, from `update()`, a `[key, value]` item and `[key, value] in md.items()` alike. A list that is simply shorter by the next iteration of `update()` still ends the update there, as before (#1379) and as in a plain `for` loop.

No. One helper in `hashtable.h`, used at three sites.

## Related issue number

None.

## Checklist

- [ ] Documentation reflects the changes -- N/A, no API change
- [ ] If you provide code modification, please add yourself to `CONTRIBUTORS.txt` -- N/A, no such file


multidict/_multilib/hashtable.h
```diff
@@ -2459,6 +2459,28 @@ _err_cannot_fetch(Py_ssize_t i, const char* name)
...
+   atomically (locking the list only if its lock-free attempt fails) and
+   _list_item_gone() reports an item that is no longer there. GIL builds
+   keep the macro and compile the check away. */
+#ifdef Py_GIL_DISABLED
+static inline PyObject*
+_list_getitem_ref(PyObject* list, Py_ssize_t i)
+{
+    PyObject* item = PyList_GetItemRef(list, i);
+    if (item == NULL) {
+        PyErr_Clear();
```
asvetlov, line 2635: Why is the error suppressed?
Isn't it better to propagate an error about a list modification (probably by another thread) during iteration?

asvetlov (changes requested):

multidict/_multilib/hashtable.h
rodrigobnogueira: Addressed. The old helper cleared the IndexError and each caller papered over the vanished item. A miss after the length check can only mean another thread shrank the list, so it now becomes RuntimeError("list changed size during iteration") and all three sites propagate it; other errors pass through untouched. The #1379 per-iteration length check is unchanged.

asvetlov (approved):

asvetlov:
Thanks @rodrigobnogueira

Issue #1432, opened 2026-09-09
Support genuine free-threaded mode


asvetlov:
This complex task covers many C functions for MultiDict (and family) methods and slots.

Some of them are trivial and could be solved by wrapping the call in a critical section, for example, ' repr '.

But applying critical sections to all methods kills performance.

The pragmatic approach should combine critical sections with atomic operations, as CPython does for `dict`.

I expect a bunch of PRs for this issue. Let's eat the elephant piece by piece.

Vizonex:
@asvetlov I will be on vacation for bit so the most I can do is reviews but hopefully next week I could start making my own prs to help with this cause.

asvetlov:
@Vizonex I guess most of the work is done already.
I really appreciate the review; maybe we still have some bugs and missing parts.

asvetlov:
Closing the issue. The FT mode is supported.
If anybody will find a bug -- it should be reported and processed on a regular basis.

PR #1438, opened 2026-09-11
Protect MultiDict single-item ops and view set-algebra with a critical section

asvetlov:
## What do these changes do?

Extends the free-threaded-build locking from #1431/#1433 to the operations
that were still running without holding `self`'s lock at all: `add()`,
`__setitem__`/`__delitem__`, `get()`/`getone()`/`__getitem__`, `__contains__`,
`getall()`, `setdefault()`, `pop()`/`popone()`/`popall()`/`popitem()`,
`__eq__`, iteration (`keys()`/`items()`/`values()`), and the
`&`/`|`/`-`/`^`/`in`/`isdisjoint()` operations on `.keys()` and `.items()`
views. A resize triggered by one thread's mutation could free the hash
table while a concurrent, unprotected reader on another thread was still
walking it.

The view set-algebra methods (`itemsview`/`keysview` `and`/`or`/`sub`/
`contains`/`isdisjoint`) use the same `_lock_held`-style split CPython's
own `dictobject.c` uses: the existing body is extracted into a `_impl`
function, and a thin wrapper acquires the critical section around it,
since several of these have multiple `goto`/return exits that can't be
wrapped with `Py_BEGIN_CRITICAL_SECTION`/`Py_END_CRITICAL_SECTION` in
place.

This also fixes a compile error in the in-progress atomic-reads branch
this stacks on: `htkeys.h` referenced `atomic_load_ssize_relaxed()`
without including the header where it's defined. That header (renamed
`atomic_helpers.h`) now selects its atomic-load backend the same way
CPython's `Include/cpython/pyatomic.h` does: GCC/Clang builtins by
default, C11 `stdatomic.h` next, and MSVC intrinsics only as the last
resort for `cl.exe`, which supports neither.

`len()` stays lock-free (a relaxed atomic read of `md->used`), matching
CPython's own `dict_length()`, which is also unlocked even though the
rest of `dictobject.c` is fully critical-section-protected.

Lock-free reads are intentionally out of scope here. CPython's dict
gets its `get()`/`__contains__`/iteration speed back with a separate,
validated `_Py_dict_lookup_threadsafe()` path (optimistic lookup with a
retry-on-concurrent-mutation check); that's a distinct follow-up, not
this PR.

No behavior change for single-threaded code or the default GIL build
(`Py_BEGIN_CRITICAL_SECTION` compiles to a no-op there). On the
free-threaded build, these operations no longer race a concurrent
resize.

No. The change only adds locking around existing logic and reworks an
internal-only header; no new public API, and no behavior change on the
GIL build.

## Related issue number

#1432

## Checklist

<details>

Tests: `pytest -q` on the default GIL build (1833 passed), on a
free-threaded `3.13.7t` release build (1828 passed, 5 skipped), and on
a `MULTIDICT_DEBUG_BUILD=1` free-threaded build (1828 passed, 5
skipped). Re-verified on the GIL and free-threaded release builds
again after the `atomic_helpers.h` rework.

Two new regression tests were added
(`test_single_item_ops_thread_safety`, `test_view_set_ops_thread_safety`)
and verified against the unpatched code:
`test_single_item_ops_thread_safety` reliably crashed the
free-threaded interpreter (within 3 runs) before this fix, and passes
10/10 after. `test_view_set_ops_thread_safety` passes reliably after
the fix, but I could not reproduce a crash on the unpatched code in 40
pytest runs, a dedicated 32k-operation stress script, or under the
debug build -- likely because this pyenv build's allocator doesn't
poison freed memory, so the use-after-free (same unlocked-dereference
shape as the fixed single-item path) often reads stale-but-coherent
bytes rather than crashing. Flagging this so the weaker empirical
signal on that one test is not overstated. Both tests use
`contextlib.suppress()` rather than a bare `try`/`except: pass` for
the racing-with-concurrent-mutation exceptions they intentionally
tolerate.

`make doc-spelling` currently fails on a pre-existing misspelling
("fallbacks") in an already-released `CHANGES.rst` entry, unrelated to
this PR's fragment; `towncrier build --draft` renders this PR's
`CHANGES/1438.bugfix.rst` fragment cleanly on its own.

Drafted with Claude Code (Sonnet 5); reviewed by andrew.svetlov@gmail.com.
</details>


tests/test_multidict.py
```diff
@@ -1761,6 +1762,108 @@ def updater(_n: int) -> None:
...
+                elif op == 1:
+                    d[key] = i
+                elif op == 2:
+                    d.setdefault(f"sd{n}-{i}", i)
+                elif op == 3:
+                    with contextlib.suppress(KeyError):
+                        del d[key]
+                elif op == 4:
+                    d.pop(key, None)
+                elif len(d):
```
asvetlov:
```suggestion
                else:
```

PR #1442, opened 2026-09-11
Track md_finder_t's visited entries out of band instead of marking hashes

asvetlov:
## What do these changes do?

`md_finder_t` (the getall()/popall()/key-replacement lookup helper used
when a key has duplicate entries) used to deduplicate a repeated slot in
the hash probe sequence by temporarily ORing a sentinel bit into the
entry's hash, then restoring it once the lookup finished. It now keeps
its own small buffer of entry indices it has already returned instead:
an 8-element array embedded in the finder, falling back to a
heap-allocated, growable one for keys with more duplicates than that.
The finder also now holds its own reference to the key identity instead
of borrowing the caller's.

No.

No, it is an internal-only rework of one C helper with no new public
surface.

## Related issue number

N/A

## Checklist

- [ ] If you provide code modification, please add yourself to `CONTRIBUTORS.txt` (N/A: no `CONTRIBUTORS.txt` in this repo)

<details>

Tests: `pytest -q` (1839 passed) against the default C-extension build, a
debug build (`MULTIDICT_DEBUG_BUILD=1`, asserts enabled), and the
pure-Python leg (`MULTIDICT_NO_EXTENSIONS=1`); two tests added exercising
more than 8 duplicate keys for `getall()` and for `d[key] = value`
replacement (the C extension's fixed inline buffer overflows past 8, so
both need coverage). The pure-Python implementation has no equivalent
probe-sequence construct, so it needed no changes.

Lint: `pre-commit run` (ruff, clang-format, mypy) passes.

Docs: `make doc-spelling` passes.

</details>

### DIFF
--- a/multidict/_multilib/hashtable.h
+++ b/multidict/_multilib/hashtable.h
@@ -23,12 +23,22 @@ typedef struct _md_pos {
     uint64_t version;
 } md_pos_t;
 
+/* Number of already-found entry indices md_finder_t can track without
+   falling back to a heap allocation; see the comment above md_finder_t
+   for why the tracking is needed at all. */
+#define MD_FINDER_EMBEDDED_VISITED 8
+
 typedef struct _md_finder {
     MultiDictObject* md;
     htkeysiter_t iter;
     uint64_t version;
     Py_hash_t hash;
-    PyObject* identity;  // borrowed ref
+    PyObject* identity;  // owned ref
+    Py_ssize_t visited_embedded[MD_FINDER_EMBEDDED_VISITED];
+    Py_ssize_t* visited;       // == visited_embedded until it overflows
+    Py_ssize_t visited_count;  // number of valid entries in `visited`
+    Py_ssize_t
+        visited_capacity;  // capacity of the storage `visited` points to
 } md_finder_t;
 
 typedef enum _UpdateOp {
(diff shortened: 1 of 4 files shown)


multidict/_multilib/hashtable.h
```diff
@@ -59,14 +69,24 @@ changeing and if the number of DKIX_DUMMY slots grows to 1/4 of the total
...
+instance, not a mutation of the entry itself, so a single lookup's bookkeeping
+cannot be observed by anything looking at the entry through another path.
+
+`.update()` / `.extend()` / `.merge()` solve a related but different problem
+(recognizing, across a whole pass over another mapping's entries, an entry
+already touched earlier in the *same* call) with a separate mechanism: they
+still mark the entry's hash in place with MD_HASH_MARK (the hash range's high
+bit; entry hashes are folded non-negative by _unicode_hash(), so a real
+folded hash never has that bit set) and unmark every entry in one pass at the
+end via md_post_update(). That scheme is unrelated to md_finder_t.
```
```suggestion
```

multidict/_multilib/hashtable.h
```diff
@@ -756,17 +776,73 @@ md_prev(MultiDictObject* md, md_pos_t* pos, PyObject** pidentity,
     return ret;
 }
 
+/* Returns true if `index` is already among the entries this finder has
+   returned. Membership is checked with a linear scan: the embedded/overflow
+   array is not sorted, and for the expected case of very few duplicate
+   keys a scan of at most a handful of Py_ssize_t is cheaper than keeping
+   it sorted or hashed. */
```
```suggestion
```

multidict/_multilib/hashtable.h
```diff
@@ -756,17 +776,73 @@ md_prev(MultiDictObject* md, md_pos_t* pos, PyObject** pidentity,
...
+    for (Py_ssize_t i = 0; i < finder->visited_count; i++) {
+        if (finder->visited[i] == index) {
+            return true;
+        }
+    }
+    return false;
+}
+
+/* Records `index` as visited, growing from the embedded array to a
+   heap-allocated one (and doubling that one) as needed. */
```
```suggestion
```

multidict/_multilib/hashtable.h
```diff
@@ -845,17 +921,10 @@ md_finder_cleanup(md_finder_t* finder)
...
-    for (; finder->iter.index != DKIX_EMPTY; htkeysiter_next(&finder->iter)) {
-        if (finder->iter.index < 0) {
-            continue;
-        }
-        entry_t* entry = entries + finder->iter.index;
-        if (entry->hash < 0) {
-            entry->hash = finder->hash;
-        }
+    if (finder->visited != finder->visited_embedded) {
+        PyMem_Free(finder->visited);
```
Vizonex, line 925: Honestly believe a smarter mechanism should be used unless I'm missing something here. Memory Leaks were common a problem last time (for reference the fixes that I and other contributors made in 6.6.4). A better system for taking care of freeing memory should be necessary. Like a Boolean flag saying that memory was heap allocated.

Vizonex (commented):
I just have one little concern about this but other than that I support this Pull Request.

asvetlov:
Withdrawing this for now: the visited-entry tracking that replaces the in-place hash marking has a real performance regression I haven't resolved (O(N^2) blowup on lookups with many duplicate keys for the same key, up to ~18x slower at 500 duplicates, and it already regresses the existing 10-duplicate getall benchmark). Will revisit with a proper fix later rather than land this as-is.

PR #1445, opened 2026-09-12
Drain md->retired from reader exit, not just resize/clear

### LINKED ISSUE #1443: Free-threaded build: retired hash tables (md->retired) can accumulate unboundedly under sustained concurrent reads
On the `threadsafe-lookup` branch (#1441), `MultiDictObject` uses a reader-counted retirement scheme for lock-free `get()`/`getone()`/`__contains__`/`__getitem__`: a resize or `clear()` retires the old `htkeys_t` table onto `md->retired` instead of freeing it immediately if `num_active_readers != 0`, and only frees retired tables once `num_active_readers` reads 0.

That drain only happens inside `_md_retire()`, which runs when a new resize/clear retires another table (`_md_drain_retired()` is called opportunistically as a side effect). `_md_reader_exit()` itself never triggers a drain. So if read traffic is frequent enough that `num_active_readers` rarely (or never) returns to exactly 0 at the moment some other resize/clear happens to check it, tables already on `md->retired` are never freed, and the list grows without bound for the life of the object. This is a memory leak under sustained concurrent read+write load, not a crash.

Originally flagged by Greptile's automated review on #1441 as a P2/non-blocking finding and deferred.

Proposed direction: have `_md_reader_exit()` opportunistically drain `md->retired` when it is the reader that brings `num_active_readers` to 0 (it already gets the pre-decrement value for free from `atomic_fetch_add_ssize`, so detecting "I was last" costs nothing extra). This needs care: `_md_reader_exit()` currently runs with no critical section at all, and draining means decref'ing entries (which can run arbitrary `__del__` / trigger GC) and freeing memory from that lock-free path, a real change in what that path is allowed to do, not a one-line patch. It should get its own PR and its own free-threaded stress tests rather than folding into #1441.

<details>

Found during a manual review pass of `multidict/_multilib/hashtable.h`'s lock-free read path on the `threadsafe-lookup` branch.
</details>

asvetlov:
## What do these changes do?

On the free-threaded build, a resize/`clear()` retires its old hash table onto `md->retired` instead of freeing it immediately when a lock-free reader might still be using it, and only `_md_retire()` ever drained that list, as a side effect of a later resize/`clear()` checking whether the active-readers gate had reached 0. `_md_reader_exit()` never triggered a drain itself, so under read traffic frequent enough that the gate rarely lands on exactly 0 at the moment some other resize/`clear()` happens to check it, a retired table could sit on `md->retired` for the rest of the object's life: an unbounded memory leak under sustained concurrent read and write load.

`_md_reader_exit()` now drains `md->retired` itself when its own decrement brings the gate to 0. That makes `md->retired` a genuine lock-free stack (pushes from `_md_retire()` and pop-alls from `_md_drain_retired()` now run concurrently with each other, not just with readers walking a table), so pushing needs a CAS loop instead of a plain link, and `_md_retire()` has to push before checking whether the gate is already at 0 rather than after; checking first left a window where a concurrent reader's own check could run in between and find nothing yet to reclaim.

This is a C-extension-only, free-threaded-build-only concern; the pure-Python implementation has no retirement scheme to regress.

No observable behavior change; this only affects internal memory reclamation timing on the free-threaded build.

No. The change is localized to `_md_reader_exit()`/`_md_retire()`/`_md_drain_retired()` in `hashtable.h` plus two small pointer-atomic helpers in `atomic_helpers.h` (GCC/Clang, C11, and MSVC backends), with a dedicated regression test.

## Related issue number

Fixes #1443

## Checklist

- N/A If you provide code modification, please add yourself to `CONTRIBUTORS.txt`

<details>

Tests: full suite passed on the free-threaded debug build (`MULTIDICT_DEBUG_BUILD=1`, Python 3.14.7 free-threaded, `python -X gil=0 -m pytest -q`), the free-threaded release build, the default GIL-enabled C extension build, and the pure-Python build (`MULTIDICT_NO_EXTENSIONS=1`).

New test (`test_reader_exit_drains_retired_thread_safety`) was verified to fail reliably (10/10 runs) against the pre-fix code and pass reliably (20/20 runs) against the fix.

Lint: `pre-commit run` (ruff, clang-format, mypy) passed on all changed files. `make doc-spelling` passed.
</details>

### DIFF
--- a/CHANGES/1443.bugfix.rst
+++ b/CHANGES/1443.bugfix.rst
@@ -1,0 +1,10 @@
+Fixed an unbounded memory leak on the free-threaded build where a retired
+hash table could sit on ``md->retired`` for the rest of the object's
+lifetime under sustained concurrent read traffic: only a later
+resize/``clear()`` ever drained it, and if the active-readers gate never
+happened to land on exactly 0 at the moment one of those ran, nothing
+else would ever free it. Reader exit now opportunistically drains
+``md->retired`` itself, and retiring a table pushes it onto that list
+before checking whether it can be freed immediately, closing a window
+where a concurrent reader's own check could otherwise run between the
+two and observe nothing to reclaim -- by :user:`asvetlov`.
--- a/CHANGES/1443.bugfix.rst
+++ b/CHANGES/1443.bugfix.rst
@@ -1,10 +1,4 @@
 Fixed an unbounded memory leak on the free-threaded build where a retired
 hash table could sit on ``md->retired`` for the rest of the object's
-lifetime under sustained concurrent read traffic: only a later
-resize/``clear()`` ever drained it, and if the active-readers gate never
-happened to land on exactly 0 at the moment one of those ran, nothing
-else would ever free it. Reader exit now opportunistically drains
-``md->retired`` itself, and retiring a table pushes it onto that list
-before checking whether it can be freed immediately, closing a window
-where a concurrent reader's own check could otherwise run between the
-two and observe nothing to reclaim -- by :user:`asvetlov`.
+lifetime under sustained concurrent read traffic instead of being freed
+-- by :user:`asvetlov`.
--- a/multidict/_multilib/atomic_helpers.h
+++ b/multidict/_multilib/atomic_helpers.h
@@ -63,6 +63,28 @@ atomic_load_ptr(void* const* obj)
     return __atomic_load_n(obj, __ATOMIC_SEQ_CST);
 }
 
+static inline void*
+atomic_exchange_ptr(void** obj, void* value)
+{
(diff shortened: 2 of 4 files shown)


asvetlov:
Addressed the reader-synchronization finding: the test now has each reader signal a per-thread `threading.Event` after a warm-up batch of reads and then keep running, unblocked, straight into its main loop. `clear()` only proceeds once every reader has signaled, so it can no longer land before any reader has actually started. Also raised the reader count from 8 to 16, which empirically brings the fraction of runs that land `clear()` while a reader is genuinely inside the lock-free section to consistently over 90% against the pre-fix code (40/40 clean runs against the fix, 0 false failures).

PR #1446, opened 2026-09-12
Use atomic ops for version incrementation in free-threading mode

### LINKED ISSUE #1444: Use atomic ops for version incrementation in free-threading mode

asvetlov:
## What do these changes do?

`NEXT_VERSION()` bumps `state->global_version`, a counter shared by
every `MultiDict`/`CIMultiDict` instance in the process, since it
lives on the module state rather than the object. Each object's own
lock only serializes access to that one object, so under a
free-threaded build two threads mutating two *different* instances
could bump the shared counter at the same time with a plain `++`,
losing an update and handing out a duplicate (or non-monotonic)
version number from `getversion()`. The fix routes the increment
through an atomic fetch-add under `Py_GIL_DISABLED`
(`atomic_fetch_add_uint64_relaxed()`, added to `atomic_helpers.h`
alongside the existing `Py_ssize_t`/pointer helpers), leaving the
GIL-enabled build's plain increment untouched.

The pure-Python implementation has the identical shape of bug (a
module-level `_version` array shared by every instance, bumped with
an unsynchronized `v[0] += 1`), but it has no other thread-safety
machinery of its own, so fixing just this one counter there is out
of scope for this PR; left as-is.

Verified the race and the fix empirically (not just via the test
suite) with a throwaway stress script under a free-threaded build:
16 threads each mutating their own multidict and recording every
`getversion()` reading. Before the fix, roughly 5-30% of readings
collided across runs; after the fix, all readings were unique across
several runs.

No visible behavior change; `getversion()` values remain an
opaque, per-mutation token. This only fixes a free-threaded-build
race where two unrelated multidicts could momentarily be assigned
the same version number.

No. It's a small, self-contained fix mirroring the existing atomic-op
pattern already used elsewhere in `atomic_helpers.h`.

## Related issue number

Fixes #1444

## Checklist

<details>

Tests: `pytest -q`, all green on the C extension under a normal build, a free-threaded 3.14t build, and a free-threaded debug build (`MULTIDICT_DEBUG_BUILD=1`); also confirmed the pure-Python leg still passes (`--no-c-extensions`) and correctly deselects the new test.
New test: `tests/test_multidict.py::test_version_thread_safety`, marked `c_extension`.
Lint: `pre-commit run` (ruff check/format, clang-format, mypy) passed on the changed files.
Docs: `make doc-spelling` passed with no wordlist additions needed.

Pure Python left unfixed on purpose (see "What do these changes do?" above); happy to file a follow-up issue for it if wanted.

</details>

### DIFF
--- a/multidict/_multilib/state.h
+++ b/multidict/_multilib/state.h
@@ -1,6 +1,8 @@
 #ifndef _MULTIDICT_STATE_H
 #define _MULTIDICT_STATE_H
 
+#include "atomic_helpers.h"
+
 #ifdef __cplusplus
 extern "C" {
 #endif
@@ -128,7 +130,17 @@ get_mod_state_by_def(PyObject* self)
 static inline uint64_t
 NEXT_VERSION(mod_state* state)
 {
+    /* state->global_version is shared by every MultiDict/CIMultiDict
+       instance in the process (it lives on the module state, not the
+       object), while each object's critical section only serializes
+       access to that one object. Under Py_GIL_DISABLED, two threads
+       mutating two different multidict instances can bump this counter
+       at the same time, so a plain increment can lose an update. */
+#ifdef Py_GIL_DISABLED
+    return atomic_fetch_add_uint64_relaxed(&state->global_version, 1) + 1;
+#else
     return ++state->global_version;
+#endif
 }
 
 #ifdef __cplusplus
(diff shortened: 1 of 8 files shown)


multidict/_multilib/state.h
```diff
@@ -128,7 +130,17 @@ get_mod_state_by_def(PyObject* self)
 static inline uint64_t
 NEXT_VERSION(mod_state* state)
 {
+    /* state->global_version is shared by every MultiDict/CIMultiDict
+       instance in the process (it lives on the module state, not the
+       object), while each object's critical section only serializes
+       access to that one object. Under Py_GIL_DISABLED, two threads
+       mutating two different multidict instances can bump this counter
+       at the same time, so a plain increment can lose an update. */
```
asvetlov:
```suggestion
```

PR #1441, opened 2026-09-11
Make __contains__ lock-free with reader-counted table retirement

asvetlov:
## What do these changes do?

First slice of the lock-free read path deferred out of #1438: instead
of a critical section, `__contains__` now protects itself with a
reader-counted table retirement scheme, and the hash table a lock-free
reader might still be walking is kept alive until it's safe to free.

Two counters, gated entirely behind `Py_GIL_DISABLED`:

- `MultiDictObject.active_readers`: a coarse "some lock-free reader is
  in flight on this object" gate, bumped before a reader ever
  dereferences a keys table and dropped once it's done
  (`_md_reader_enter()`/`_md_reader_exit()`).
- `htkeys_t.readers`: a per-table count, advisory only, used purely as
  a defensive assertion once the coarse gate is already known closed.
  It is not itself load-bearing for safety: an embedded refcount can't
  safely announce "about to use this" without a bootstrap hazard,
  since the increment itself would dereference memory that might
  already be freed by the time it runs.

`_md_retire()`/`_md_drain_retired()`: a resize/shrink/clear no longer
frees the outgoing table immediately, only once `active_readers` reads
0 at a point synchronized (`seq_cst` on both the `md->keys` swap and
the `active_readers` ops -- anything weaker leaves a real,
architecture-dependent race, the same shape as Dekker's algorithm). A
table that can't be freed yet moves to a per-object retired list and
is freed opportunistically the next time the object resizes.

`_md_shrink()` gets a second implementation under `Py_GIL_DISABLED`:
the existing one compacts the currently-published table in place,
which races a lock-free reader walking the very memory being
rewritten. The free-threaded version reuses `_md_resize()` at the
current size instead (build a new table, swap, retire the old one),
which already has the right shape; the GIL build keeps the original
in-place version unchanged.

`md_clone_from_ht()` gets a matching fix: it `memcpy`s another live
table's header wholesale, which was copying that table's in-flight
reader count and retirement link along with everything else.

`md_contains()` with `pret == NULL` (`__contains__`'s actual call
shape) is the first lock-free consumer: `entry->identity`/`hash` are
write-once-at-insertion and never mutated again outside `getall()`'s
already-critical-section-only `MD_HASH_MARK` toggling, so this walk
needs nothing beyond the table's own lifetime protection -- no atomic
entry-field stores or `TryIncRef` machinery required, unlike `get()`'s
value read (out of scope here; deferred as its own follow-up). The
critical sections around `multidict_sq_contains` and
[...]

PR #1447, opened 2026-09-13
Make pure-Python MultiDict safe under free-threading

asvetlov:
## What do these changes do?

Adds per-instance locking to the pure-Python `MultiDict`/`CIMultiDict`
fallback (`multidict/_multidict_py.py`) so it gives the same
crash-free, corruption-free guarantee as the C extension under a
free-threaded (no-GIL) build. Every method that touches an instance's
hash table takes that instance's own `RLock` (created once in
`__new__`, reused across a later re-init); two-object operations
(`update()`, `extend()`, `merge()`, `__eq__()`, copy construction) lock
both objects together in a fixed, deadlock-safe order; view iterators
take the lock per step rather than for the whole iteration, so a
concurrent mutation between steps surfaces as the same "changed during
iteration" `RuntimeError` a single-threaded caller already gets today
instead of a deadlock; and the version counter shared by every
instance gets its own lock. On a regular GIL-enabled interpreter the
locked methods are the exact same function objects as before this
change -- no wrapper, no lock, no measurable overhead.

Yes, but only on a free-threaded build: concurrent use of a
pure-Python `MultiDict`/`CIMultiDict` from multiple threads is now
safe against crashes and data corruption, matching the C extension.
On a regular (GIL-enabled) build there is no behavior or performance
change.

No. The change is self-contained to `_multidict_py.py`, gated entirely
behind a runtime free-threading check computed once at import, and
mirrors the locking approach already reviewed and merged for the C
extension (#1438, #1441, #1445, #1446). The C extension itself is
unaffected; this only touches the pure-Python fallback.

## Related issue number

N/A -- no linked issue. Continues the free-threading hardening already
done for the C extension.

## Checklist

- [ ] If you provide code modification, please add yourself to `CONTRIBUTORS.txt` (N/A, this repo has no `CONTRIBUTORS.txt`)

<details>

Tests: `pytest -q` passes on both backends; the pure-Python leg was
additionally run under a free-threaded interpreter (CPython 3.13.7t,
`PYTHON_GIL=0`) multiple times with coverage active, plus a standalone
stress script driving 16 threads through thousands of concurrent
mutations, with no crashes or corruption.

Lint: `ruff check`, `ruff format --check`, and `mypy` (3.11 and 3.13,
via `pre-commit`) all pass on the changed files.

Docs: `make doc-spelling` passes.

Note: while calibrating test load, a pre-existing, unrelated race in
`update()`/`extend()`/`merge()`/`clear()` was found -- reachable even
on a plain GIL-enabled interpreter under heavy concurrent use (pure
Python bytecode isn't atomic even under the GIL). Confirmed present on
unmodified `master` too, so it predates and is out of scope for this
PR; a follow-up is planned separately. Two of the new tests that would
otherwise be flaky against it are skipped on non-free-threaded builds,
with the reason stated inline.

</details>

Drafted with Claude Code; reviewed by asvetlov.


asvetlov:

asvetlov:
Addressed the reciprocal-view deadlock: `_locked_md` only held the calling multidict's own lock, but a view's set-algebra methods can iterate an `other` argument that is itself a view over a *different* multidict, and each iteration step takes that multidict's lock too. `a.items() & b.items()` racing `b.items() & a.items()` could hold one lock while blocked on the other. Fixed in 56b8a94..da6c743 by extending `_other_lock()` to resolve a view's backing multidict and pairing both locks up front, in the same fixed order `_PairLock` already uses for `update()`/`extend()`/`merge()`/`__eq__()`/copy construction (this also covers the same exposure there, e.g. `d1.update(d2.items())`). Reproduced the deadlock against the prior code, confirmed the fix resolves it, and added `test_pure_python_reciprocal_view_ops_no_deadlock` as a regression test.

The `_md_restore_all_hashes()` finding is accurate but out of scope for this PR: it's a C-extension-only race in `to_dict()` (added by #1288, merged into this branch from master, not introduced here), the same class of bug #1448 already fixed for the neighboring `md_find_next()`/`md_post_update()` paths. The pure-Python `to_dict()` has no equivalent hash-marking scheme and isn't affected. Filed as a separate follow-up to keep this PR scoped to the pure-Python implementation.

asvetlov:
Good catch, fixed in da6c743..45cbb9a: \`_other_lock()\` recognized the view classes but not the \`_Iter\` that \`iter(view)\` returns, so \`a.items() & iter(b.items())\` racing \`b.items() & iter(a.items())\` (and the equivalent for \`update()\`/\`extend()\`/\`merge()\` given a bare iterator argument) could still deadlock the same way. \`_Iter\` already carries the RLock its view resolved at construction time, so recognizing it there was a one-line fix. Reproduced the deadlock against the prior code, confirmed the fix resolves it, and added a regression test.

PR #1449, opened 2026-09-13
Lock pure-Python MultiDict mutators against concurrent corruption

asvetlov:
## What do these changes do?

Every method in the pure-Python `MultiDict`/`CIMultiDict` fallback
(`multidict/_multidict_py.py`) that mutates `self._keys`/`self._used`
(`add()`, `__setitem__`/`__delitem__`, `setdefault()`,
`pop()`/`popone()`/`popall()`, `popitem()`, `update()`, `extend()`,
`merge()`, `clear()`, and re-`__init__`) ran without any
synchronization on a GIL-enabled build. Pure-Python bytecode is not
atomic even under the GIL, so two threads calling these concurrently
on the same instance (or reading one as the argument to another
thread's `update()`/`extend()`/`merge()`) can interleave mid
hash-table insert or deletion and corrupt the shared index table
(hanging in an infinite probe loop or raising `AttributeError`), or
silently drop a completed write (e.g. a concurrent `add()` writing
into a table an in-progress `extend()`'s resize is about to replace).
This is reproducible on a regular, GIL-enabled interpreter given
enough contention; the coverage tracing this test suite already runs
under is enough to hit it reliably, no free-threaded build required.

Every `MultiDict`/`CIMultiDict` instance now gets its own `RLock`,
held for the duration of these operations. Two-object operations
(`__init__`/`update()`/`extend()`/`merge()` reading another
`MultiDict` as their argument) lock both instances in a fixed order to
avoid deadlock. `_add_with_hash()`/`_add_with_hash_for_upd()` and
`popitem()` also had their write ordering fixed so an index is only
published once its entry is reachable, and an entry's index is
cleared before the entry is dropped from the entries list.

Read-only methods (`getall()`, `getone()`, `__contains__()`,
`__eq__()`, `__repr__()`, `to_dict()`, iteration, view operations) are
unaffected and stay locked only on a free-threaded build, matching the
scope of #1447: this PR is about lost or corrupted writes between
mutators, not about a reader observing a torn read.

This builds directly on #1447's locking infrastructure (`RLock`,
`_PairLock`, `_other_lock`) rather than adding a separate scheme:
`_locked_always`/`_locked_pair_always` are unconditional counterparts
to `_locked`/`_locked_pair` that skip the free-threading gate, applied
to every mutating method. `__new__` now always constructs the lock
(not just under free-threading), since these methods need it on every
build.

The C extension already serializes all of these operations with a
critical section (see #1438, #1441) and is unaffected by this change.

No observable API change. Concurrent use of a pure-Python `MultiDict`/
`CIMultiDict` from multiple threads no longer risks corruption, a
hang, or a silently lost write; single-threaded use is unaffected
apart from a small added lock overhead (see benchmark note below).

No. The change is self-contained to `_multidict_py.py`'s mutating
methods and mirrors the locking approach already used by the C
extension's critical sections, and reuses #1447's infrastructure
rather than adding a parallel one.

## Related issue number

N/A -- no linked issue. This race predates and is unrelated to the
free-threading work in #1447; that PR's own testing flagged this exact
race as reachable on a plain GIL build and out of scope for it (see
the `_gil_build_race_skip` marker it added to `tests/test_multidict.py`,
with a comment noting a follow-up was planned separately). This is
that follow-up.

## Checklist

- [ ] If you provide code modification, please add yourself to `CONTRIBUTORS.txt` (N/A, this repo has no `CONTRIBUTORS.txt`)

<details>

Repro: confirmed the race on unmodified master with a repro script
running `update()`/`merge()`/`extend()` from 8 threads; it hung with
multiple threads stuck in `iter_hash()`'s probe loop (`faulthandler`
stack dump captured during the session). The same script passes
cleanly after this fix, run 10+ times, plus an extended variant that
also exercises `popitem()`/`clear()` and asserts internal consistency
(every live entry reachable via its own hash bucket) after the run,
run 8+ times.

This branch was rebased onto master after #1447 merged (it touched
the same file, adding the free-threading-only locking this PR now
also reuses); an automated review (Greptile) on the pre-rebase commit
correctly flagged two real gaps, both addressed in the second commit:
a reentrancy deadlock (moot after picking up #1447's `RLock`, now
covered by a regression test) and `add()`/`__setitem__()`/etc. still
running unlocked, which this PR's second commit closes by locking
every mutator, not just the original five methods.

Tests: `pytest -q` (1908 passed) against the default C-extension
build, `MULTIDICT_NO_EXTENSIONS=1` + `pytest -q --no-c-extensions`
(963 passed, 6 skipped) against the pure-Python GIL build, and the
same pure-Python command under a free-threaded interpreter (CPython
3.13.7t, 969 passed), all in isolated venvs built from this worktree.
Coverage stayed at 99% on `_multidict_py.py` on every leg, with the
free-threading-only branches covered by the free-threaded leg and
vice versa (the same split #1447's own decorators already have).

Performance: locking every mutator (not just the original five) adds
roughly 150-200ns of uncontended lock overhead per call on the
pure-Python backend. For O(n) bulk operations (`update()`/`extend()`/
`merge()`) this is a few percent at a realistic size (~20 items, e.g.
typical HTTP headers) and within measurement noise; for O(1)
single-item operations (`add()`/`popone()`/etc.) it's a fixed cost
that doesn't shrink with size, roughly 15-20% by `timeit` best-of-7
samples. The C extension is unaffected either way. Flagging per the
trade-off note in `AGENTS.md` rather than trying to eliminate it,
since correctness under a GIL build (data loss, not just a crash) was
the point; the `@contextmanager`-based approach tried initially added
roughly 5x more overhead than this and was dropped in favor of
`with self._lock:` directly and, for the two-object case, plain
`acquire()`/`release()` calls.

Lint: `pre-commit run` (ruff, mypy for 3.11 and 3.13) passed on the
changed files.

Docs: `make doc-spelling` passed after adding `bytecode` to
`docs/spelling_wordlist.txt`.

Drafted with Claude Sonnet 5; reviewed by asvetlov.

</details>


Vizonex (approved):
The Use of a wrapper seems to be the correct approach here. I Approve this pull request since it will help strengthen against concurrent edits in free-threaded mode.

asvetlov:
Locks are slow, but without locks the code is literally not thread-safe.
I expect that people usually install C Extension anyway; which in GIL mode provides multithreading capability without extra overhead.

PR #1453, opened 2026-09-14
Fix items() set algebra corrupting state on reentrant equality

asvetlov:
## What do these changes do?

`items()` set algebra (`&`, `|`, `-`, `in`, `isdisjoint()`) walks a key's
hash chain while temporarily marking visited entries, then compares stored
values against a caller-supplied one before unmarking. That comparison can
run arbitrary user code through a custom `__eq__`, and code that calls back
into the same `MultiDict` from there (`getall()`, for example) while the
chain is still marked observes some matching entries as missing. Depending
on the operation this returns wrong data, raises a bogus `KeyError`, or, in
the C extension, hangs the process outright (see
[#1317 (review)](https://github.com/aio-libs/multidict/pull/1317#pullrequestreview-5187377912)
for the report and a reproduction).

The fix materializes the matching `(key, value)` pairs into a plain list
after the hash chain has been fully walked and restored, and only then runs
the value comparisons against that snapshot. This covers the C extension's
`and1`/`and2`/`or1`/`sub2`/`contains`/`isdisjoint` paths via a new
`md_finder_collect()` helper, and the pure-Python `__and__()`, the only
pure-Python method using the same mark-then-compare pattern (the other
pure-Python methods don't mark and were never affected).

No, other than the crash/corruption going away. Ordinary comparisons (an
`__eq__` that doesn't touch the multidict) behave exactly as before.

No. The fix is a mechanical materialize-then-compare change with no new
public API, plus regression tests.

## Related issue number

Follows up on the review comment at
https://github.com/aio-libs/multidict/pull/1317#pullrequestreview-5187377912.
No separate issue is filed.

## Checklist

- [x] Documentation reflects the changes -- N/A

<details>

Tests: `pytest -q` passed (1927 passed, 12 skipped) under both the C
extension and `MULTIDICT_NO_EXTENSIONS=1` pure-Python builds, plus a debug
build (`MULTIDICT_DEBUG_BUILD=1`, asserts on, `-O0`).
Lint: `pre-commit run` (ruff, ruff-format, clang-format, mypy for 3.11 and
3.13) passed on all changed files.
Docs: `make doc-spelling` passed with no wordlist additions needed.
Confirmed pre-fix: the new tests hang (C extension) or fail with wrong
results/`KeyError` (pure Python) against the unmodified code.

</details>


asvetlov:
Well, slight performance degradation is an acceptable price for correctness.
The set algebra is not the hot path anyway; it exists for the sake of ABC commitment.

PR #1457, opened 2026-09-14
Stop case-folding istr keys in case-sensitive MultiDict

youdie006:
## What do these changes do?

`_key_to_identity` in `multidict/_multilib/hashtable.h` returned the cached lower-cased `canonical` for an `istr` key, so a case-sensitive `MultiDict` matched that key case-insensitively. The pure Python `_CSMixin._identity` (`multidict/_multidict_py.py:596`) has no `istr` branch and returns the key unchanged, so the two backends disagreed.

The C container also contradicted itself:

```pycon
>>> d = MultiDict([(istr("A"), 1)])
>>> list(d.keys())
['A']
>>> "A" in d
False
>>> "a" in d
True
```

Dropping the branch lets `istr` fall through to the existing `PyUnicode_Check` path at `hashtable.h:141`, which preserves case exactly as the other two `str` branches in that same function already do. The case-insensitive sibling `_ci_key_to_identity` is untouched.

Yes. A case-sensitive `MultiDict` no longer folds case for `istr` keys. That restores the documented contract in `docs/multidict.rst`, added by #1403 when #1397 was closed: "case-insensitive matching is handled by :class:`CIMultiDict`, not by ``istr`` equality". `CIMultiDict` is unaffected.

No; it removes three lines and adds two tests to the existing parametrized suite.

## Related issue number

Related to #1397 and #1403, which state the contract this restores. No issue to close.

## Checklist

- [x] Documentation reflects the changes (`docs/multidict.rst` already states the restored contract; no change needed)
- [ ] If you provide code modification, please add yourself to `CONTRIBUTORS.txt`. N/A, no such file in this repo.
- [ ] `make doc-spelling` passes and any new technical words are added to `docs/spelling_wordlist.txt`. This could not be run here: `sphinxcontrib.spelling` needs the `enchant` C library, which is not installed and cannot be without root. It fails identically with and without the new fragment, so the failure is the environment rather than the fragment. As a substitute, every word the checker would see after roles and literals are stripped is ordinary English, so no addition to `docs/spelling_wordlist.txt` is needed. Worth a CI run to confirm.

Drafted with Claude (Anthropic); not yet reviewed by a human, so it stays in draft until :user:`youdie006` has read it.

<details>

Base `dd74db1`.

Differential, case-sensitive `MultiDict`, C versus pure Python:

| expression | C | pure Python |
|---|---|---|
| `istr('A') in MultiDict([('a',1)])` | True | False |
| `MultiDict([('a',1)]).get(istr('A'))` | 1 | None |
| `'A' in MultiDict([(istr('A'),1)])` | False | True |
| `'a' in MultiDict([(istr('A'),1)])` | True | False |

Rows that must agree, and do, so the sweep is not vacuous: `istr('A') in CIMultiDict([('a',1)])` True/True; `'A' in CIMultiDict([('a',1)])` True/True; `'A' in MultiDict([('a',1)])` False/False; `'a' in MultiDict([('a',1)])` True/True. Both backends already agree on plain `str` case sensitivity and on `CIMultiDict` folding; only `istr` in a case-sensitive `MultiDict` diverges.

Tests, with the loaded `.so` md5 printed on every row:

| row | `hashtable.h` md5 | scope | result |
|---|---|---|---|
| base, new tests | `78be1260` | focused | 2 failed `[c]`, 2 passed `[py]` |
| patched, new tests | `b8eaf175` | focused | 4 passed |
| base, original tests | `78be1260` | full | 1904 passed, 12 skipped |
| patched, original tests | `b8eaf175` | full | 1904 passed, 12 skipped |
| patched, new tests | `b8eaf175` | full | 1908 passed, 12 skipped |

The pure Python backend passes the new tests before the fix, so they encode the existing reference behaviour rather than a new opinion. Base and patched both give 1904 on the original suite, so nothing regresses and no existing test reached this divergence.

Removing the `istr` branch from the case-insensitive sibling as well survives the suite. That is an equivalent mutant, not a test gap: `istr.h:59` builds `canonical` by calling `str.lower()`, so on the folding path the branch is only a cache. Checked by execution over 900 key and query combinations including `ß`, `İ`, `ı`, `Σ`, `ς`, `ẞ`, with zero differences against the unmutated build. The same fact is what makes the case-sensitive path wrong, since there it applies a real `lower()`.

Lint: `clang-format --dry-run -Werror` on `hashtable.h` exit 0; `ruff check` and `ruff format --check` on `tests/test_multidict.py` clean; `mypy --python-version=3.13` reports no issues.

</details>


asvetlov (approved):
Nice catch!
The bug exists for more than 1 year!

PR #1460, opened 2026-09-14
Add Hypothesis property and stateful fuzz tests

asvetlov:
## What do these changes do?

Adds a Hypothesis-based property and stateful test suite covering
`MultiDict`/`CIMultiDict` core semantics (`add`, `setitem`, `delitem`,
`popone`, `popall`, `popitem`, `setdefault`, `update`, `extend`,
`merge`, `clear`), case-insensitive key folding, proxies, views
(`items`/`keys`/`values` set algebra), the iterator mutation guard,
and a handful of Hypothesis-fuzzed threaded stress scenarios. Split
across four new files (`test_hypothesis_multidict.py`,
`test_hypothesis_views.py`, `test_hypothesis_iters.py`,
`test_hypothesis_concurrency.py`) plus a shared `hypothesis_helpers.py`
with the strategies and a small independent reference model used by
the stateful test. Everything reuses the existing `any_multidict_class`
-style fixtures in `tests/conftest.py`, so each property already runs
against both the C-extension and pure-Python backends and both
case-sensitive and case-insensitive classes.

`update()`/`merge()`'s handling of duplicate keys within a single call
turned out to be more intricate than the docstrings suggest (existing
duplicates are matched positionally against incoming duplicates,
excess incoming ones are appended, excess existing ones are dropped);
the reference model's comments and this description spell out the
exact rule, verified empirically against `tests/test_update.py`'s
existing cases and a few thousand random operation sequences before
being encoded as the oracle.

While fuzzing the iteration guard I found two narrow, real dual-backend
divergences (a pure-Python case where mutating before the first
`next()` call raises `StopIteration` instead of `RuntimeError`, and a
C-extension case where `update()` overwriting an already-existing key
doesn't invalidate live iterators the way `__setitem__` does). Both are
filed as separate follow-ups rather than folded into this test-only
change; the new tests here are written to avoid asserting into either
gap.

No. Test-only change, plus a mypy pre-commit config tweak (see below).

No. It adds one new runtime test dependency (`hypothesis`, pinned in
`requirements/pytest.txt`) and the new tests run automatically as part
of the existing `pytest`/`make test` invocation; no new CI job is
needed.

## Related issue number

N/A

## Checklist

<details>

Tests: `pytest -q` on both the C-extension build (2042 passed, 17
skipped) and the pure-Python build (`MULTIDICT_NO_EXTENSIONS=1`) --
the four new files individually verified passing on both backends;
a full-suite pure-Python confirmation run was still in progress at
the time this PR was opened.

Lint: `pre-commit run` (ruff check, ruff format, mypy for 3.11 and
3.13) all passing. `hypothesis` was added to both mypy hooks'
`additional_dependencies`, and `.mypy.ini` gained a per-file
`disable_error_code = misc` for the four new test files, since
Hypothesis's `@given`/`@rule`/`@invariant` decorators are typed as
`Callable[..., X]`, which `disallow_any_decorated` otherwise flags on
every decorated test regardless of how precisely the test itself is
annotated.

Docs: `make doc-spelling` passing (`backends`, `stateful` added to
`docs/spelling_wordlist.txt`).

Drafted with Claude Code; reviewed by asvetlov.

</details>


asvetlov:
Pushed a follow-up commit addressing the automated review feedback:

- Fixed `case_variant()` to reject a case flip that would change a character's own `str.lower()` identity (Greek final sigma `ς` upper-cases to `Σ`, whose `.lower()` is plain `σ`, not the final form back) -- could otherwise generate a `CIMultiDict` key not equivalent to the one actually stored.
- Removed `from __future__ import annotations` from all five new files.
- Added deterministic coverage tests for `_run_op`, `_read_second`, `_reciprocal`, and both `_apply_mutation` helpers, so every branch runs regardless of what Hypothesis draws for the corresponding fuzz test.
- Dropped an unused `md` local and an unnecessary lambda wrapping `iter` (CodeQL).
- Renamed the CHANGES fragment to `1460.contrib.rst` to match this PR.

Also merged in master, which had picked up the two dual-backend bugs this PR's fuzzing surfaced (#1458, #1459) since this branch was opened -- full suite still green on both backends (2143 passed) after the merge, and `pre-commit run --all-files` passes.

Vizonex (approved):
Hopefully the fuzzing will allow us and other contributors capture any new bugs that may arise, especially given the circumstances of the C Extension. I have no problems with adding the hypothesis extension to the current library for running the tests.

PR #1466, opened 2026-09-15
Speed up adding many values for the same key

bdraco:
## What do these changes do?

Adding many values under one key got slower with every value, since each insert walked past all earlier values for that key on the same probe sequence; resizes rebuilt the table the same way. Once the perturb shift reaches 0 the probe sequence only depends on the slot, so large tables now remember the last slot used from that point and the next insert resumes there. Both the C extension and the pure Python implementation are changed.

The C lookup loop is unswitched by index size by hand so the extra check stays cheap.

Memory is only spent where the problem exists. The hint array is a separate allocation made the first time a table sees a long probe after the perturb shift reaches 0, which in practice means many values for one key. Multidicts without that pattern, including large ones with distinct keys, only pay for one pointer in the keys header, 8 bytes at any size; none of 120 random distinct key multidicts with 1000 to 50000 entries allocated it. Tables that do allocate use 2 bytes per slot below `2**16` slots and 4 bytes from there on, only for tables of 1024 slots or more. The raw allocator is used so allocating can't suspend the critical section on free threaded builds, and if it fails probing just continues without hints. `copy()` does not copy the hints. The pure Python implementation follows the same rules with a dict created on the first long probe.

No, only performance. The benchmarks added in #1465 cover it.

No, the change is contained in the slot lookup used for inserts and table rebuilds.

## Related issue number

N/A

## Checklist

- [ ] Documentation reflects the changes N/A
- [ ] If you provide code modification, please add yourself to `CONTRIBUTORS.txt` N/A

<details>
<summary>Benchmark and test details</summary>

Local `pytest --codspeed` wall time with the benchmarks from #1465, run time per iteration:

| benchmark | master | this PR |
|---|---|---|
| `test_create_multidict_with_items_same_key[cs-c]` | 598us | 11.9us |
| `test_create_multidict_with_items_same_key[ci-c]` | 620us | 40.1us |
| `test_create_multidict_with_items_same_key[cs-py]` | 20.6ms | 1.23ms |
| `test_multidict_add_same_key[cs-c]` | 987us | 109us |
| `test_multidict_add_same_key[ci-c]` | 1.01ms | 137us |
| `test_multidict_add_same_key[cs-py]` | 33.3ms | 2.59ms |
| `test_create_multidict_with_many_items[cs-c]` | 40.7us | 42.3us |
| `test_create_multidict_with_many_items[cs-py]` | 2.16ms | 2.17ms |

Callgrind instruction counts against master (gcc, arm64 Linux, `PYTHONHASHSEED=1`): 5000 distinct keys through the constructor +1.5%, 5000 `add()` calls +0.004%, `CIMultiDict` of 16 headers +0.04%; 20000 values for one key 2.2B instructions on master, 6.8M here.

`sys.getsizeof()`:

| entries | master | distinct keys | one repeated key |
|---|---|---|---|
| 60 | 2960 | 2968 | 2968 |
| 342 | 23984 | 23992 | 26040 |
| 1000 | 47888 | 47896 | 51992 |
| 5000 | 191248 | 191256 | 207640 |

Tests pass on 3.10 and 3.14 with the C extension (release and `MULTIDICT_DEBUG_BUILD=1`), on 3.14t with a debug build, and with `MULTIDICT_NO_EXTENSIONS=1 --no-c-extensions`. The new tests fail on both backends if the hint is deliberately corrupted.

</details>

Drafted with Claude Code; reviewed by @bdraco.


bdraco:
> 1. Tails add about 16-17% memory to every table with 1024+ slots, and the PR doesn't mention it
> [multidict/_multilib/htkeys.h:301-308](https://github.com/aio-libs/multidict/blob/ebd50afec466443c58af5838b5d5f8d9ddbc750b/multidict/_multilib/htkeys.h#L301-L308)

Most multidict have < 60 keys so has no effect on common case. Not worth the complexity. This PR is concerned about making the pathological case survivable

multidict/_multilib/htkeys.h
```diff
@@ -276,12 +294,32 @@ static const htkeys_t empty_htkeys = {
...
+   with perturb == 0. From there the probe sequence only depends on the
+   slot and slots never become empty again, so probing can resume.
+   Tables of 2**32 slots or more have none, slot + 1 must fit. */
+static inline size_t
+htkeys_tails_bytes(uint8_t log2_size)
+{
+    if (log2_size < HT_LOG_TAILS_MINSIZE || log2_size >= 32) {
+        return 0;
+    }
+    return sizeof(uint32_t) << log2_size;
```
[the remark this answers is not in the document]
bdraco: Most multidict have < 60 keys so has no effect on common case. Not worth the complexity. This PR is concerned about making the pathological case survivable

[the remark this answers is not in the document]
bdraco: Could still do a uint16_t for `log2_size < 16`, that should be rather cheap and avoid the lazy complexity

asvetlov (changes requested):
@bdraco I think we need to discuss this change.

The speedup of tests is impressive. 
But the price is high; every large multidict gets an additional tails array.
If it were for small dicts, I don't care too much; the memory usage is still relatively low.
Instead, the PR makes memory consumption for big multidicts worse.
The optimization works only for big multidicts with multiple values for the same key. If the multidict has no duplicates or few duplicates, we reserve extra memory and pay the price anyway.

I'm curious: is there any metric that confirms the fact that most huge multidicts have a massive amount of duplicates?

Anyway, the patch demonstrates a non-trivial approach that I never considered.

bdraco:
Please check aiohttp security queue

bdraco:
Switched to lazy allocation so the common case with unique keys only RAM cost is for the null pointer

bdraco:
> * `multidict_sizeof` reads `self->keys` several times without a critical section. The tails check adds two more reads to an existing free threaded race.

Its pre-existing
`if (self->keys != &empty_htkeys) size += htkeys_sizeof(self->keys);` was original code.

But simple enough to wrap it. Makes me wish we had RAII

bdraco:
nearly 0230 here so have to wait till tomorrow to address anything else.

bdraco:
probably should rename it now

asvetlov:
@bdraco are you refering to ~~https://github.com/aio-libs/aiohttp/security/advisories/GHSA-r7v6-hxr5-4cpr~~?
No, perhaps https://github.com/aio-libs/aiohttp/security/advisories/GHSA-xgw8-8ccx-rmqh

multidict/_multidict.c
```diff
@@ -1264,7 +1264,15 @@ static PyObject*
 multidict_sizeof(MultiDictObject* self)
 {
     Py_ssize_t size = sizeof(MultiDictObject);
-    if (self->keys != &empty_htkeys) size += htkeys_sizeof(self->keys);
+    Py_BEGIN_CRITICAL_SECTION(self);
```
asvetlov, line 1267: It's very minor, but I doubt that the critical section is required here.
For all other readers, we apply relaxed atomic loading for such situations.

multidict/_multilib/htkeys.h
```diff
@@ -335,6 +373,9 @@ htkeys_free(htkeys_t* dk)
 {
     /* TODO: CPython uses freelist of key objects with unicode type
        and log2_size == PyDict_LOG_MINSIZE */
+    if (dk->resume_slots != NULL) {
+        PyMem_RawFree(dk->resume_slots);
```
asvetlov, line 377: Why is a raw allocator used? For htkeys, we call `PyMem_Free`
I guess mimalloc fits better in FT mode, while pymalloc is optimized for GIL.
Raw system malloc could get slower and more fragmented memory allocation

asvetlov (commented):
There my brief review

In Free-threaded mode, we use atomic readers/writers.
Why does `_htkeys_find_empty_slot_resume` avoid it?

The resume slots are not copied by multidict cloning.
Perhaps it is good, I don't know

multidict/_multilib/htkeys.h
bdraco: Will switch it to `PyMem_Calloc`/`PyMem_Free`

I went with the raw allocator because of your comment at https://github.com/aio-libs/multidict/blame/838ebf7ba71a79c33407136c4a2be8bf25c17d42/multidict/_multilib/hashtable.h#L807   I was worried every `md->keys` needed a recheck because of it, but I'm not that familiar with the free threaded apis. Does the resume slot allocation need a recheck of `md->keys` too?

bdraco:
> The resume slots are not copied by multidict cloning.
> Perhaps it is good, I don't know

That's an intentional trade-off. If there is never another long probe, the copy allocates nothing, and if there is one, rebuilding is cheap. So the common case doesn't pay for another allocation.

multidict/_multilib/htkeys.h
asvetlov: Interesting point. I use PyMem_Malloc for htkeys allocation because CPython does so for `dict`.
I recall that raw system allocators are slower and dont support tracing, I'll check this later.

multidict/_multidict.c
bdraco: I added it because the previous bot review was concerned about the `self->keys` access. I read the concern as a concurrent resize leaving it pointing to freed memory. This path isn't hot so a critical section seemed reasonable to add.

bdraco:
> In Free-threaded mode, we use atomic readers/writers.
> Why does `_htkeys_find_empty_slot_resume` avoid it?

The access is already guarded by the critical section on every insert path (`multidict_add`) and the lock free reads never read `resume_slots` so there shouldn't be anything to protect with atomics unless I missed somthing.

multidict/_multilib/htkeys.h
bdraco: went through it and then I asked claude to check it and I couldn't find a place where it is unsafe, maybe I missed something.

switched in 3d2876df4a19473c43638ec5e6c83dbcf288a2a4

asvetlov: yes, python uses custom mem pools to speed-up memory allocations.
For free-threading it uses mimalloc under the hood for the same purpose.
I'll dig deeper into the FT subject for curiosity, but I'm pretty sure that we have to use PyMem_Malloc instead of PyMem_RawMalloc.

multidict/_multilib/htkeys.h
```diff
@@ -358,13 +399,59 @@ _unicode_hash(PyObject* o)
...
+    while (htkeys_get_index(keys, i) != DKIX_EMPTY) {
+        i = (i * 5 + 1) & mask;
+        steps++;
+    }
+    if (resume_slots == NULL) {
+        size_t nbytes = htkeys_resume_slots_bytes(keys->log2_size);
+        if (steps < HT_RESUME_SLOTS_MIN_STEPS || nbytes == 0) {
+            return (Py_ssize_t)i;
+        }
+        resume_slots = keys->resume_slots = PyMem_Calloc(1, nbytes);
```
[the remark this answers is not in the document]
bdraco: I don't think this can happen. Please prove it can.

AFAICT `PyMem_Calloc` on free-threaded `mi_heap_calloc` on the threads own heap and mimalloc uses atomics so it should be safe.. maybe I am reading it wrong.  Definitely not an expert on desktop allocator design as its far more complex than the ESP8266s I'm used to debugging

multidict/_multidict.c
aiolibsbot: Critical section justified here. `keys->resume_slots` is a plain store (`htkeys.h:431`), not atomic. A relaxed lock-free read would race with it. Relaxed loads would need atomic stores on the write side too. The `sizeof` path isn't hot.

multidict/_multilib/htkeys.h
bdraco: > I'll dig deeper into the FT subject for curiosity, but I'm pretty sure that we have to use PyMem_Malloc instead of PyMem_RawMalloc.

Thank you

asvetlov: Yes, it was an oversight generated by my claude and not caught by me.
I'll publish the cleanup of this overcomplicated double-allocation logic tomorrow if I have time for it.
But please let me finish reading the allocator's code first.

multidict/_multilib/htkeys.h
asvetlov: neat:
```suggestion
        resume_slots = keys->resume_slots = PyMem_Malloc(nbytes);
```
It really changes nothing, but the proposed version doesn't scrape my eyes :)

asvetlov (approved):

multidict/_multilib/htkeys.h
asvetlov: See #1469 for discussed dead code removal

asvetlov:
https://github.com/aio-libs/multidict/pull/1473 demonstrates a different approach: linear probe after the first match instead of keeping the perturbation probe.
I really don't know what is better. Let me see on benchmarks and sleep on it

bdraco:
> #1473 demonstrates a different approach: linear probe after the first match instead of keeping the perturbation probe. I really don't know what is better. Let me see on benchmarks and sleep on it

Thanks. Much simpler design. 👍  If it doesn't regress performance, I'd prefer that.  Guess we need to wait for codspeed to decide

asvetlov:
@bdraco it seems like the probe from #1473 behaves worse than already applied for pathological cases.
I suggest applying your fix from this PR.
Could you please fix CI and make it green?

multidict/_multilib/htkeys.h
aiolibsbot: Applied in 2a0fe6a. One caveat: zeroing stays load-bearing. Garbage hint points past an empty slot, entry lands beyond it, `iter_hash()` stops at that empty slot and never finds it. Follow-up 211d2c8 re-adds `memset()` after the malloc, so net behaviour matches `PyMem_Calloc`.

bdraco:
Thanks

PR #1469, opened 2026-09-16
Drop dead code

asvetlov:
PyMem_Malloc never suspends an acquired critical section.
The code for this guard was added by oversight by #1441; it is never executed.


asvetlov:
The dead code was never published; the removal doesn't require a changelog.

PR #1473, opened 2026-09-16
Use linear probing after the initial slot in the hash table

asvetlov:
## What do these changes do?

Switches the internal hash table probe sequence, in both the C
extension and the pure Python implementation, from CPython dict's
perturbation formula to linear probing after the initial slot.
Collision chains land in consecutive slots instead of being
scattered across the table, which is friendlier to the cache; the
effect is largest when many values are added under the same key.

No, only performance. Iteration order is unaffected since it comes
from the separate entries list, not the probe sequence.

No. The change is contained to the probe-sequence step in
``htkeys_build_indices``, ``htkeys_find_empty_slot`` and
``htkeysiter_next`` (C extension) and their pure Python counterparts,
with no public API change.

## Related issue number

N/A. Related to #1466, which targets the same slow path (many values
added under one key) with a different, more targeted technique: it
caches where the probe sequence's post-perturbation "tail" last ended
and resumes from there, leaving the general probe sequence untouched.
This PR is a simpler alternative offered for comparison; see the
benchmark note below.

## Checklist

- [ ] Documentation reflects the changes N/A
- [ ] If you provide code modification, please add yourself to `CONTRIBUTORS.txt` N/A, file does not exist in this repo

<details>

No new dedicated test was added; the existing suite (including the
`hypothesis`-marked property tests) already exercises every changed
line under both backends, confirmed via the coverage report before
committing.

Tests: `pytest -q` on master before the change and on this branch,
both backends parametrized: 2020 passed, 10 skipped, both runs.
`pytest -q -m hypothesis`: 142 passed. Debug build with
`MULTIDICT_DEBUG_BUILD=1 MULTIDICT_ASAN_BUILD=1`, ASan+UBSan
preloaded: 2009 passed, no sanitizer reports.

Lint: `pre-commit run` on the changed files (ruff, clang-format,
mypy x2): all passed. `make doc-spelling`: passed.

Benchmark (own microbenchmark, not the codspeed one from #1465,
wall clock, C extension, release build): building a `MultiDict` from
200 keys with 200 `add()` calls each (40000 inserts total, all
colliding within each key's 200-entry chain) went from ~20-24ms to
~6-8ms, roughly 3x. Unique-key workloads (build/get/iterate on 20000
distinct keys) were unchanged within noise. This is a constant-factor
cache-locality win; each insert is still O(chain length). By
contrast, #1466's own numbers show an asymptotic improvement for the
same scenario (2.2B to 6.8M instructions for 20000 values under one
key), since it stops re-walking already-placed entries entirely. The
two approaches are not mutually exclusive but this PR does not
attempt to combine them.

</details>

Drafted with Claude Code; reviewed by @asvetlov.


Mukller:
LGTM. The linear probing change is clean — perturbation-based probing (i * 5 + perturb + 1) replaced with simple (i + 1) & mask in both Python and C implementations.

Cache locality improvement for collision chains is a reasonable trade-off, especially for the dd-many-values-to-same-key use case. The HT_PERTURB_SHIFT constant removal and perturb field cleanup in the iterator struct are thorough.

CHANGES file included. Approved.

asvetlov:
The experiment show bad result with pathological cases, when different keys land on the  same slot

PR #1480, opened 2026-09-19
Add a public C API capsule, with optional Cython support

asvetlov:
## What do these changes do?

Adds a public C API capsule (`multidict._multidict.CAPI`, reached
through `multidict_capi.h` / `multidict_capi_struct.h`) so other C
extensions can create and manipulate `MultiDict`, `CIMultiDict`,
their proxies, and `istr` directly, without going through the
Python-level API. Covers type checks and constructors for all four
mapping types plus `istr`, `getversion()` and `len()`, and
`add()`, `clear()`, `setdefault()`, and single-item access (`in`,
`[]`, `del`, `pop()`) for the mutable types. `GetItem`, `Pop`, and
`SetDefault` follow the `PyDict_GetItemRef` / `PyDict_SetDefaultRef`
design, reporting a missing key through the return code rather than
an exception. `MultiDict_ForEach` visits every item, or just the
items for one key, through a callback, without allocating a list.

The same capsule is also reachable from Cython: `multidict/__init__.pxd`
lets third-party Cython code `cimport multidict`, mirrored by a
`multidict/_testcyapi.pyx` test harness. Cython is optional and never
a real dependency of `multidict` itself; an ordinary or release build
never sees it, and a dedicated CI job exercises the opt-in path.

No behavior change for existing Python-level usage. This adds a new,
opt-in C API surface for other C extensions and Cython code to use.

The struct only ever grows at the end, so older clients keep working
against newer `multidict` builds. The Cython binding is test-only and
opt-in, so it adds no ongoing dependency burden.

## Related issue number

Supersedes and closes #1183, #1190, #1191 (earlier C API drafts).

## Checklist

<details>

Tests: `pytest -q` on both the accelerated build (with and without the
Cython opt-in) and the pure-Python build (`MULTIDICT_NO_EXTENSIONS=1`);
`tests/test_capi.py`'s `api` fixture runs every case against both the
C and Cython bindings when Cython is opted in.
Lint: `pre-commit run --all-files` (ruff, mypy for 3.13/3.11, yamllint,
clang-format, GitHub Actions validators) clean.
Docs: strict Sphinx HTML build and `doc-spelling` clean.

</details>


asvetlov:
@Vizonex this PR is heavily based on your previous work.
Design note: not all methods are exposed. For example, `md.update()` has O(N) complexity; calling it as `PyObject_CallMethod(md, "update", "O", other)` and family; the extra overhead is relatively small.
Also, views and iterators are omitted; the new `MultiDict_ForEach` provides a native C API for iteration.

docs/conf.py
```diff
@@ -367,7 +367,11 @@
 
 # -- Strictness options --------------------------------------------------
 nitpicky = True
-nitpick_ignore: list[str] = []
+nitpick_ignore: list[tuple[str, str]] = [
```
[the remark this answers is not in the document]
asvetlov: It's the sphinx documented flag, it's used

Vizonex (approved):
Glad to see us coming back to this. I had worked several months on this on in particular and was quite disappointed to not see any further progress. The problem with last time was that we were having trouble triggering the tests but hopefully this time this won't be a problem. This should speedup aiohttp with parsing and writing requests and I don't think this will be a burden for new contributors.

Vizonex:
I'm sure will find a way around the coveralls workflow this still looks good to me I would happy to work on merging this to aiohttp once it's over and I'm ok with rewriting my old PR because the code is pretty tiny and normally when I come back and rewrite something a few times it normally turns out cleaner and better in the long run.

Vizonex:
@asvetlov I do have a small question however, are the tests and coveralls checks ignorable or do we need to find a way to fulfill these workflow requirements?

Vizonex:
Also Feel free to mark that this closes #1183 and the other PRs we did back in 2025.

asvetlov:
@Vizonex I'll add the coverage ignore instruction for this one missing line.
The PR is in a stable state but I want to apply minor polishing (give some time for review)

multidict/multidict_capi_struct.h
```diff
@@ -0,0 +1,61 @@
...
+                         PyObject** result);
+    int (*MultiDict_SetDefault)(void* state, PyObject* self, PyObject* key,
+                                PyObject* default_value, PyObject** result);
+    int (*MultiDict_SetItem)(void* state, PyObject* self, PyObject* key,
+                             PyObject* value);
+
+    Py_ssize_t (*MultiDict_ForEach)(void* state, PyObject* self, PyObject* key,
+                                    MultiDict_ItemVisitor visitor,
+                                    void* user_data);
+} MultiDict_CAPI;
```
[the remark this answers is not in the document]
asvetlov: done

multidict/__init__.pxd
```diff
@@ -0,0 +1,98 @@
...
+# per-translation-unit storage with no cross-module lookup involved.
+cdef extern from *:
+    """
+    static inline MultiDict_CAPI *
+    __Pyx_multidict_capi(void)
+    {
+        static MultiDict_CAPI *cached = NULL;
+        if (cached == NULL) {
+            cached = MultiDict_GetCAPI();
+        }
```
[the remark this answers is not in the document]
asvetlov: outdated

multidict/_multilib/capsule.h
```diff
@@ -0,0 +1,471 @@
...
+_md_foreach_key(MultiDictObject* md, PyObject* key,
+                MultiDict_ItemVisitor visitor, void* user_data)
+{
+    PyObject* identity = md_calc_identity(md, key);
+    if (identity == NULL) {
+        return -1;
+    }
+    Py_ssize_t count = 0;
+    bool failed = false;
+    md_finder_t finder = {0};
```
asvetlov:
```suggestion
    md_readonly_finder_t finder;
```

multidict/_multilib/capsule.h
```diff
@@ -0,0 +1,471 @@
...
+    Py_BEGIN_CRITICAL_SECTION(md);
+    if (md_init_finder(md, identity, &finder) < 0) {
```
asvetlov:
```suggestion
    if (md_readonly_finder_init(md, identity, &finder) < 0) {
```

multidict/_multilib/capsule.h
```diff
@@ -0,0 +1,471 @@
...
+        failed = true;
+    } else {
+        PyObject* k;
+        PyObject* v;
+        int found;
+        while ((found = md_find_next(&finder, &k, &v)) > 0) {
```
asvetlov:
```suggestion
        while ((found = md_readonly_find_next(&finder, &k, &v)) > 0) {
```

multidict/_multilib/capsule.h
```diff
@@ -0,0 +1,471 @@
...
+                break;
+            }
+            if (ret == 0) {
+                break;
+            }
+        }
+        if (found < 0) {
+            failed = true;
+        }
+        md_finder_cleanup(&finder);
```
asvetlov:
```suggestion
        md_readonly_finder_cleanup(&finder);
```

tests/test_capi.py
```diff
@@ -0,0 +1,576 @@
...
+except ImportError:
+    _testcyapi = None
+
+if _testcyapi is None and os.environ.get(
+    "MULTIDICT_TEST_REQUIRE_CYAPI"
+):  # pragma: no cover
+    raise RuntimeError(
+        "multidict._testcyapi did not import, but MULTIDICT_TEST_REQUIRE_CYAPI "
+        "is set: the Cython opt-in build was expected to succeed"
+    )
```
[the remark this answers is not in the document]
asvetlov: it's a line in tests that enforces Cython API usage. It doesn't require coverage

asvetlov:
@aiolibsbot rr

docs/conf.py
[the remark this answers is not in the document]
aiolibsbot: Correct, CodeQL false positive. `nitpick_ignore` gets read by Sphinx, not by Python callers. Static analysis cannot see module-level config consumption. Safe to dismiss.

multidict/multidict_capi_struct.h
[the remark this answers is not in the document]
aiolibsbot: Confirmed at HEAD. `api_version` sits first in `MultiDict_CAPI`, before any appended field, so an older runtime's smaller struct still exposes it. `_MultiDict_CheckAPIVersion` rejects older runtimes; `test_check_api_version_rejects_older` covers it.

tests/test_capi.py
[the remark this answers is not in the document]
aiolibsbot: Agreed. Env-var guard only fires on misconfigured CI. `# pragma: no cover` correct there.

asvetlov:
@aiolibsbot rr

multidict/_multilib/capsule.h
aiolibsbot: Applied at the reviewed head. `capsule.h:358-383` now uses `md_readonly_finder_t`, `md_readonly_finder_init`, `md_readonly_find_next`, `md_readonly_finder_cleanup` — all four suggestions here landed.

One leftover: the comment at `capsule.h:311-313` still explains the reentrancy ban via marked entries and `md_finder_cleanup`. The readonly finder never marks. Real guard is the version check at `hashtable.h:1598`. Same stale wording in `docs/capi.rst:321-325`.

multidict/__init__.pxd
```diff
@@ -0,0 +1,175 @@
...
+# The trampoline can't just `raise` on a callback exception -- it's
+# assigned to a `noexcept` C function pointer, so Cython would treat that
+# as an unraisable exception here, print it, and clear it rather than
+# propagate it. PyErr_SetObject is a raw C call, not a Cython `raise`: it
+# reinstates the caught exception as the current one without going
+# through that noexcept-triggered unraisable-and-clear handling, so the
+# -1 this function then returns has a real exception attached, which
+# MultiDict_ForEachAll/Key's `except -1` propagates normally.
+
+cdef inline int _py_visitor_trampoline(void *user_data, PyObject *key, PyObject *value) noexcept:
```
Vizonex: I'm still questioned about using a python callback over a C Callback but I get the idea of needing to capture exceptions took me a bit to realize now that I finally made a copy onto my PC to review.

multidict/__init__.py
```diff
@@ -58,3 +60,12 @@
...
 upstr = istr
+
+
+def get_include() -> str:
+    """Return the directory containing multidict's C API headers.
+
+    Pass it as an ``-I`` include path when building a C extension
+    against ``multidict_capi.h``.
+    """
+    return str(Path(__file__).absolute().parent)
```
Vizonex, line 71: Glad to see this coming back as the old ways were quite confusing and made the process of compiling to cython frustrating.

multidict/multidict_capi.h
```diff
@@ -0,0 +1,267 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include "multidict_capi_struct.h"
+
+static inline int
+_MultiDict_CheckAPIVersion(MultiDict_CAPI* capi)
+{
+    if (capi->api_version < MultiDict_CAPI_VERSION) {
```
Vizonex, line 13: I'm quite fond of this section however wouldn't it be better to utilize multidict's version number instead of the C-API Version?

multidict/_multilib/capsule.h
```diff
@@ -0,0 +1,471 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include "../multidict_capi_struct.h"
+#include "dict.h"
+#include "hashtable.h"
+#include "state.h"
+
+#define __MULTIDICT_VALIDATION_CHECK(SELF, STATE, ON_FAIL)           \
```
Vizonex, line 13: Quite happy to see my good old macros being put to good use. These sections can be quite repetitive if needed over and over again.

Vizonex (commented):
I'll keep my suggestions short and sweet so that we can merge this after over a year of waiting. I mostly just have questions about a few things seeing that I have a strong desire to start making Aiohttp's Cython sections faster at a future point in time. You don't have to take my advice if it isn't needed.

multidict/_multilib/capsule.h
```diff
@@ -0,0 +1,471 @@
...
+    return ret;
+}
+
+static int
+MultiDict_SetItem(void* state_, PyObject* self, PyObject* key, PyObject* value)
+{
+    __MULTIDICT_VALIDATION_CHECK(self, state_, -1);
+    int ret;
+    Py_BEGIN_CRITICAL_SECTION(self);
+    ret = md_replace((MultiDictObject*)self, key, value);
```
Vizonex:
```suggestion
#ifdef Py_GIL_DISABLED
    md_deferred_decref_t defer;
    md_deferred_decref_init(&defer);
    ret = md_replace((MultiDictObject*)self, key, value, &defer);
#else
    ret = md_replace((MultiDictObject*)self, key, value);
#endif
```
@asvetlov You forgot an important step that is causing the inconsistencies with free-threaded related builds. I'm sure there might be others but this was the only one I could find being a problem.

Vizonex (changes requested):
Something isn't right with the compilation of `md_replace` I've left you a suggestion on what to fix for these builds.

multidict/__init__.pxd
asvetlov: Good catch.
Perhaps in Cython we need a trampoline that calls `cdef int cb(object key, object val, <void*> callback)`. Argument types are adjusted to more convenient `object` ones, but it is still a C function, not a pyobject callable.

multidict/multidict_capi.h
asvetlov: it's for protecting from future-incompatibilities when a C Ext is compiled with v2 multidict library version but called with v1.
In this case, C Ext may access `MultiDict_CAPI` slots that were added in v2. These slots are absent in v1; thus we have a boundary access violation and either a segfault or a jump to a garbage address.

multidict/_multilib/capsule.h
asvetlov: The master should accept `&defer` in both modes starting from today; it simplifies the total logic. Deferred decref is still useful in GIL mode when `__del__` is implemented and has side-effects that modifies the multidict itself. 
Let me rebase the PR properly.

multidict/__init__.pxd
asvetlov: @Vizonex please check the updated Cython iteration API. Feel free to raise a flag if something should be improved or fixed

multidict/_multilib/capsule.h
asvetlov: done

Vizonex (approved):
Nothing to complain about now. I like this and I hope other developers will enjoy using it too. I had some ideas for ls-hpack which is another C Library that packs http2 frames and utilizing multidict with it would enhance it's speed so know that I have some ideas for this C-API outside of aiohttp in general.

Vizonex:
I'm sure the [codspeed](https://github.com/apps/codspeed) problem might be from the updated patches and bug fixes. But as far as I'm aware That should be an ignorable problem caused by branches not being updated

Vizonex:
Were very close, `test_multidict_getall_str_hit_many_values` just needs readjustment unless you decide something different.

asvetlov:
ok, let's merge
The master has no critical bugfixes now that prevent crashing the interpreter in *normal* usage.
Rare cases like collection-modification-in-del are not blockers.
Thus, I'll merge this PR; the next release will be 7.0

asvetlov:
aiohttp probably should stay with conditional logic for a while; use CAPI if available with fallback to python method calls

Vizonex:
@asvetlov Thank you. I have been waiting for this for over a year now, glad to see our work finally accomplished.

PR #1515, opened 2026-09-22
Stop re-testing cibuildwheel wheels on macOS and Windows

asvetlov:
## What do these changes do?

Removes duplicated test runs from CI, mostly on macOS where runners are scarce. Numbers below come from master run 35703746757.

The macOS and Windows legs of the `test` matrix installed the same wheels cibuildwheel had just tested and ran the same suite again. They upload no coverage (coverage is collected on ubuntu only), so they only repeated a pass/fail signal cibuildwheel already gives. They are dropped; cibuildwheel stays the single place where macOS and Windows wheels are tested, pure-Python tests included.

Every leg of `test` waited for all wheel builds, the slow macOS one included, even the debug legs that build from source and the pure-Python legs. Those 13 legs move to a new `test-without-cibuildwheel` job that needs only the sdist and the pure-Python wheel, so they now start around 0.5 min into the run instead of about 9.5 min. The pure-Python legs therefore install the real `py3-none-any` wheel instead of the platform wheel.

Rosetta: `macos-latest` is an arm64 runner. The x86_64 wheels are cross-compiled there and then tested by running an x86_64 CPython under Rosetta 2. Per ABI that pytest run takes 23 to 34 s, against 11 to 16 s natively on arm64, and all 8 x86_64 ABIs take about 6.8 of the job's 9.4 min, including build and test environment setup. Outside release tags, `CIBW_TEST_SKIP` now skips the x86_64 test run for everything except `cp314` and `cp314t`. All 8 x86_64 wheels are still built on every run, so compile and link problems still show up immediately. One GIL and one free-threaded ABI still exercise x86_64 code generation and its memory model on every PR, and tag builds test every x86_64 ABI as before. What is given up is catching an x86_64-only bug that affects only 3.10 to 3.13 or 3.15 before the release tag.

cibuildwheel no longer runs the CodSpeed benchmark bodies (`*_benchmarks.py`), which only matter in the Benchmark job; this takes the per-wheel suite from 2151 to about 1860 tests.

The `windows-11-arm` exclude was meant to limit that build to release tags but never matched: `cond && '' || 'none'` always yields `'none'` because `''` is falsy. The condition is inverted, and the equivalent musllinux exclude is rewritten the same way.

Expected effect per PR run:

| | before | after |
|---|---|---|
| macOS runners | 8 (cibuildwheel 9.4 min, plus 7 test jobs of 1 to 2 min) | 1 (cibuildwheel, about 4 to 5 min) |
| Windows runners | cibuildwheel 5.7 min, windows-11-arm 7.3 min, 7 test jobs | cibuildwheel only |
| Rosetta test runs | 8 ABIs | 2 ABIs (all 8 on tags) |
| debug and pure-Python legs start | after about 9.5 min | after about 0.5 min |

Wall clock is then bounded by the Benchmark job (12.8 min); sharding it is left for a follow-up.

No, CI only.

No. The test job steps are shared between the two test jobs through YAML anchors, so they are still written once.

## Related issue number

N/A

## Checklist

- [ ] Unit tests for the changes exist: N/A, CI configuration
- [ ] Documentation reflects the changes: N/A
- [ ] If you provide code modification, please add yourself to `CONTRIBUTORS.txt`: N/A

Drafted with Claude Code (Claude Opus 5); reviewed by @asvetlov.

<details>

Lint: pre-commit (check-github-workflows, timeout-minutes check, yamllint) and actionlint 1.7.12 pass on `ci-cd.yml`.
Spelling: `make doc-spelling` passes.
New cibuildwheel test command against a local C build (CPython 3.13): 1858 passed, 19 skipped, 9 deselected.
Pure-Python leg against the `py3-none-any` wheel, `python -Im pytest tests --no-c-extensions` from the unpacked sdist: 1079 passed, 15 skipped. Same run with the C build installed differs only by `test_no_warnings[multidict._multidict]`, which is not collected without the C module.
Job layout, Rosetta skip and windows-11-arm exclusion still need confirming on this PR's CI run.
</details>


asvetlov:
@webknjaz, please review if you have time.
The PR is related to CI infrastructure; you are the expert in this area.

webknjaz (changes requested):
This seems to go against the overall strategy of testing pre-built dists in dedicated jobs by installing them the way the end-users would, and collecting coverage on that.

Additionally, there's a bunch of instrumentation that has to be kept in-sync across projects and so random changes that make the infra differ across them makes it rather impossible to have a shared env later.

YAML anchors have proven to be difficult to maintain in the past, when overused. Even in here, there's a few obvious places where it creates spaghetti across GHA/YAML render boundaries which is going to be difficult to debug once a mistake is made.

Finally, any change to the infra must be made as a part of of a series of PRs prepared and approved together across all the repositories sharing this infra so that it doesn't drift over time.

---

If there's test duplication, it probably makes sense to drop it from the build jobs and keep it in the test jobs.

